PR #634, opened 2014-05-12
Fixed ActiveRecordObject->save() returning wrong results, Fixed concrete inheritence save-loop-bug 

tals56:
I have added code to ObjectBuilder.php which fixes returning wrong results when performing updates,
and now throwing an exception when trying to update non-existant objects.

Also, 2 bugs in concrete inheritence were fixed:
1.getSyncParent() - now the inherited object stores its parent, and returns the same one when the parent object has been created \ retrieved.

2.Loop-saving when saving parent - a quite massive bug causing 2 parent objects to be saved, due to loop-saving in inherited objects, and extra non-required SQL statements to be performed.


dereuromark:
Almost 6 years stale, closing then.

Issue #687, opened 2014-07-31
[Migration] propel fetches version from wrong table when having two database in two schemas


marcj:
Given:
- Two connections defined (in propel.yml), connection-a, connection-b. Both point to a different schema.
- Two databases defined (in schema.xml), database-a, database-b
- One `PropelMigration_*` file exists.
- At Database-a was the migration file applied.
- I deleted all tables in connection-b (hence no propel version table).
- Run `migration:diff`

Result: 'All migration files already applied' since it reads migration information from connection-a.

It seems that migration commands fetches versions of the first connection only instead of all version in all defined connections.

PhilinTv:
Justification to include in Beta6.

**Scenario**: Two connections (connection-a, connection-b) pointing to separate schemas/databases.
**Bug**: Migration status checks read only from the first connection’s migration table.
**Impact**: On diff or migrate, Propel falsely reports that migrations are already applied even if they are not applied in one of the other schemas.
**Result**: Migrations are skipped unintentionally, possibly leaving the DB in an inconsistent state.

PR #724, opened 2014-08-23
[Migration] propel fetches version from wrong table when having two database in two schemas

snoopckuu:
Fix for #687.

Not sure is it correct to fix it like that, but it work's. :)

Previously it checked only first connection. Now it checks each database and migrate step by step.


marcj:
To keep things clean I'm closing this PR due to inactivity and outdated codebase. If you want to reopen it, please leave a comment, I'd love to merge it :)

PR #768, opened 2014-10-06
Fix wrong DateTime comparison in setter methods

dhm80:
Hey there,

I noticed that the setter methods don't check for modified columns correctly when handling `date`, `time` or `timestamp` values.

One problem was that on every call first a new instance of the given date and/or time parameter `$v` was generated and then compared via php's `!==` operator (which checks if two variables reference the same object and NOT if the values are equal). The result could not become true as there were always two different objects that were compared.

Another "problem" was that when e. g. comparing objects belonging to a `date` column the time part should not be considered, etc. (even if an object is given to the setter method).

An example:
```php
$test = TimingQuery::create()
	->filterById(123)
	->findOneOrCreate()
;

$almostTomorrow = new \DateTime("tomorrow -1 second");
$test
	->setDate($almostTomorrow)
	->setTime($almostTomorrow)
	->setDateTime($almostTomorrow)
;

// should only be true on first execution
echo "isModified: " . (int) $test->isModified() . "\n";

if ($test->isNew()) {
	$test->setId(null);
}

$test->save();
```

So I searched the generator classes and found the part responsible for generating setter methods for mutable columns.

Please have a look at my changes and let me know if there is a better way to solve the problem or you need any other information.

Otherwise: thanks for merging ;)

### DIFF
--- a/src/Propel/Generator/Builder/Om/ObjectBuilder.php
+++ b/src/Propel/Generator/Builder/Om/ObjectBuilder.php
@@ -1615,12 +1615,22 @@ abstract class ".$this->getUnqualifiedClassName().$parentClass." implements Acti
                 || (\$dt->format($fmt) === $defaultValue) // or the entered value matches the default
                  ) {";
         } else {
+            switch ($col->getType()) {
+                case 'DATE':
+                    $format = 'Y-m-d';
+                    break;
+                case 'TIME':
+                    $format = 'H:i:s';
+                    break;
+                default:
+                    $format = 'Y-m-d H:i:s';
+            }
             $script .= "
-            if (\$dt !== \$this->{$clo}) {";
+            if (\$this->{$clo} === null || \$dt === null || \$dt->format(\"$format\") !== \$this->{$clo}->format(\"$format\")) {";
         }
 
         $script .= "
-                \$this->$clo = \$dt;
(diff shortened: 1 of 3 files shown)


marcj:
ping @dhm80 :smile:

dhm80:
Sorry @marcj - hat alles ein wenig länger gedauert, aber nun ist der Test hochgeladen und von CircleCI abgenommen.

dhm80:
@marcj are the provided tests okay? If anything else is missing just let me know.

marcj:
Yes, pretty good. Would be cool if you could squash your commits, then this baby is ready to be merged :smile:

dhm80:
All changes are squashed into  951bb5a now ;)

marcj:
Oh, I guess it's not correctly squashed :) The tab `commits` should contain only one commit.

marcj:
Well, I cherry-picked it now. Thanks :)
https://github.com/propelorm/Propel2/commit/8bd56568ca7eacb3ff7e90410accb67acc9f90ae

dhm80:
Hey @marcj, thanks for merging the changes. I was able to sqush the commits into one new commit but I was not able to "remove" the existing commits in this pull requests. Is there a trick I missed? How should I have squashed the commits correctly to have only one commit left in this pull request? Cheers ;)

PR #786, opened 2014-10-19
String values in native ENUM column

oujesky:
Some RDBMS (e.g. MySQL) provides native ENUM column type. Propel however keeps using numeric values internally for the enumerated column even if you force the column type with `sqlType="ENUM('one', 'two', 'three')"` and `phpType="string"` in `schema.xml`.

Back in the day the solution was to modify the value set in Peer class (as discussed here: https://groups.google.com/forum/#!msg/propel-development/3DhvA-9bUDg/Q0Z3zY39DUkJ). With the advent of Peer classes, there is no such simple solution. 

This PR modifies the behaviour of ENUM columns which are also set as native ENUM through `sqlType` attribute in schema by allowing the values actually inserted from Propel to database to be string.

The solution is bit hackish for my taste, but I am not very familiar with Propel internals, so I would use some guidance to see if there is a better place for doing this ENUM "magic". 

Example column definition:

``` xml
<column name="status" type="ENUM" valueSet="NEW,OPEN,COMPLETED,CANCELLED" sqlType="enum('NEW','OPEN','COMPLETED','CANCELLED')" phpType="string" />
```

It would be also nice to have `sqlType` ENUM values automatically filled from `valueSet` to avoid repetition, but since this is RDBMS specific, I don't know how it fits in Propel's philosophy (Behavior?).

### DIFF
--- a/src/Propel/Generator/Builder/Om/ObjectBuilder.php
+++ b/src/Propel/Generator/Builder/Om/ObjectBuilder.php
@@ -185,6 +185,8 @@ class ObjectBuilder extends AbstractObjectBuilder
                 throw new EngineException(sprintf('Default Value "%s" is not among the enumerated values', $val));
             }
             $defaultValue = array_search($val, $valueSet);
+            settype($defaultValue, $column->getPhpType());
+            $defaultValue = var_export($defaultValue, true);
         } elseif ($column->isPhpPrimitiveType()) {
             settype($val, $column->getPhpType());
             $defaultValue = var_export($val, true);
@@ -1794,8 +1796,9 @@ abstract class ".$this->getUnqualifiedClassName().$parentClass." implements Acti
 
         $script .= "
         if (\$v !== null) {
+            \$v = (string) \$v;
             \$valueSet = " . $this->getTableMapClassName() . "::getValueSet(" . $this->getColumnConstant($col) . ");
-            if (!in_array(\$v, \$valueSet)) {
+            if (!in_array(\$v, \$valueSet, true)) {
                 throw new PropelException(sprintf('Value \"%s\" is not accepted in this enumerated column', \$v));
             }
             \$v = array_search(\$v, \$valueSet);
--- a/src/Propel/Generator/Builder/Om/QueryBuilder.php
+++ b/src/Propel/Generator/Builder/Om/QueryBuilder.php
@@ -1029,14 +1029,14 @@ abstract class ".$this->getUnqualifiedClassName()." extends " . $parentClass . "
             $script .= "
         \$valueSet = " . $this->getTableMapClassName() . "::getValueSet(" . $this->getColumnConstant($col) . ");
         if (is_scalar(\$$variableName)) {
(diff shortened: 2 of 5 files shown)


gharlan:
@oujesky any news?

oujesky:
Sorry for the delay guys. I rebased and squashed the commits. 

However I gave it a second look and I still think this is not the best solution. I did a little research on the supported DB platforms and how enum-like types works there. I think the generator can decide which actual DB column type should be used based on phpType setting in schema. The default phpType is integer and with that stays the current default enum behaviour. But when it would be set to string, the string based enum type can be enforced. 

On MySQL and PostgreSQL, there is the native ENUM column type:

    ... column ENUM('small', 'medium', 'large') ...

On MSSQL, SQLite and Oracle it can be simulated to certain degree with CHECK constraint:

    ... column VARCHAR(10) CHECK(column IN ('small', 'medium', 'large')) DEFAULT 'small' ...

I guess it would be better to create a separate "request for comments" issue though, as it needs to be discussed a bit. The main purpose is not to change the default behaviour of how Propel approaches enums, but to allow legacy DB structure work well with Propel. What do you think?

marcj:
After a second look I'd like to note following:

1. As I want this behavior as default I guess it's better to make it vice-versa, meaning `phpType="string"` is not necessary to have real enums in the database
2. `phpType` is not the right place to say the enum type if they should store it as integer or string, as `phpType` means the type of actual php code, not database schema.
3. We should make sure that strings are now at the object property, not that legacy hack with integers. Integers should only be read at runtime through `array_search`.

So this means making Propelv2 enum-aware is much more work, so I'd like either to close this PR or someone is taking the time to invest more energy in it to make it the right-way™ :smile:

jeremydenoun:
I work on Native Enum usage and I work about more simple solution based on @oujesky solution
but with only :
- default value (Table creation) based on schema if column sqlType is defined
- getValue for Enum return 'content' (and not index) if domain sqlType != 'TinyInt' (workaround for create/update keep aware from internal structure and query raw value)

this work if I define a schema like :
`<column name="state" type="ENUM" valueSet="draft,running,expired,cancel" sqlType="enum('draft','running','expired','cancel')" required="true" defaultValue="draft"/>`

I think this is not perfect because I check original sqlType with "Type Checking" (TINYINT) but Oracle use (NUMBER) and PGSQL (INT2) but this is compatible with indexed actual approach (without sqlType)

If you consider my approach is good, I can implement platform checking and workaround if domain sqlType differ to waited ENUM default type and asked a PR

ref: 
https://github.com/jeremydenoun/Propel2/tree/propel-enum-type

marcj:
@jeremydenoun, looks pretty good! We are looking forward to your pull request :)

marcj:
I saw @oujesky added also some tests, could you then also please add some tests? We should make sure this works as intended.

marcj:
I'm closing this since we implemented enums in our data-mapper branch, which will be merged into master soon.

Issue #803, opened 2014-11-09
Throw exceptions instead of returning null when no row found


mpscholten:
Propel's Query classes should be able to throw exceptions instead of returning null when no object was when calling `find()`.

This could look like:
```php
$post = PostQuery::create()->filterById($id)->findOneOrFail();
```
The `findOneOrFail` method would throw an exception (should be configurable in the propel build settings) in case no Post with the specific id is found.

---

This is how laravel's eloquent implements this:
```php
$post = Post::findOrFail(1);
```
*(see http://laravel.com/docs/4.2/eloquent)*

This is how Rails ActiveRecord implements this:
```ruby
post = Post.find(id) # will raise an ActiveRecord::RecordNotFound exception when no post found
```
*(see http://guides.rubyonrails.org/active_record_querying.html)*

---

The advantage you gain by throwing exceptions instead of returning null is that you now don't need to check for null in any controller actions your mvc application has. The exception can be catched at a central point (e.g. the symfony kernel would catch the exception, or your front controller, or your plain old index.php) and a generic 404 error message could be rendered.

With the current propel query interface our controller action might look like this:
```php
public function show($id) {
    $post = PostQuery::create()->findOneById($id);
    if ($post === null) {
        throw new \NotFoundException('Post not found');
    }
   
    return $post;
}
```

With the new interface this could look like this:
```php
public function show($id) {
    return PostQuery::create()->filterById($id)->findOneOrFail();
}
```
```yml
propel:
  generator:
    exceptionClass: NotFoundException # Propel should also provide a default ResourceNotFoundException out of the box
```

You can see that the "new" controller is very elegant and does not has to worry about the error handling.

---

I'd appreciate any feedback on this idea, so let me know what you think about this :)

vworldat:
Totally forgot to answer once more:

> That's true. But as suggested by @Celc we could add a feature flag for this, so propel1 users can use the old behaviour.

This would split all propel dependent code (think libraries/bundles/plugins) into pre-BC and post-BC (feature on/off), since hardly anything would work in both situations. No, if we have to break, we should just do it and get it over with. I'm just not convinced this break is needed at all, since the non-breaking wording (`requireOne`) sounds much better to me.

mpscholten:
> This would split all propel dependent code

That's true. I think we should just go with the require api now because of bc. Maybe we can do this for propel 3 ... :)

PR #841, opened 2014-12-10
require-api: requireOne, requireOneByXXX, requireByArray

mpscholten:
Implementation of #803 

This adds `requireOne()`, `requireOneBy($column, $value)` and `requireOneByArray($values)` to `ModelCriteria`. The `ModelCriteria` also provides magic methods for `requireOneByXXX` and `requireOneByXXXAndXXX`. The generated query classes have `@method`-annotations for these magic methods to support autocompletion. 

---

You can now do something like:

```php
<?php
// ...
try {
  echo 'Hello ' . UserQuery::create()->requireOneById($_GET['id'])->getName();
} catch (\Propel\Runtime\Exception\ModelNotFoundException $e) {
  echo $e->getMessage();
}
```

Which will output 'Hello Marc' or 'User could not be found'  depending on whenever the given user-id exists.

---

You can customize the exception class in the propel config:
```yaml
propel
    generator:
        objectModel:
            notFoundExceptionClass: \Vendor\Project\ResourceNotFoundException
```
or by overwriting the `modelNotFoundExceptionClass` attribute in your query class (or you can overwrite the `createModelNotFoundException` method to customize your exception even more, e.g. when you use a custom constructor signature)

### DIFF
--- a/src/Propel/Common/Config/PropelConfiguration.php
+++ b/src/Propel/Common/Config/PropelConfiguration.php
@@ -326,6 +326,7 @@ class PropelConfiguration implements ConfigurationInterface
                                 ->scalarNode('classPrefix')->defaultNull()->end()
                                 ->booleanNode('useLeftJoinsInDoJoinMethods')->defaultTrue()->end()
                                 ->scalarNode('pluralizerClass')->defaultValue('\Propel\Common\Pluralizer\StandardEnglishPluralizer')->end()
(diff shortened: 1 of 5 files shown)


marcj:
Excellent @mpscholten! When you are gonna squash those commits I'm going to merge this beauty. :smile:

mpscholten:
When the discussion in #803 is finished :) But I think we will just take the `require` syntax.

I haven't rebased yet because I already use this feature and the rebase would change the commit-hash my code is requiring in the composer lock file. But here we go! :)

marcj:
Thanks, great job :)

PR #842, opened 2014-12-10
Fixes #715 IDE warning cast in generated crosslink methods

### LINKED ISSUE #715: IDE warning cast in generated crosslink methods.
When I have a table crosslinked with another, the `propel model:build` generate the corresponding classes for the tables and the query with methods for access to the linked fields like `addXXX` or `filterByXXX`.
The base abstract table class (`Base/TableTest`) and base abstract query class (`Base/TableTestQuery`) use in the code the reference of the main class `/TableTest` (in the phpdoc and checking the parameter with `instanceof`) and not the `Base/TableTest`.
This cause some warning in the IDE because the class `Base/TableTest` is casted to `/TableTest`.

nymo:
#715 
I added the base class as a second param in the phpdoc. I did this because when I changed the typehint of the passed parameter to be our base class then the IDE has thrown another warning (see picture). I couldn't find any clear method to get the base class so I created a new method in AbstracOMBuilder. If there is a better way please let me know.

warning

### DIFF
--- a/src/Propel/Generator/Builder/Om/AbstractOMBuilder.php
+++ b/src/Propel/Generator/Builder/Om/AbstractOMBuilder.php
@@ -147,6 +147,22 @@ abstract class AbstractOMBuilder extends DataModelBuilder
     }
 
     /**
+     * Returns the qualified base classname (e.g. Model\Base\Book)
+     *
+     * @return string
+     */
+    public function getFullyQualifiedBaseClassName()
+    {
+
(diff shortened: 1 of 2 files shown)


marcj:
Mh, I don't see code-changes that changes the type-hint of this first argument?

`public function add".$this->getRefFKPhpNameAffix($refFK, false)."($className \$l)` and `$className` is the same as before?

marcj:
ping @nymo

PR #845, opened 2014-12-14
Fix #840 Migration of unique/index keys in PgSQL

### LINKED ISSUE #840: Propel diff always generates change of unique/index keys
`propel diff` always generate change of unique/index keys, even if there's no change.
It's because `SchemaReader` passes XML attributes of the `unique-column` so `IndexComperator` compares size of the `unique-column` (which is mostly `null`) with the size of indexed columns.

I've [tried](https://github.com/propelorm/Propel2/pull/839) to fix that but without any success.

xxdavid:
My second try to fix issue #840.


marcj:
Ok, thanks @xxdavid! I'm going to merge this and add some tests later on.

PR #844, opened 2014-12-15
Add an API for exists().

Celc:
If I want to check if something exists currently I would do something like this:

```php
CmsCompanyQuery::create()->count() > 1;
```

Executes

````sql
SELECT COUNT(*) FROM cms_company;
````

But `count()` is expensive on larger tables, so I'd need to keep in mind to limit it:

````php
CmsCompanyQuery::create()->limit(1)->count() > 1;
````
Which surprisingly produces
```sql
SELECT COUNT(*) FROM (SELECT cms_company.com_id, cms_company.media_id, cms_company.admin_id, cms_company.com_name, cms_company.com_email, cms_company.com_address, cms_company.com_address2, cms_company.com_postal, cms_company.com_country, cms_company.com_employee, cms_company.com_skin, cms_company.com_theme, cms_company.com_activity, cms_company.com_status, cms_company.com_created, cms_company.com_modified, cms_company.com_creator, cms_company.pay_id, cms_company.pay_amount, cms_company.com_last_payment, cms_company.com_next_payment, cms_company.is_reward, cms_company.vote_start, cms_company.vote_end, cms_company.is_banner, cms_company.is_goal, cms_company.skin_path FROM cms_company LIMIT 1) propelmatch4cnt
```
Limiting the select column
```php
CmsCompanyQuery::create()->select(['com_id'])->limit(1)->count();
```

Turns into:

```sql
SELECT COUNT(*) FROM (SELECT cms_company.com_id AS "com_id" FROM cms_company LIMIT 1) propelmatch4cnt
```

Which is still not ideal because it's quite convoluted and the SQL query is still getting table data unnecessarily. A faster way would be to add an `exists()`:

```php
CmsCompanyQuery::create()->exists();
```
Produce.
```sql
SELECT 1 FROM cms_company LIMIT 1;
```
Which should be the fastest query to check for existence since we aren't interested in retrieving any data.

### DIFF
--- a/src/Propel/Runtime/ActiveQuery/ModelCriteria.php
+++ b/src/Propel/Runtime/ActiveQuery/ModelCriteria.php
@@ -1250,6 +1250,39 @@ class ModelCriteria extends BaseModelCriteria
     }
 
     /**
+     * Issue an existence check on the current ModelCriteria
+     *
+     * @param ConnectionInterface $con an optional connection object
+     *
+     * @return bool column existence
+     */
+    public function exists(ConnectionInterface $con = null)
+    {
+        if (null === $con) {
+            $con = Propel::getServiceContainer()->getReadConnection($this->getDbName());
+        }
+
+        $this->basePreSelect($con);
+        $criteria = $this->isKeepQuery() ? clone $this : $this;
+        $criteria->setDbName($this->getDbName()); // Set the correct dbName
+        $criteria->clearOrderByColumns(); // ORDER BY will do nothing but slow down the query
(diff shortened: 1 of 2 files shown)


marcj:
Thanks @Celc! We should also add a note in our documentation.

staabm:
@Celc using a query like `SELECT 1 FROM cms_company LIMIT 1;` will always return 1 no matter if the table has rows or not, doesn't it?

marcj:
@staabm no, it doesn't if the result set is empty.

PR #852, opened 2015-01-06
Improved error message when external schema cannot be found

mpscholten:
This improves the error message when you include an external schema which could not be found by propel. Currently you get the following error:

```
Warning: DOMDocument::load(): Empty string supplied as input
```

Now you get

```
Unable to find external schema 'my-external-schema.xml'
```

Propel applies `realpath` to the path given by the user. If `realpath` cannot resolve the given path it'll return `false`. Now `false` is given to `DOMDocument::load`, which logs a php warning because of invalid input.

### DIFF
--- a/src/Propel/Generator/Manager/AbstractManager.php
+++ b/src/Propel/Generator/Manager/AbstractManager.php
@@ -364,6 +364,10 @@ abstract class AbstractManager
 
             $externalSchema->parentNode->removeChild($externalSchema);
 
+            if (!is_readable($include)) {
+                throw new BuildException("External schema '$include' does not exist");
+            }
+
             $externalSchemaDom = new \DOMDocument('1.0', 'UTF-8');
             $externalSchemaDom->load(realpath($include));
 


src/Propel/Generator/Manager/AbstractManager.php
```diff
@@ -364,6 +364,11 @@ protected function includeExternalSchemas(\DOMDocument $dom, $srcDir)
 
             $externalSchema->parentNode->removeChild($externalSchema);
 
+            $path = realpath($include);
+            if (!$path) {
+                throw new BuildException("Unable to find external schema '$path'");
```
staabm: $include instead of $path

src/Propel/Generator/Manager/AbstractManager.php
```diff
@@ -364,6 +364,11 @@ protected function includeExternalSchemas(\DOMDocument $dom, $srcDir)
...
+            }
+
             $externalSchemaDom = new \DOMDocument('1.0', 'UTF-8');
             $externalSchemaDom->load(realpath($include));
```
staabm: $path instead of realpath($include)

src/Propel/Generator/Manager/AbstractManager.php
mpscholten: Thanks for pointing out :)

staabm:
Looks good, thx :+1:

src/Propel/Generator/Manager/AbstractManager.php
```diff
@@ -364,8 +364,13 @@ protected function includeExternalSchemas(\DOMDocument $dom, $srcDir)
 
             $externalSchema->parentNode->removeChild($externalSchema);
 
+            $path = realpath($include);
+            if (!$path) {
```
marcj: Why not just `file_exists` on `$include`? It resolves symlinks on its own.

staabm: Should we also check `is_readable`?

marcj: probably a good idea.

mpscholten: Doesn't `is_readable` without `file_exists` also works? `is_readable` basically includes `file_exists`

mpscholten:
> maybe add a unit test?

We don't have unit tests for this class at the moment. Not sure whether it's worth it to add them for this small change.

marcj:
I don't think its worth, but if you are incredible motivated, it won't hurt. :D

Sent from my iPhone

mpscholten:
Haha no. Just wanted to fix my php warning, and then thought this might be useful for other people to :)

mpscholten:
Just added `is_readable` as suggested. Could you take a look again?

src/Propel/Generator/Manager/AbstractManager.php
```diff
@@ -364,6 +364,10 @@ protected function includeExternalSchemas(\DOMDocument $dom, $srcDir)
 
             $externalSchema->parentNode->removeChild($externalSchema);
 
+            if (!is_readable($include)) {
+                throw new BuildException("Unable to find external schema '$include'");
```
staabm: "Either does not exist or is not readable"

mpscholten: Ok, just updated it

marcj:
Ok, cool, thanks! :smiley:

Issue #784, opened 2014-10-17
toJSON => fromJSON doesn't works


daniele-webmaori:
```php
/*
 * EXPORT
 */

$export = $form->toJSON();
// {"Id":16,"Name":"asd","AdminFrom":"asd","AdminSubject":"asd","AdminTo":"asd","CreatedBy":1,"CreatedAt":{"date":"2014-10-01 15:32:26","timezone_type":3,"timezone":"Europe\/Rome"},"ModifiedBy":1,"ModifiedAt":{"date":"2014-10-16 16:35:41","timezone_type":3,"timezone":"Europe\/Rome"},"Overwrote":false}

/*
 * IMPORT
 */

$form->fromJSON($export);

// Fatal error: Uncaught exception 'Exception' with message 'DateTime::__construct() expects parameter 1 to be string, array given' in /var/www/repo/kmaori/k-components/composer/propel/propel/src/Propel/Runtime/Util/PropelDateTime.php:90 Stack trace: #0 /var/www/repo/kmaori/k-components/composer/propel/propel/src/Propel/Runtime/Util/PropelDateTime.php(90): DateTime->__construct(Array) #1 /var/www/repo/kmaori/k-includes/models/Base/Form.php(666): Propel\Runtime\Util\PropelDateTime::newInstance(Array, NULL, 'DateTime') #2 /var/www/repo/kmaori/k-includes/models/Base/Form.php(1412): Base\Form->setCreatedAt(Array) #3 /var/www/repo/kmaori/k-includes/models/Base/Form.php(1444): Base\Form->fromArray(Array, 'phpName') #4 /var/www/repo/kmaori/k-includes/models/Base/Form.php(2459): Base\Form->importFrom('JSON', '{"Id":16,"Name"...') #5 /var/www/repo/kmaori/k-plugins/form/lib/FormQuery.php(21): Base\Form->__call('fromJSON', Array) #6 /var/www/repo/kmaori/k-plugins/form/lib/FormQuery.php(21): Form->fromJSON('{"Id":16,"Name"...') #7 /var/ in /var/www/repo/kmaori/k-components/composer/twig/twig/lib/Twig/Template.php on line 294


```

fitefoe:
Thank you for the fix, that works very well. In my application a timestamp is serialized like this: `2014-12-31T20:30:00+0100`

For me it would be better, if the timestamp is returned in UTC rather than a local time.
[...]

PR #853, opened 2015-01-06
Changed toArray to format temporal columns as ISO8601 strings

mpscholten:
Second fix for #784


marcj:
Good job @mpscholten, thanks!

PR #855, opened 2015-01-11
Add method requirePk() to Query classes

gharlan:
Thanks for the require* methods! :+1: 

I would like to have a `requirePk` method (like `findPk`). Is there a reason why this method was not added?


mpscholten:
Looks great, thanks :)

marcj:
thanks

PR #859, opened 2015-01-13
NestedSetBehavior improvements

gharlan:
* type hints
* better doc comments
* code format
* using `->exists()` instead of `->count() > 0`
* `getFirstChild` and `getLastChild`: Returning `null` instead of `array()` for leaves


marcj:
Nice stuff, thanks!

PR #864, opened 2015-01-20
Added support for polymorphic relations. Fixes #487.

### LINKED ISSUE #487: Add support for conditional FK-s
I'd like to propose feature to add extra conditions for foreign keys, e.g:

```xml
<table name="object_log_entry" description="Table which stores object changes">
  <column name="object_type" type="varchar" size="50" required="true" />
  <column name="object_id" type="bigint" required="true" />
  ...

  <foreign-key foreignTable="user" onDelete="cascade" onUpdate="cascade">
    <reference local="object_type" expression="USER" />
    <reference local="object_id" foreign="id" />
  </foreign-key>
</table>
```

Which would create something like:
```sql
...
FROM
  user u JOIN object_log_entry ole ON ole.object_type='USER' AND ole.object_id=u.id
...
```

Of course, this would imply that database-level FK is not created, cascading and stuff would be done by model.

marcj:
#487

### DIFF
--- a/src/Propel/Generator/Behavior/AggregateColumn/AggregateColumnBehavior.php
+++ b/src/Propel/Generator/Behavior/AggregateColumn/AggregateColumnBehavior.php
@@ -12,6 +12,7 @@ namespace Propel\Generator\Behavior\AggregateColumn;
 
 use Propel\Generator\Builder\Om\ObjectBuilder;
 use Propel\Generator\Model\Behavior;
+use Propel\Generator\Model\ForeignKey;
 
 /**
  * Keeps an aggregate column updated with related table
@@ -95,9 +96,15 @@ class AggregateColumnBehavior extends Behavior
 
         $bindings = array();
         $database = $this->getTable()->getDatabase();
-        foreach ($this->getForeignKey()->getColumnObjectsMapping() as $index => $columnReference) {
-            $conditions[] = $columnReference['local']->getFullyQualifiedName() . ' = :p' . ($index + 1);
-            $bindings[$index + 1]   = $columnReference['foreign']->getPhpName();
+
+        if ($this->getForeignKey()->isPolymorphic()) {
+            throw new \InvalidArgumentException('AggregateColumnBehavior does not work with polymorphic relations.');
+        }
+
+        foreach ($this->getForeignKey()->getMapping() as $index => $mapping) {
+            list($localColumn, $foreignColumn) = $mapping;
+            $conditions[] = $localColumn->getFullyQualifiedName() . ' = :p' . ($index + 1);
+            $bindings[$index + 1]   = $foreignColumn->getPhpName();
         }
         $tableName = $database->getTablePrefix() . $this->getParameter('foreign_table');
         if ($database->getPlatform()->supportsSchemas() && $this->getParameter('foreign_schema')) {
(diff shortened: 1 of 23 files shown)


src/Propel/Generator/Behavior/AggregateColumn/AggregateColumnBehavior.php
```diff
@@ -95,9 +96,15 @@ protected function addObjectCompute(ObjectBuilder $builder)
 
         $bindings = array();
         $database = $this->getTable()->getDatabase();
-        foreach ($this->getForeignKey()->getColumnObjectsMapping() as $index => $columnReference) {
-            $conditions[] = $columnReference['local']->getFullyQualifiedName() . ' = :p' . ($index + 1);
-            $bindings[$index + 1]   = $columnReference['foreign']->getPhpName();
+
+        if ($this->getForeignKey()->isPolymorphic()) {
+            throw new \InvalidArgumentException(sprintf('AggregateColumnBehavior does not work with polymorphic relations.'));
```
mpscholten: The `sprintf` call can be removed here

src/Propel/Generator/Model/ForeignKey.php
```diff
@@ -778,6 +792,56 @@ public function isComposite()
...
+    /**
+     * Whether this relation is a polymorphic association.
+     *
+     * At least one reference with a expression attribute set.
+     *
+     * @return boolean
+     */
+    public function isPolymorphic()
+    {
+        return (boolean) array_filter($this->localValues, function($value){
```
mpscholten: There should be a space after `function($value)`

src/Propel/Generator/Model/ForeignKey.php
```diff
@@ -778,6 +792,56 @@ public function isComposite()
...
+    /**
+     * Whether this relation is a polymorphic association.
+     *
+     * At least one reference with a expression attribute set.
+     *
+     * @return boolean
+     */
+    public function isPolymorphic()
+    {
+        return (boolean) array_filter($this->localValues, function($value) {
```
staabm: Wouldnt a simple `foreach` be easier here?

marcj:
Thanks guys!

src/Propel/Generator/Model/ForeignKey.php
```diff
@@ -625,9 +640,8 @@ public function getColumnObjectsMapping()
      */
     public function getMappedForeignColumn($local)
     {
-        $m = $this->getLocalForeignMapping();
-
-        return isset($m[$local]) ? $m[$local] : null;
+        $index = array_search($local, $this->localColumns);
+        return $this->foreignColumns[$index];
```
staabm: @marcj  

should this return null when `$local` could not be found using `array_search` instead of issue a php warnig because of a undefined index `$this->foreignColumns[$index]`?

same below in `getMappedLocalColumn()`

marcj: No, if it can't be found, it's a internal error and should trigger warnings/errors. returning null would be wrong.

staabm: but then we should throw a exception instead of implicitly rely on the warning? This is public api, so it should help the user to find the cause of errors

marcj: Feel free. I don't think it's necessary as it's only used in internal code.

PR #867, opened 2015-01-21
Added "typeHint" to column element and "interface" to foreign-key elemen...

### LINKED ISSUE #84: [RFC] Add `typehint` attribute on `Column` and `Table`
Hi,

I started a new behavior for Propel 1.6 named [**Typehintable**](https://github.com/willdurand/TypehintableBehavior) and I thought that should be nice to provide a native solution for the following use case: to allow users to typehint generated code in order to be compliant with _interfaces_.

### Why ?

* Because it's possible to type hint methods, and to implement interfaces using proxy classes is almost always a pain;
* Because it's cool to implement interfaces;
* Because it's cool to type hint methods.

Feel free to add more use cases :)

### What ?

We should be able to typehint the following methods:

* `setXXX()`
* `addXXX()`
* `removeXXX()`

We should allow PHP interfaces and primitive types (actually `array`).

### How ?

By adding new attributes to the `column` and `table` tags:

``` xml
<column name="roles" type="array" typehint="array" />
```

Will generate:

``` php
<?php

public function setRoles(array $roles)
{
    // ...
}
```

Or:

``` xml
<table name="group" typehint="\GroupInterface">
```

A `typehint` on a `table` tag will allow to typehint `addGroup()`, `removeGroup()` methods on related tables (in a N-N relation for instance). That way, you'll deal with relations, which is the most common use case.

``` php
<?php

// in Model/User.php for instance

public function addGroup(\GroupInterface $group)
{
    // ...
}
```

Oh, and it will also modify the phpdoc.

The `typehint` has to be an interface or `array`, if it's not the case, you'll throw an exception at build time.

Thoughts?

William

marcj:
...t. Fixed #84.

This basically implements the feature set from https://github.com/willdurand/TypehintableBehavior.

### DIFF
--- a/resources/dtd/database.dtd
+++ b/resources/dtd/database.dtd
@@ -103,6 +103,7 @@ PHP class or method name.
   inheritance (single|false) "false"
   phpNamingMethod (nochange|underscore|phpname) #IMPLIED
   description CDATA #IMPLIED
+  typeHint CDATA #IMPLIED
   lazyLoad (true|false) "false"
   primaryString (true|false) "false"
   valueSet CDATA #IMPLIED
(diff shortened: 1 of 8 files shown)


staabm:
nice!

PR #868, opened 2015-01-21
groupByArray

nibsirahsieu:
Currently, if we want to do grouping by multiple columns, we have to write *groupBy* for each column. For example: 
```php
FooBarQuery::create()
   ->joinBaz()
   ->select(['Foo', 'Bar'])
   ->withColumn('COUNT(Baz.Id)', 'nbBazs')
   ->groupBy('Foo')
   ->groupBy('Bar')
```

with this PR, we just simply write:
```php
FooBarQuery::create()
   ->joinBaz()
   ->select(['Foo', 'Bar'])
   ->withColumn('COUNT(Baz.Id)', 'nbBazs')
   ->groupByArray(['Foo', 'Bar'])
```

### DIFF
--- a/src/Propel/Runtime/ActiveQuery/ModelCriteria.php
+++ b/src/Propel/Runtime/ActiveQuery/ModelCriteria.php
@@ -270,6 +270,26 @@ class ModelCriteria extends BaseModelCriteria
 
         return $this;
     }
+    
+    /**
+     * Adds a GROUP BY clause for certain columns to the query
+     * Examples:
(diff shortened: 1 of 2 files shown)


Celc:
I don't think you need to create a new method for this, just enable `groupBy` to accept an array, much like how `select` works,

staabm:
agree. please overlay `groupBy` instead of creating a new method.

marcj:
Jupp

PR #871, opened 2015-01-27
Fixed setter for objects columns

IceShack:
Compare serialized object against stream content, instead of the stream handle


marcj:
Thanks! :smile:

PR #872, opened 2015-01-29
Added polimorphic joining ability from one to many

apinnecke:
- Fixed setup.mysql.sh to use DB_PW EV Param
- Added some more LogicExceptions on with-usage and limit


marcj:
Cool, thx!

PR #873, opened 2015-02-03
fix typehining on generated doc blocks by adding default 'mixed'' value

apinnecke:
On columns without explicit type hinting, like PropelType::Object, no param type was set:

```php
/**
     * Set the value of [sms_per_day] column.
     *
     * @param $v new value
     * @return \Valid\NameSpace\Model The current object (for fluent API support)
     */
```

 Leading to wrong IDE assumption for param type equals `Base\new`. By adding this fix doc block generation results in:

```php
/**
     * Set the value of [sms_per_day] column.
     *
     * @param $v mixed new value
     * @return \Valid\NameSpace\Model The current object (for fluent API support)
     */
```


staabm:
is this the only place where we have this error?

apinnecke:
Thanks for the hint @staabm, i added it to the default accessor as well after having the same issue there. Most of the other occurrences are using specific code for generating type hinting.

staabm:
looks good to me, thanks.

marcj:
Thanks :smile:

PR #876, opened 2015-02-08
Fix #865: set limit to -1 in clear method

### LINKED ISSUE #865: ActiveQuery\Criteria::clear function sets limit to 0 instead of -1
Hi,
the clear function indicates that it is used to bring the criteria back to its initial state, but the limit is set to 0 instead of -1.
So when reusing the QueryObject without setting the limit, it always returns an empty data set.

apinnecke:
This fixes propelorm/Propel2#865.


marcj:
Thanks :)

PR #879, opened 2015-02-09
Fixed #711 by throwing an excetion if one try to add a foreign key twice

### LINKED ISSUE #711: Try to catch name collisions
When you define multiple foreign keys to the same table without specifying a different `phpName` then you should get a exception that tells you you have to name at least one of those foreign keys different to not generate naming collisions.

For example: http://stackoverflow.com/questions/12897278/multiple-function-declaration-inside-base-queryafter-propel-class-rebuild

At the moment you just get compile errors from PHP itself, which is not cool at all, because you don't know where it comes from.

apinnecke:
Was a hell of work fixing all the tests -.-


marcj:
Cool, thanks :)

PR #882, opened 2015-02-16
Fix performance issues on OnDemandFormatter as in ObjectFormatter

apinnecke:
Thus we don't fill our local instance collection inside the OnDemandFormatter we don't get into filling memory issues over time.


marcj:
awesome =)

PR #884, opened 2015-02-20
`ModelCriteria::clear()` resets the `$select` value

mlebkowski:
Please merge if it’s the desired behaviour. There is no other way of reseting the `select` field, since the `select()` method doesn’t allow `null`.

### DIFF
--- a/src/Propel/Runtime/ActiveQuery/ModelCriteria.php
+++ b/src/Propel/Runtime/ActiveQuery/ModelCriteria.php
@@ -779,7 +779,8 @@ class ModelCriteria extends BaseModelCriteria
 
         $this->with = array();
         $this->primaryCriteria = null;
-        $this->formatter=null;
+        $this->formatter = null;
+        $this->select = null;
 
         return $this;
     }


staabm:
IMO clearing models for re-use is most of the time a error prone thing, which doesnt save much.
creating a model from scratch using `new` works much better.

@marcj do you agree that dropping methods like `clear` in propel2 would be a good thing?

mlebkowski:
Don’t mind the code, it’s just an example. In my case, I have something like:
```php
$ids = $this->query->clear()->select('id')->limit(10)->find();
foreach ($ids as $id) {
	if (something()) {
		$item = $this->query->clear()->findPk($id);
```

I am not creating new query class since I’d like to have it injected. If `clear()` is a bad solution, I’d have to clone it:
```php
private function getQuery() {
	return clone $this->query;
}
```

Would this be an elegant solution in your opinion?

OTOH I can use a factory:
```php
private function getQuery() {
	return call_user_func($this->queryFactory);
}
```

staabm:
> I am not creating new query class since I’d like to have it injected

why do you want that? a model class is not something I would consider for DI.

Nevertheless, using a method which returns a clone of a fresh instance should work like using `new` though.

mlebkowski:
why not? I tend to use `new` only for models, value objects and exceptions ;) 

In this case I needed to test it, and the easiest way was to mock the query class. I’m sticking with cloning for now.

marcj:
> @marcj do you agree that dropping methods like clear in propel2 would be a good thing?

marcj:
> @marcj do you agree that dropping methods like clear in propel2 would be a good thing?

Actually not, since `clear()` is really a common method used in many use cases. We should stay with this method and rather make it useable for use cases @mlebkowski mentioned. So I'm going to merge this.

PR #889, opened 2015-03-01
Migration: Missing "AFTER"/"FIRST" when adding columns (MySQL)

IceShack:
- Copied code from Propel 1
- Changed methods, so the statements can get merged


marcj:
Thanks, looks pretty good :)

gharlan:
:+1: thanks!

robin850:
It looks like the test suite isn't covering the case with `FIRST`, no ?

IceShack:
@robin850 
Added a unit test for that. #909

Issue #894, opened 2015-03-11
Decimal with a zero scale 


kairichard:
When working with a legacy database that has field like the following

```mysql
  `field_name` decimal(11,0) DEFAULT NULL
```

propel generates a wrong schema.xml omitting the scale of the field

This can be reproduced using `make test` from https://github.com/kairichard/propel-issue-decimal-zero-scale where a second migration will be created because the schema does not match with the generated migration.

robin850:
I guess this can be closed as #897 has been merged.

PR #897, opened 2015-03-12
Fix reverse command on decimal field with 0 as scale

kairichard:
This fixes issue #894


marcj:
I'm not sure about that since `0` is the default for MySQL, so we can actually omit it when its empty.

kairichard:
Yes, it's the default, but for legacy databases where there is a field defined with `DECIMAL(11, 0)` it is needed in order to get valid schema. See https://github.com/kairichard/propel-issue-decimal-zero-scale for the isolated issue.

marcj:
ok, thanks!

Issue #895, opened 2015-03-11
Wrong type definition in generated migration


kairichard:
When working with a legacy database that has field like the following

```mysql
  `field_name` decimal(10,3) DEFAULT '1.000' NOT NULL
```

propel generates a wrong migration omitting the `10,3` in the type definition

This can be reproduced using `make test` from https://github.com/kairichard/propel-issue-decimal-default where a second migration will be created because the schema does not match with the generated migration.

robin850:
It looks like this can be closed as well since #898 has been merged.

PR #898, opened 2015-03-13
Fix default type size with scale

kairichard:
This fixes issue #895

### DIFF
--- a/src/Propel/Generator/Platform/DefaultPlatform.php
+++ b/src/Propel/Generator/Platform/DefaultPlatform.php
@@ -1400,7 +1400,7 @@ if (is_resource($columnValueAccessor)) {
 
         foreach ($table->getColumns() as $column) {
             if ($column->getSize() && $defaultSize = $this->getDefaultTypeSize($column->getType())) {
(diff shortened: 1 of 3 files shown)


src/Propel/Generator/Platform/DefaultPlatform.php
```diff
@@ -1400,7 +1400,7 @@ public function normalizeTable(Table $table)
 
         foreach ($table->getColumns() as $column) {
             if ($column->getSize() && $defaultSize = $this->getDefaultTypeSize($column->getType())) {
-                if (intval($column->getSize()) === $defaultSize) {
+                if (null === $column->getScale() && intval($column->getSize()) === $defaultSize) {
```
marcj: Shouldn't this be a logical OR?

kairichard: `getScale` can still return `null` while `getSize` returns the `defaultSize` and only then can the `size` be overriden to `null`

marcj: aah,I see :)

marcj:
Thanks @kairichard :)

Issue #900, opened 2015-03-20
Propel is not recording milliseconds on MySQL/MariaDB timestamp(3) column


dmelo:
The schema.xml row that maps my timestamp columns is:

    <column name="time" phpName="Time" type="TIMESTAMP" size="3" required="true" defaultValue="0000-00-00 00:00:00.000"/>

It is mapping, of course, to a DateTime object. I'm filling the microseconds info on the object but, when I save the row, it always ends with ".000", indicating Propel is most likely truncating my object.

dereuromark:
I see some commits. Have any of those reached master?
Is someone able to make a PR here with suggested changes?

PR #902, opened 2015-03-20
Fix issue #900 by changing the timestamp formatter

dmelo:
This change allow register of timestamp on microseconds level.

### DIFF
--- a/src/Propel/Generator/Builder/Om/ObjectBuilder.php
+++ b/src/Propel/Generator/Builder/Om/ObjectBuilder.php
@@ -1611,7 +1611,7 @@ abstract class ".$this->getUnqualifiedClassName().$parentClass." implements Acti
         if (($def = $col->getDefaultValue()) !== null && !$def->isExpression()) {
             $defaultValue = $this->getDefaultValueString($col);
             $script .= "
-            if ( (\$dt != \$this->{$clo}) // normalized values don't match
+            if ( (\$dt->format($fmt) != \$this->{$clo}->format($fmt)) // normalized values don't match
                 || (\$dt->format($fmt) === $defaultValue) // or the entered value matches the default
                  ) {";
         } else {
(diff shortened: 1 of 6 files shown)


staabm:
could you please add a unit test to cover the change?

do we have a similar issue in other DB plattform classes?

dmelo:
I think the problem will be the same on other DBs. I will write the unit tests and verify other DB platforms, this weekend.

marcj:
Thanks @dmelo, that sounds good :)

dmelo:
So far, I've got the unit test for for MySQL and the solution with impact only on MySQL. It seems other databases have the same problem.

I will look further. If other dbs have indeed the same problem, it might be better if I send it on different pull requests.

dmelo:
I'm checking the failures on CircleCI.

dmelo:
The problem on node 0 (db agnostic) is that I have changed how two datetimes are compared on src/Propel/Generator/Builder/Om/ObjectBuilder.php . I did it so because PHP considers two objects that differ only by microseconds to be equal. So, working on Propel, your column value (on a specific row) is '2014-01-01 00:00:00.123456' and you want to change it to '2014-01-01 00:00:00.444444', you can't. Because when you ask to save it, Propel sees no modified object and don't bother updating the row.

PHP [doesn't fully implement fractional seconds](http://php.net/manual/en/class.datetime.php#108970). I'm not sure what is the best way to solve this. 

Another issue is that datetime with fractional seconds [have been introduced in MySQL 5.6.4](http://dev.mysql.com/doc/refman/5.6/en/fractional-seconds.html). That is the probably the problem on CircleCI container 1, the node that handles MySQL. I think there is no magic here. Microseconds will only work if the database supports it. Therefore, while working with previous versions of MySQL, fractional seconds won't be available.

I see that a similar solution (to make something available for only specific version of a db) have been employed for foreign key on SQLite. See src/Propel/Generator/Platform/SqlitePlatform.php, line 34.

The other two nodes failed because I didn't restrict the tests to MySQL. I've just submitted the commit that fix  this

I'm working on the first two problems that I have described. Suggestions would be very helpful.

src/Propel/Generator/Builder/Om/ObjectBuilder.php
```diff
@@ -1611,7 +1611,7 @@ protected function addTemporalMutator(&$script, Column $col)
         if (($def = $col->getDefaultValue()) !== null && !$def->isExpression()) {
             $defaultValue = $this->getDefaultValueString($col);
             $script .= "
-            if ( (\$dt != \$this->{$clo}) // normalized values don't match
+            if ( (\$dt->format($fmt) != \$this->{$clo}->format($fmt)) // normalized values don't match
```
marcj: guess you need here a null check

marcj:
Ok cool, to get the test suite running you can probably just check in the unit-test if the correct MySQL version is running using a `SHOW VARIABLES LIKE "version"` query.

dereuromark:
ping @dmelo 
Can you also rebase on top of latest master?

dereuromark:
Already in master it seems

PR #916, opened 2015-04-09
Fix #915  serialize custom properties in __sleep

### LINKED ISSUE #915: Properties of child classes lost after serialize->unserialize
An object that is serialized using sleep looses instance properties that are declared in child classes:

```php
class Book extends Base\Book
{
  private $color;
  public function setColor($color) { $this->color = $color; }
  public function getColor() { return $this->color; }
}

$book = new Book();
$book->setColor('blue');

echo 'Before: '.$book->getColor()."\n";
$unserializedBook = unserialize(serialize($book));
echo 'After: '.$unserializedBook->getColor()."\n";
```
Output
```
Before: blue
After:
```

I would expect that the property would be restored by default, similar to PHP's default behavior.
[baseObjectMethods.php:209](https://github.com/propelorm/Propel2/blob/fb7aa33479797a5e2fe794eeb3762691cb6416b2/src/Propel/Generator/Builder/Om/templates/baseObjectMethods.php#L209) is causing this problems. `get_object_vars` is scope aware and does not return the attributes of child classes. [ReflectionClass::getProperties](http://php.net/manual/en/reflectionclass.getproperties.php) should work as expected here.

david0:
I did not find a good way to place the test fixture, so i included it in the testcases file.

### DIFF
--- a/src/Propel/Generator/Builder/Om/templates/baseObjectMethods.php
+++ b/src/Propel/Generator/Builder/Om/templates/baseObjectMethods.php
@@ -206,5 +206,10 @@
     {
         $this->clearAllReferences();
 
-        return array_keys(get_object_vars($this));
+        $cls = new \ReflectionClass($this);
+        $propertyNames = [];
+        foreach($cls->getProperties() as $property) {
+            $propertyNames[] = $property->getName();
+        }
+        return $propertyNames;
     }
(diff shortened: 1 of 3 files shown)


marcj:
Well the test suite is pretty much red :smile:

david0:
Yes, it was a stupid idea to use the bookstores test fixture.
Now its green.

marcj:
Awesome, thanks :)

PR #918, opened 2015-04-12
Fix PostgreSQL's default port running the init command

robin850:
Hello,

This pull request fixes PostgreSQL's default port number to 5432 when running the `propel init` command. 3306 is actually MySQL's default port number.

http://www.postgresql.org/docs/9.4/static/runtime-config-connection.html

Have a nice day.


marcj:
thanks :smile:

Issue #151, opened 2012-03-13
QUERY methods are not delegated


wurststulle:
in the delegation behaviour, the QUERY methods are not delegated.

when using

```xml

<table name="account">
  <column name="id" required="true" primaryKey="true" autoIncrement="true" type="INTEGER" />
  <column name="login" type="VARCHAR" required="true" />
  <column name="password" type="VARCHAR" required="true" />
  <behavior name="delegate">
    <parameter name="to" value="profile" />
  </behavior>
</table>
<table name="profile">
  <column name="email" type="VARCHAR" />
  <column name="telephone" type="VARCHAR" />
</table>

```
you can not do a accountQuery::create()->filterByEmail('foo')->find()

(copied from https://github.com/propelorm/Propel/issues/157)

SCIF:
@wurststulle , you can to try patch in #919, it must solve your issue

solidevolution:
Since the merge of this commit: 

https://github.com/SCIF/Propel2/commit/06914b7a4c688d3ce86b27e651e19e8be9b5afbf

we have the following problem with "php propel build"

 [Propel\Runtime\Exception\PropelException]                                                                                          
  Column with name «created_at» (delegated from table «user_payment») already exists in table «user». Probably database design mista  
  ke    

I wondered because in our composer.json we decide to take version "2.0.0-alpha" ...

SCIF:
Field is double defined i.e. for example: created_at defined in parent and delegated table. How Query class can detect which table field you want to filter?

solidevolution:
Yes, that's what we want. A created_at & updated_at timestamp on every join table.

It was possible in past so we designed it like that.

marcj:
maybe we should just skip those columns which are double defined like before @SCIF?

SCIF:
But it's not correct, i think. What way of «overlap» detection do you suppose? What field it must filter if you will use multiple delegations with created_at/updated_at?

```
  <behavior name="delegate">
    <parameter name="to" value="profile, preference" />
  </behavior>
```

It will causes unpredictable behavior or just skip that fields.

SCIF:
@marcj was some faster than me. :) Yeah, we can skip double defined. Really don't know what is better — BC or correct design forcing :)

solidevolution:
I understand your fears @SCIF but then you need another behaviour where it's possible to use default timestampable on multiple 1:1 tables. Let me explain what we do/want:

        <behavior name="delegate">
[...]

PR #919, opened 2015-04-13
Some improvements of DelegateBehavior

SCIF:
Implemented filtering and orderBy*() by delegated table properties. Also, toArray() now returns delegated properties too.


SCIF:
#151 must be solved by current PR

marcj:
Good job, thanks!

Issue #920, opened 2015-04-13
slaves configuration


jeremydenoun:
In Propel/Common/Config/PropelConfiguration.php:137 you authorize only dsn nodes but we must provide login/pass to PdoAdapter for use it..

I propose to add : 
 	       	       	       	       	       	->scalarNode('user')->end()
	       	       	       	       	       	->scalarNode('password')->treatNullLike('')->end()
l.140 to fix it

jeremydenoun:
my patch permit to use same options than write connection in read (slaves)  connections ...
I do it for use Propel\Runtime\Connection\ProfilerConnectionWrapper for slaves but for be complete we should add a parameter in log for determine if request come from read or write connection. I'm work on it

marcj:
Guess this is fixed in https://github.com/propelorm/Propel2/pull/1146. Please give me a ping when not.

PR #921, opened 2015-04-13
custom connections options for slaves

jeremydenoun:
fix behavior described in : https://github.com/propelorm/Propel2/issues/920


marcj:
Well, actually this won't fix anything as those config values are not read by our service container or PropelBundle. A better approch is to do it the way PropelBundle handles it: It merges the user/password values into the slave array, so master=slave credentials. https://github.com/propelorm/PropelBundle/blob/2.0/PropelBundle.php#L72

jeremydenoun:
I check source code especially :
https://github.com/propelorm/Propel2/blob/master/src/Propel/Runtime/Connection/ConnectionManagerMasterSlave.php#L172

you use complete slave read configuration

your proposal is good but : 
- rarely slaves connection can have different login/password than master (and we can't force them in dsn)
- sometimes we can want specific "options/queries" for slaves database init and finally for profiling and stats we can use another Connection classname

so I think 2 approach are necessary 

Best

marcj:
Mh, then we should not mark those values as required. If they are not defined, they should be taken from the master connection. So, make your newly added options optional and merge those values if they are not defined with the master connection, then I'm ok with it and it addresses all issues. :)

jeremydenoun:
done in 96bac07

jeremydenoun:
I close this pull request because this is linked to my master branch... I will open another only with slave patch... FYI, I improve timestampable behavior and remove Symfony Legacy Validator (throw notice) in my branch

PR #923, opened 2015-04-21
External attribute added

SCIF:
Implemented handled of new schema.xml attribute «external» which force to skip marked table on `migration:diff` action.

### DIFF
--- a/src/Propel/Common/Config/PropelConfiguration.php
+++ b/src/Propel/Common/Config/PropelConfiguration.php
@@ -33,6 +33,7 @@ class PropelConfiguration implements ConfigurationInterface
         $rootNode = $treeBuilder->root('propel');
 
         $this->addGeneralSection($rootNode);
+        $this->addExcludeTablesSection($rootNode);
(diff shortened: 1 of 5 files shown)


marcj:
I'd rather want to see this through "--skip-table" option.

SCIF:
Explain please your point. I suggest mark tables in schema as skipped. How it's to --skip-tables option??

marcj:
schema.xml is the wrong place to put this information at. Better is to place it in the config.yml or pass those information as command option as stated `skip-tables`, which already exists.

SCIF:
Do you mean to add [here](http://propelorm.org/documentation/reference/configuration-file.html#database-settings) something like:
```yaml
skipped_tables:
  - table1
  - table2
```

marcj:
:+1:

SCIF:
Migration diff implemented.

marcj:
very nice :+1:

marcj:
Looks very good, thanks @SCIF, good work!

PR #931, opened 2015-05-05
PDO Statement Wrapper not correctly return getExecutedQueryString with parameter

jeremydenoun:
Hi,

I find a bug in wrapper, if we execute PDO Statement with input_parameters into a debug context, the workaround for fetch getExecutedQueryString is based on boundValues (and in this case they don't set).

I propose following pull request for support this (with a silent fallback)

Best


marcj:
Thanks!

PR #932, opened 2015-05-07
DelegateBehavior: double defined fields now skips instead of throwing…

SCIF:
… exception. Fixed bug appeared in #151.


marcj:
Thanks :)

Issue #949, opened 2015-06-19
Clone DateTimes before returning them from temporal column getters


rimas-kudelis:
There is a problem in how ActiveRecord currently treats temporal columns. In particular, changes to these columns may be left unnoticed. For example, let's say I have a "Start" column in my table, and I do something like this:

```
$start = $object->getStart();
$start->modify('+1 day');
$object->setStart($start);
$object->save();
```

After doing that, I would expect the DB record to be updated, but it is not. And the reason why it is not being updated seems to be the following:
- getStart() returns the original DateTime object, which is stored as a property of $object
- modify() modifies that same DateTime object, without $object noticing the change
- setStart() compares the dates and only changes field value and marks it as modified if the dates are different. Obviously, comparison between references to the same object will always yeld equality, so the column is not marked as modified.
- since no columns are marked as modified, save() does nothing.

(Note I'm actually using Carbon to work with dates, but it is based on DateTime, so I don't think it makes any difference in this case.)

A solution to this would be to either use DateTimeImmutable class, which is new to PHP 5.5, or to clone the DateTime before returning it from the getter. Assuming that Propel allows users to choose their own DateTime implementations for temporal columns, the latter seems to make more sense.

rimas-kudelis:
It seems that tests provided by #768 almost cover this, but not quite, which is probably why the problem had been left unnoticed.

rimas-kudelis:
Wow, apparently, my patch (#951) breaks another test which asserts that DateTime is returned as a reference to the original object.

Is this indeed by design? If so, is it really acceptable that ActiveRecord simply ignores changes to its properties, when they just happen to also be objects of some class? To me, this seems completely unexpected.

staabm:
This PR introduced the return-by-ref behaviour/test:
https://github.com/propelorm/Propel2/pull/87

rimas-kudelis:
Thanks!
Huh, #87 supposedly fixes #59, which has the following example of an intuitive API in the initial comment:
```
$foo->getDate()->modify('+1 days'); $foo->save();
```
which is exactly the thing that is currently not working... :)

[...]

PR #951, opened 2015-06-20
Fix issue #949: Clone DateTimes before returning them from temporal column getters.

rimas-kudelis:
DateTime class is mutable, so if returned without cloning, it might be modified without the ActiveRecord object ever noticing the change.

### DIFF
--- a/src/Propel/Generator/Builder/Om/ObjectBuilder.php
+++ b/src/Propel/Generator/Builder/Om/ObjectBuilder.php
@@ -898,7 +898,7 @@ abstract class ".$this->getUnqualifiedClassName().$parentClass." implements Acti
 
         $script .= "
         if (\$format === null) {
-            return \$this->$clo;
+            return \$this->$clo instanceof \DateTime ? clone \$this->{$clo} : null;
         } else {
             return \$this->$clo instanceof \DateTime ? \$this->{$clo}->format(\$format) : null;
         }";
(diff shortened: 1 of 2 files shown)


src/Propel/Generator/Builder/Om/ObjectBuilder.php
```diff
@@ -898,7 +898,7 @@ protected function addTemporalAccessorBody(&$script, Column $column)
 
         $script .= "
         if (\$format === null) {
-            return \$this->$clo;
+            return \$this->$clo instanceof \DateTime ? clone \$this->{$clo} : null;
```
staabm: Shouldnt it return a clone for datime and this->$clo else?

rimas-kudelis: Hm, makes sense. I assumed $this->{$clo} will always be either DateTime or null, but you're probably right, I can adjust the commit when we have a consensus of what's the best way to proceed in #949.

marcj:
Mh, there are failing tests? Would you mind to fix those? :)

dereuromark:
ping @rimas-kudelis

rimas-kudelis:
Uhm, not really. I haven't worked with Propel for years now.

dereuromark:
Closing as different way needs to be done here.

PR #953, opened 2015-06-24
Fixed #950. Now magic {left|right|inner}Join{Relation}() methods and …

### LINKED ISSUE #950: Self referencing *JoinRelatedBy* producing invalid syntax on MariaDB
I have a self-referencing table which I am trying to join.

    <table name="category">
        <column name="id" type="integer" primaryKey="true" autoIncrement="true"/>
        <column name="name" type="varchar" size="250" required="true" primaryString="true"/>
        <column name="parent" type="integer"/>

        <foreign-key foreignTable="category" onDelete="CASCADE" onUpdate="CASCADE">
            <reference local="parent" foreign="id"/>
        </foreign-key>
    </table>

When I try to join based on the foreign key like so:

    CategoryQuery::create()
        ->filterByParent(null)
        ->leftJoinCategoryRelatedByParent()
        ->find();

I get a syntax error due to the two table references needing to be distinct. i.e. the left table needs a different name/alias to the right table.

For example, I can manually achieve what I want with the below statement:

    SELECT root.id, root.name, child.id, child.name
    FROM category root
    JOIN category child 
    ON child.parent = root.id 
    WHERE root.parent IS NULL;

Sorry if this is an erroneous report, I am new to Propel.

SCIF:
…joinWith() works correct on self-referenced relations.

### DIFF
--- a/src/Propel/Generator/Builder/Om/QueryBuilder.php
+++ b/src/Propel/Generator/Builder/Om/QueryBuilder.php
@@ -1317,6 +1317,7 @@ abstract class ".$this->getUnqualifiedClassName()." extends " . $parentClass . "
     {
         \$tableMap = \$this->getTableMap();
         \$relationMap = \$tableMap->getRelation('" . $relationName . "');
+        \$relationAlias = \$relationAlias ?: '" . $relationName . "';
 
         // create a ModelJoin object for this join
         \$join = new ModelJoin();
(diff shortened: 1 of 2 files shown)


SCIF:
Heh, the most seems like it requires to update tests. @marcj , do you agree that i must just to update tests?

marcj:
It seems like you have made many tests red :smile: So I guess that's not the right way to implement it also because it would be a bigger BC break, wouldn't it?

Issue #960, opened 2015-07-01
paginate() (because of count()) EXTREMELY slow/inefficient (When using QueryCache).


EVILoptimist:
Ran into a performance issue today when I needed to use the ->paginate() functionality on a table with 30k+ records on MySQL. For a simple query, it appears as though Propel chooses to perform the count on a derived table instead of simply selecting COUNT(*) instead of the field list.

How I would query MySQL in this case: 0.292sec
```sql
SELECT COUNT(*) FROM  `items` propelmatch4cnt
```

How PropelORM chooses to query MySQL: 58.187sec
```sql
SELECT COUNT(*) FROM (SELECT <<LENGTHY LIST OF FIELDS>> FROM `items`) propelmatch4cnt
```

Considering that the ->paginate() functionality exists almost exclusively for dealing with large recordsets, this seems to be extremely counterintuitive.

EVILoptimist:
After looking into the Base classes, I can see that there is logic in place that is meant to execute a much faster query.

To clarify, this is the code being used that generated this query:
```php
$query = ItemQuery::create()->count();
```

EVILoptimist:
Additionally the count() method of the PropelModelPager returns the getLimit() value instead of the value that it took 58 seconds to retrieve :)

```php
$items = ItemQuery::create()->paginate(1, 10);
$items->count(); // int(10)
```

EVILoptimist:
OK. Looks like I found a simple solution after reviewing the way this is handled in other parts of the code.

Pull request submitted! :)

marcj:
Fixed through the submitted PR #961, thanks!

PR #961, opened 2015-07-02
Alter comparison so that an int(-1) value for getLimit() evaluates correctly

EVILoptimist:
Simple fix for #960. This one-liner updates this comparison to be consistent with other comparisons for this method.

### DIFF
--- a/src/Propel/Generator/Behavior/QueryCache/QueryCacheBehavior.php
+++ b/src/Propel/Generator/Behavior/QueryCache/QueryCacheBehavior.php
@@ -228,9 +228,11 @@ public function doCount(ConnectionInterface \$con = null)
 
         \$needsComplexCount = \$this->getGroupByColumns()
             || \$this->getOffset()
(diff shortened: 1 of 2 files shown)


SCIF:
Wow, so curious test failure. Leap second effect or just slow data insertion to db? Seems interesting.

staabm:
Any chanche to add a unit test supporting the change?

EVILoptimist:
Ohhh..... Hahaha.... Good idea. Yeah. Lemme take a look :)

EVILoptimist:
Copied the block from the /Propel/Runtime/ActiveQuery/Criteria class. Seems more complete and includes the fix I proposed.

EVILoptimist:
Wrote a test to count() a table. If it executes anything but a simple count query, test will fail.

staabm:
Looks good to me thx. :+1:

marcj:
Very nice, thanks @EVILoptimist!

Issue #962, opened 2015-07-02
DateTime field value may become incorrect when an object persisted to MySQL is later retrieved


rimas-kudelis:
When MySQL connection timezone differs from PHP's "current timezone", a DateTime value persisted to MySQL storage will differ from the DateTime later being retrieved. For example, my my.cnf contains the following line:
`default-time-zone='+0:00'`
however, in the PHP config, I have the following line:
`date.timezone = "Europe/Vilnius"`
With this config, when I store a DateTime representing the following date to MySQL:
`2015-01-01 01:00:00 UTC`
and later retrieve it, I get back a DateTime representing the following date instead:
`2015-01-01 01:00:00 Europe/Vilnius`
which differs from what I expected by three whole hours (Vilnius is +0300)!

This effect is even more interesting if I attempt to store the DateTime in a completely different timezone. For example, I just stored the following date to the Database:
`2015-10-01T21:15:00-0900`
But what I see in the databaseis the following:
`2015-10-01 21:15:00` // connection timezone is +0000 and the field is TIMESTAMP, thus the value differs by nine hours from the original by now.
And when I retrieve my object back, here's what I get instead of the expected timezone:
`2015-10-01T18:15:00Z` // differs by 12 hours from the original (the timezone is Z again, because toArray() converts datetime to UTC).

I see a few ways how this problem could be fixed. Since MySQL does not store timezone info in any of temporal fields, that info should be considered lost right at the beginning. Instead, I would concentrate on storing actually correct timestamps. So, here are the options I see:
1. convert all timestamps to the timezone used by MySQL connection before storing them, and convert them from that same timezone when retrieving them back.
2. use FROM_UNIXTIME() when storing datetime and UNIX_TIMESTAMP() when retrieving.
3. explicitly set connection timezone when connecting to the DB.

Not sure which one of these is best preferred. But in any way, I'm sure that this is a bug and it should be fixed.

rimas-kudelis:
A test case has been attached to #963.

marcj:
As this is a result of wrong configuration of PHP and MySQL which leads almost always also to other issues like displaying the wrong date in a view I'd recommend to trigger a warning if Propel detects in MySQL a different timezone than PHP is using. What do you think about that?

marcj:
Well my idea would also provide a lot of queries at the beginning. So I believe your option 2) is the best if that fixes all of the timezone issues.

rimas-kudelis:
Hi Marc,
[...]

PR #963, opened 2015-07-02
Added a test case for #962: DateTime field value may become incorrect when an object persisted to MySQL is later retrieved

rimas-kudelis:
#962

### DIFF
--- a/tests/Propel/Tests/Issues/Issue962Test.php
+++ b/tests/Propel/Tests/Issues/Issue962Test.php
@@ -1,0 +1,77 @@
+<?php
+/**
+ * This file is part of the Propel package.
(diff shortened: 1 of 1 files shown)


Incognito:
Fails on latest, but not in the expected way. 

```
1) Propel\Tests\Issues\Issue962Test::testTimeStampIsStoredCorrectly
TypeError: Argument 1 passed to Propel\Runtime\Connection\ConnectionFactory::create() must be of the type array, null given, called in /home/travis/build/spryker/Propel2/src/Propel/Runtime/Connection/ConnectionManagerSingle.php on line 76
```

Test will need minor changes.

gechetspr:

PR #967, opened 2015-07-13
Fix #954 - Sortable behavior on enum scope

### LINKED ISSUE #954: Sortable behavior not working with ENUM scope
Here is a problem that I encountered using Propel1, I just reproduced it on Propel2, so here it is:

When adding a sortable behavior scopped to a column that has a ENUM type, the rank is wrong on creation (and on all the futher rank manipulations done on the object).

I am kind of new to this project so I don't know if I will manage to add the tests and modify the generator to produce a PR... So here is the workaround I am currently using:

I override the `getScopeValue` method in the object class. In the base class, the method looks like this:

```php
public function getScopeValue($returnNulls = true)
{
    return $this->get{MyEnumColumn}();
}
```

So I simply override the method to make it return the actual SQL value corresponding to my enum value like so:

```php
public function getScopeValue($returnNulls = true)
{
    return array_search($this->get{MyEnumColumn}(), {MyTableMap}::getValueSet({MyTableMap}::COL_{MYENUMCOLUMN}));
}
```

pimpreneil:
Here is a fix for the enum scope issue on Sortable behavior
It might not be perfect but at least it illustrates the issue and provides a test for it.

Here are the things that, to me, are arguable in this PR:
- I had to copy the regexp for the enum constants from the table map generator in the sortable generator
- I created the objects in the fixtures (you might prefer having them directly in the test case)

### DIFF
--- a/src/Propel/Generator/Behavior/Sortable/SortableBehaviorObjectBuilderModifier.php
+++ b/src/Propel/Generator/Behavior/Sortable/SortableBehaviorObjectBuilderModifier.php
@@ -313,6 +313,11 @@ public function getScopeValue(\$returnNulls = true)
 
     return \$onlyNulls && \$returnNulls ? null : \$result;
 ";
+        } else if ($this->behavior->getColumnForParameter('scope_column')->isEnumType()){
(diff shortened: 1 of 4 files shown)


marcj:
Awesome, thanks :)

PR #973, opened 2015-07-30
Allow ObjectCollections->toArray() to serialise without foreign objects.

kitsunde:
The default for `ObjectCollections->toArray()` was also to **always** serialise foreign objects. But the default for `Model->toArray()` was to **not** serialise foreign objects. I've put the `ObjectCollections->toArray()` to not serialise foreign objects by default to keep the API consistent.


marcj:
Guess it would be better to make the default to `true` or at least adjust all failing tests :smile:

gechetspr:

PR #981, opened 2015-08-10
Use generated slug instead of stored value when querying existence

EVILoptimist:
I can't think of any reason why $getter() should be preferred over $slug in this function. 

$getter() causes problems as described in #982.

Any thoughts or feedback would be appreciated :)

### DIFF
--- a/src/Propel/Generator/Behavior/Sluggable/SluggableBehavior.php
+++ b/src/Propel/Generator/Behavior/Sluggable/SluggableBehavior.php
@@ -285,18 +285,19 @@ protected function makeSlugUnique(\$slug, \$separator = '" . $this->getParameter
     if (!\$alreadyExists) {
         \$slug2 = \$slug;
     } else {
-        \$slug2 = \$slug . \$separator;";
+        \$slug2 = \$slug . \$separator;
+        \$filterSlug = (\$this->$getter()) ? \$this->$getter() : \$slug;";
 
         if (null == $this->getParameter('slug_pattern')) {
             $script .= "
 
         \$count = " . $this->builder->getStubQueryBuilder()->getClassname() . "::create()
-            ->filterBySlug(\$this->$getter())
+            ->filterBySlug(\$filterSlug)
             ->filterByPrimaryKey(\$this->getPrimaryKey())
         ->count();
 
         if (1 == \$count) {
-            return \$this->$getter();
+            return \$filterSlug;
         }";
         }
 


src/Propel/Generator/Behavior/Sluggable/SluggableBehavior.php
```diff
@@ -291,7 +291,7 @@ protected function makeSlugUnique(\$slug, \$separator = '" . $this->getParameter
             $script .= "
 
         \$count = " . $this->builder->getStubQueryBuilder()->getClassname() . "::create()
-            ->filterBySlug(\$this->$getter())
+            ->filterBySlug(\$slug)
```
staabm: in case this change is correct line 299 also need to change?

staabm: I dont know, why we dont use the $slug parameter here, but there are even more conditions in this method which need to be adopted in case you are right.

EVILoptimist: Yeah. 299 would make perfect sense also.
Not sure about 317. And there's one more that I think should not be changed as well.

The only thing I can think is that this might interfere if you're trying to explicitly change the slug using setSlug()

I will investigate further.

EVILoptimist:
OK. I understand better what's happening here.
If the slug is already set and available via the getter() everything works and everyone is happy.
If the getter() returns NULL and the value stored in the database is NULL (i.e. when you add the sluggable behavior to an existing table). The current implementation assumes that it's already unique and moves on.

I've added a conditional to make sure that the database is not queried for a null value.

Questions and/or feedback would be appreciated.

marcj:
Can you add please a failing unit test? I'm not sure what this PR is trying to fix actually.

EVILoptimist:
Yeah. I started on that and ran into a couple snags. I'll see if I can dig back into it this week.

marcj:
thank you, looking forward to it :)

marcj:
ping @EVILoptimist :)

dereuromark:
@EVILoptimist Do you want to proceed with this? Or anyone else?

Incognito:
Tests are green when rebased.

https://travis-ci.org/github/spryker/Propel2/builds/706862703

dereuromark:
Is anyone opening a new (rebased) PR?

Incognito:
I suspect people aren't going to jump and rebase branches from 5 years ago, especially after I've already done it. We have a rebased branch on spryker/propel2 and we can decide what to do with it.

PR #990, opened 2015-08-28
Added option --composer-dir to the sql:build command

dylanvorster:
Hi There

I ran into a situation where I have a behaviour thats part of of my main project, and when I ran the sql:build command, it complained about not being able to find my behaviour because it was using a different composer directory, and therefore wanted me to provide the composer-dir option (I have a fairly complex project structure, with modules and plugins). Upon adding the composer-dir option, Propel told me that it wasn't an option for the command sql:build. So upon inspecting the command, lone behold it doesn't exist. After adding it and running the sql:build command again with --composer-dir, everything worked as expected.

Im aware you are in the process of making a massive change with the internals of propel, but I'm offering this small fix anyway (in case someone ran into a similar issue :) ).

Cheers


marcj:
Sounds good, thanks!

PR #996, opened 2015-09-11
Fix no drop table and constraint when migrate:diff (PostgreSQL)

ktounet:
Removed tables was not retrieved from the current database state.
Therefore the two variables ```$fromDatabaseTables``` and ```$toDatabaseTables``` had the same tables at the moment of the comparison in the DatabaseComparator...

### DIFF
--- a/src/Propel/Generator/Builder/Om/ObjectBuilder.php
+++ b/src/Propel/Generator/Builder/Om/ObjectBuilder.php
@@ -469,7 +469,8 @@ abstract class ".$this->getUnqualifiedClassName().$parentClass." implements Acti
 
         $script .= "
     /**
-     * The value for the $clo field.";
+     * The value for the $clo field.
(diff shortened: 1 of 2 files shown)


marcj:
Thanks, but isn't information_schema a postgresql v9+ feature only?

ktounet:
It's a major change of version 7.4 :
http://www.postgresql.org/docs/7.4/static/release-7-4.html (section named "SQL-standard information schema")

However, most of improvements of the information schema have been made since version 9.
I haven't tested it with an earlier version of PostgreSQL, should I?

marcj:
Well, than anything is fine. @apinnecke WDYT?

apinnecke:
I haven't enough knowledge about PostgreSQL to give my words, but the strategy and argumentation seem sane. So :+1: from me. :)

marcj:
Thank you @ktounet!

PR #997, opened 2015-09-12
Use absolute path to include propel.php

ikeyan:
`include('propel.php')` can include the propel configuration file, so an absolute path should be used.


glensc:
damn, i got hit by this bug, especially as `propel init` creates `propel.php`. i guess next release is far away?

PR #999, opened 2015-09-16
'Complex count': COUNT(*) => SQL_CALC_FOUND_ROWS

jadbaz:
This is not a 'better' method in general, it's just one that I have found to be faster for most cases.
It's also not an extremely clean method as I'm using a string manipulation and using 2 consecutive queries.

The problem with subselects (anything in the form of SELECT  FROM (sql)) is that it is not optimized in MYSQL. Do it on a large dataset and compare it to a single select. The more subselects, the slower the query.
On 3300 rows having 40 columns, mysql 5.1.73:
_Old method: between 1 and 8 seconds (No idea why so inconsistent).
_New method: between 0.6 and 1.1 seconds consistently

### DIFF
--- a/src/Propel/Runtime/ActiveQuery/Criteria.php
+++ b/src/Propel/Runtime/ActiveQuery/Criteria.php
@@ -2490,7 +2490,17 @@ class Criteria
                 $db->turnSelectColumnsToAliases($this);
             }
             $selectSql = $this->createSelectSql($params);
-            $sql = 'SELECT COUNT(*) FROM (' . $selectSql . ') propelmatch4cnt';
+            //replace the first occurence of 'select' with 'SELECT SQL_CALC_FOUND_ROWS '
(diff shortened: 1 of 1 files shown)


jadbaz:
The failing check has to do with the expectation of SELECT COUNT(*). SELECT FOUND_ROWS() found instead.
This is the expected behaviour of this check.
Still needs to be considered

angelk:
IMO SQL_CALC_FOUND_ROWS  is [not always faster](http://stackoverflow.com/questions/186588/which-is-fastest-select-sql-calc-found-rows-from-table-or-select-count/188682#188682)

asmarovydlo:
Hello, `SQL_CALC_FOUND_ROWS`  is deprecated in MYSQL since `8.0.17` https://dev.mysql.com/doc/refman/8.0/en/information-functions.html#function_found-rows 

I think we can close the PR

gechetspr:
Is outdated

PR #1000, opened 2015-09-17
Migrate from circle to travis

Big-Shark:
I think circleci is not the best choice, and we need more different environments.
I did this PR for moving to travis.
Travis provides more different environments and it allows us to testings the propel for better quality

https://travis-ci.org/Big-Shark/Propel2

P.S. I could not understand how you update the codeclimate. Please add it yourself.


marcj:
Well, we use circle because it is almost twice as fast as travis

Big-Shark:
The circle check only one environment, but Travis environment list a lot more, we can keep both, but you need a fix the settings for the circle

marcj:
Yeah I see that argument. Let us use then both and keep circleci (because I like it more and it's faster). Could you please readd circleci stuff? :smile:

Big-Shark:
@marcj Done :)

Big-Shark:
@marcj we have any problems with that?

phpunit.xml.dist
```diff
@@ -26,26 +38,6 @@
...
-        <file>src/Propel/Generator/Reverse/OracleSchemaParser.php</file>
-        <file>src/Propel/Generator/Reverse/SqlsrvSchemaParser.php</file>
-
-        <directory>src/Propel/Runtime/Adapter/MSSQL</directory>
-        <file>src/Propel/Runtime/Adapter/Pdo/MssqlAdapter.php</file>
-        <file>src/Propel/Runtime/Adapter/Pdo/OracleAdapter.php</file>
-        <file>src/Propel/Runtime/Adapter/Pdo/SqlsrvAdapter.php</file>
-
-        <file>src/Propel/Runtime/Adapter/Pdo/SqlsrvAdapter.php</file>
-      </exclude>
```
marcj: Why removing that?

marcj:
Looks cool, but I'd like to prefer that we include experimental platforms like mssql, oracle and sqlsrv in the `filter>whitelist>exclude` of each phpunit.xml, so it doesn't destroy our code coverage.

Big-Shark:
@marcj I added

marcj:
cool, thanks :)

Big-Shark:
#1059

PR #1007, opened 2015-10-09
- Adjusted query for get the indices and the way to explore the answe…

jag2kn:
- Adjusted query for get the indices and the way to explore the answer for the reverse engineer in Oracle databases.

 - Adjustment "parse" function of OracleSchemaParser class for get the correct pattern of auto-incrementing fields correctly.


marcj:
I'm afraid we cannot test if this PR is valid since we have currently no active team member with running oracle database :(

jag2kn:
:(

marcj:
But I'm going to merge it anyway :P Thanks @jag2kn!

Issue #892, opened 2015-03-05
Joins with multiple foreign keys to same table fail


jengeb:
I have problems with two foreign keys referencing the same table `users`. In a table `movies` I store both the user who created an entry (`CreateUser`) and the user who updated this entry (`UpdateUser`), see the [sandbox](http://sandbox.propelorm.org/e5a686f).

I use `join()` to join the tables `movies` and `users`. Either the query to get `CreateUser` or `UpdateUser` is working. However, joining both fails. The executed SQL query seems to be

```
SELECT movies.ID, movies.TITLE,
  users.NAME AS CreateUserName,
  users.NAME AS UpdateUserName
  FROM `movies`
  INNER JOIN `users` ON (movies.CREATE_USER_ID=users.ID)
  INNER JOIN `users` ON (movies.UPDATE_USER_ID=users.ID)
```

Note that `users` appears twice. Probably it should read
```
SELECT movies.ID, movies.TITLE,
  CreateUser.NAME AS CreateUserName,
  UpdateUser.NAME AS UpdateUserName
  FROM `movies`
  INNER JOIN `users` AS CreateUser ON (movies.CREATE_USER_ID=CreateUser.ID)
  INNER JOIN `users` AS UpdateUser ON (movies.UPDATE_USER_ID=UpdateUser.ID)
```

Is there something wrong in my sandbox code or is there a bug in Propel2?

andrenarchy:
I also noticed that the resulting array contains a key `User` (which is not requested by the code).

jengeb:
Can someone recommend a workaround for this issue?

nicklaros:
join with alias solve this, for example:
```
$movie = MovieQuery::create()
  ->leftJoinCreator('Creator')
  ->withColumn('Creator.Name', 'CreateUserName')
  ->leftJoinUpdater('Updater')
  ->withColumn('Updater.Name', 'UpdateUserName');
```
notice that in your schema file.. `Creator` and `Updater` above is the phpName of foreign key in movie table
```
    <table name="movie">
        <column name="id"                       type="BIGINT" size="20" primaryKey="true" autoIncrement="true" />
        <column name="name"                    type="LONGVARCHAR" />
        <column name="creator_id"            type="BIGINT" size="20" />
        <column name="updater_id"           type="BIGINT" size="20" />

        <foreign-key foreignTable="user" phpName="Creator" refPhpName="CreatedMovie">
            <reference local="creator_id" foreign="id" />
[...]

PR #1009, opened 2015-10-15
Fixes issue #892 with multiple FKs to the same table

pawel-oronowicz:
Fixes issue with multiple FKs referencing the same table despite different PhpNames specified.


marcj:
Cool, thanks! But can you please add a unit test that proves your fix?

PR #1015, opened 2015-10-17
Implemented use statement for baseClass

geolysis:
Base classes in a different namespace are currently not supported. It would be great, if you would consider this change, that this becomes a part of the standard functionality.

Thanks, Andreas

### DIFF
--- a/src/Propel/Generator/Builder/Om/ObjectBuilder.php
+++ b/src/Propel/Generator/Builder/Om/ObjectBuilder.php
@@ -264,7 +264,7 @@ abstract class ".$this->getUnqualifiedClassName().$parentClass." implements Acti
         $this->declareClassFromBuilder($this->getStubQueryBuilder());
         $this->declareClassFromBuilder($this->getTableMapBuilder());
 
-        $classesToDeclare = array(
+        $this->declareClasses(
             '\Exception',
             '\PDO',
             '\Propel\Runtime\Exception\PropelException',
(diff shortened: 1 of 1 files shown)


src/Propel/Generator/Builder/Om/ObjectBuilder.php
```diff
@@ -282,6 +282,14 @@ protected function addClassBody(&$script)
             '\Propel\Runtime\Map\TableMap'
         );
 
+        $baseClass = $this->getBaseClass();
+        if (strrpos($baseClass, '\\') !== FALSE) {
+            $classesToDeclare[] = $baseClass;
+        }
+
+        $reflectionMethod = new \ReflectionMethod($this, 'declareClasses');
```
lbarulski: why not: 
```php
call_user_func_array([$this, 'declareClasses'], $classesToDeclare);
```
?

geolysis: This comes down to the preferred programming style. My preference is an object orientated approach. Of course `call_user_func_array` is valid solution, as well.

staabm: I would guess that the RefecltionMethod is much slower then cufa

marcj: Well, I'd completely not use any of those stuff.

```php
$baseClass = $this->getBaseClass();
if (false !=== strrpos($baseClass, '\\')) {
    $this->declareClasses($baseClass);
}
```

Doing it this way is easier to read, less modification and no user_func-call

geolysis:
Perfect solution, Marc.

geolysis:
Happy to implement it now with the adjustment made?

marcj:
Well, thanks @geolysis :)

PR #1016, opened 2015-10-21
vendor element missing from DTD

kitsunde:
See "Global Vendor Info" http://propelorm.org/documentation/reference/schema.html


marcj:
Thanks :)

PR #1020, opened 2015-10-22
Proposal for implement a ParentClass for query objects at database or table level

jeremydenoun:
I need to extends TableQuery objects with common methods, so I propose to add parameter similar to baseClass named parentClass for custom builder generator  extends "class name" ...

The class should extends ModelCriteria for be compatible (php doesn't support multiple extends)


marcj:
Well, I'd actually prefer another name like `parentQueryClass` or `baseQueryClass` or something like that indicating that its related to the `*Query` classes;

marcj:
Very nice, thanks @jeremydenoun!

PR #1021, opened 2015-10-26
fixed question mark recognition

lbarulski:
Fails because of invalid usage of `strpos` function.
Throws an exception when `$this->clause` will be started with question mark, f.ex. `'? in (col1, col2, col3)'`


marcj:
good catch, thank you!

Issue #1019, opened 2015-10-22
Inline assigment should be avoided


dereuromark:
There is even an error in there because of that:

In Generator DelegateBehavior:
```
if (is_callable(array('$ARFQCN', \$name))) {
    if (!\$delegate = \$this->get$relationName()) {
        \$delegate = new $ARClassName();
        \$this->set$relationName(\$delegate);
    }

    return call_user_func_array(array(\$delegate, \$name), \$params);
}";
```

The `if (!\$delegate = \$this->get$relationName()) {` makes assignment and negation in an unexpected order and leads to undesired behavior.

This should read (and then work as expected):
```
\$delegate = \$this->get$relationName();
if (!\$delegate) {
```

marcj:
Thanks for reporting this issue. Would you like to create a [Pull-Request](https://help.github.com/articles/using-pull-requests/)?

dereuromark:
Yes I will do that.
Those bugs are quite severe and that !$x = $y behavior should be fixed asap.

dereuromark:
Can you maybe add Travis for PRs, too?

PR #1024, opened 2015-10-28
Bugfix/bad inline assigment

dereuromark:
Resolves https://github.com/propelorm/Propel2/issues/1019

The `if (!\$delegate = \$this->get$relationName()) {` makes assignment and negation in an unexpected order and leads to undesired behavior.

This should read (and then work as expected):
```
\$delegate = \$this->get$relationName();
if (!\$delegate) {
```

Also fixed the other inline assigments.


marcj:
Very nice, thank you!

dereuromark:
No problem!
Whats your estimate on a new alpha6 release etc?
Seems like a lot of changes piled up over the last months.

Issue #968, opened 2015-07-13
Issue in many-to-many form


pimpreneil:
Hi,

I am experiencing a very problematic issue on a form submission involving a one-to-many relationship:

Here is what I want to achieve:
  I have a company with users attached to it. I want to be able to add / update / delete users in the company form.

Schema:

```xml
<database name="default" namespace="Acme\Bundle\BlogBundle\Model">
    <table name="company">
        <column name="id" type="integer" required="true" primaryKey="true" autoIncrement="true" />
        <column name="name" type="varchar" required="false" primaryString="true" />
    </table>                                           
    <table name="user">
        <column name="id" type="integer" required="true" primaryKey="true" autoIncrement="true" />
        <column name="company_id" type="integer" required="true" />
        <column name="firstname" type="varchar" required="false" primaryString="true" />

        <foreign-key foreignTable="company" onDelete="CASCADE" onUpdate="CASCADE">
            <reference local="company_id" foreign="id"/>
        </foreign-key>
    </table>
</database>
```

Controller:

```php
...
public function companySaveAction(Request $request)
{
    $company = CompanyQuery::create()->findOne();
    $formObj = $this->createForm('company_type', $company);
    $formObj->handleRequest($request);

    if ($formObj->isValid()) {
        $company->save();

        return $this->redirect($this->generateUrl('acme_blog_company'));
    }

    return array('form' => $formObj->createView());
}
...
```

Form type:

```php
...
public function buildForm(FormBuilderInterface $builder, array $options)
{
    $builder->add('name');
    $builder->add('users', 'collection', array(
        'type' => 'user_type',
        'allow_add' => true,
        'allow_delete' => true,
        'by_reference' => false
    ));
    $builder->add('save', 'submit');
}
...
```

My issue is that on first form submission, everything works fine: The users are created properly and the company name is updated as it should.
But if I submit my form again (without deleting the users, just changing their name), they get deleted.
After that, I can add new users, they get created and if I submit the form again, they disappear, etc...

Here is a sample of the request parameters that sent to the controller (we are here only working in a company edition scenario to avoid confusion):

First request:

```yaml
company_type[name]:Main company
company_type[users][0][firstname]:user 1
company_type[users][1][firstname]:user 2
company_type[save]:
company_type[_token]:...
```
=> All the users are created just fine

Second request:

```yaml
company_type[name]:Main company
company_type[users][0][firstname]:user 1 edited
company_type[users][1][firstname]:user 2 edited
[...]

PR #1027, opened 2015-10-30
#968 - Fix the child deletion issue in many-to-one form

pimpreneil:
Here is a fix for the Child deletion issue in many-to-one form issue.

Because this issue is caused mainly by Symfony not comparing the collection objects like Propel is made for, I unfortunately did not manage to provide any tests, though this fix does not break any existing ones...

### DIFF
--- a/src/Propel/Generator/Builder/Om/ObjectBuilder.php
+++ b/src/Propel/Generator/Builder/Om/ObjectBuilder.php
@@ -3745,6 +3745,8 @@ abstract class ".$this->getUnqualifiedClassName().$parentClass." implements Acti
 
         $collName = $this->getRefFKCollVarName($refFK);
 
+        $scheduledForDeletion = lcfirst($this->getRefFKPhpNameAffix($refFK, $plural = true)) . "ScheduledForDeletion";
+
         $script .= "
     /**
      * Method called to associate a $className object to this object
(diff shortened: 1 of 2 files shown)


marcj:
Thanks! Can you also please add a failing unit tests that covers your changes? I'm not sure if this fixes really that issue. I think we had a issue with this approach (performance/memory) and changed it a bit. Not sure why it isn't working at the moment.

pimpreneil:
Thanks @marcj , unfortunately, as mentioned I really am not capable to write the failing test (I did try!). 
This PR does fix the issue, one of my projects was stuck because of this issue, and this fix really sorted the things out.
There must indeed be a reason for this code to be removed in Propel2.

As detailed in the related issue #968 , the problem originates from an object comparison that fails when using Propel, and succeeds when using Doctrine

pimpreneil:
@marcj Sorry, I fucked up a bit and pushed my modifs twice (forgot to merge master before pushing my branch)...
Anyway, here is the requested failing test.

Maxooo:
Good test, thank you @pimpreneil :+1:

pimpreneil:
If you don't want to merge it, maybe change the tag? :)

marcj:
Very nice, could you please squash your commits into one? :)

pimpreneil:
@marcj Done. Thx :grinning:

marcj:
Awesome, thanks =)

PR #1029, opened 2015-11-04
pager always 0 when no results fixes #1025

### LINKED ISSUE #1025: Pagination wrong if no results
Hi, there is a problem with the last page if there is no result.

```php
$pager = $query->paginate(1, 100);

var_dump($pager->getNbResults()); // 0
var_dump($pager->getPage()); // 1
var_dump($pager->getFirstPage()); // 1
var_dump($pager->getLastPage()); // 0
```

It is not logical that the current page and the first page are greater than the last page !
The value of last page should be equal to 1.

nymo:


marcj:
Thanks!

PR #1038, opened 2015-11-19
allow select column from joined table

### LINKED ISSUE #1037: Select column from joined table throws exception
```php
        BookQuery::create()
            ->innerJoinAuthor()
            ->select(AuthorTableMap::COL_LAST_NAME)
            ->find();
```
throws following exception
```
Propel\Runtime\Exception\PropelException: Unable to execute SELECT statement [SELECT  AS "author.last_name" FROM book INNER JOIN author ON (book.author_id=author.id)]
```

Problem is placed in `ModelCriteria::getColumnFromName`, `$this->joins` contains keys based on string generated in join method f.ex. `BookQuery::joinAuthor`:
```php
$this->addJoinObject($join, 'Author');
```
but `AuthorTableMap::COL_LAST_NAME` uses table name as prefix, so `ModelCriteria::getColumnFromName` tries to find key `author` in array which contains key `Author`.

lbarulski:
Does anyone have a better idea how to solve it? I didn't find a way to get `TableMap` by table name.

Fixes #1037


marcj:
Well looks like a solid fix, doesn't it? I'm going to merge this for now, tests are obviously green :)

PR #1041, opened 2015-11-20
added type hint for phpdoc in get and set method  fixes 985

### LINKED ISSUE #985: Wrong datatype in PHP Doc when using typehint attribute
I have the following schema:
```
<table name="user">
    <column name="email" type="varchar" size="255" typehint="my\namespace\Email"></column>
</table>
```

The generated base class then looks the following:
```php
use my\namespace\Email;
...

abstract class User implements ActiveRecordInterface {
     ...

     /**
     * Set the value of [email] column.
     *
     * @param string $v new value
     * @return $this|\User The current object (for fluent API support)
     */
    public function setEmail(Email $v = null)
    {
        if ($v !== null) {
            $v = (string) $v;
        }

        if ($this->email !== $v) {
            $this->email = $v;
            $this->modifiedColumns[UserTableMap::COL_EMAIL] = true;
        }

        return $this;
    } // setEmail()

     /**
     * Get the [email] column value.
     *
     * @return my\namespace\email
     */
    public function getEmail()
    {
        return $this->email;
    }
```
In the setter the typehint is right but the doctag is not (as it describes $v to be string instead of Email).

In the setter the typehinted class is referenced with its fully qualified name instead of only Email (because it got already referenced with the use-statement).

I have to admit that I did not run this code. It is only what PhpStorm is marking.

But one more question: I understood that serialization of the Email object is done with the __toString() method (because of the string cast in the setter). But is Propel able to create the Email object when querying a User entity?

Also I am wondering how the query class' filterByEmail method could use Email objects as search parameter.

nymo:
Added type hint in set method and replaced the type hint in the get
method with the type hint short notation. Fixes #985

### DIFF
--- a/src/Propel/Generator/Builder/Om/ObjectBuilder.php
+++ b/src/Propel/Generator/Builder/Om/ObjectBuilder.php
@@ -1181,6 +1181,15 @@ abstract class ".$this->getUnqualifiedClassName().$parentClass." implements Acti
     {
         $clo=$column->getLowercasedName();
 
+        $typeHint = null;
+
+        if($column->getTypeHint()){
+            $typeHint = $column->getTypeHint();
+            if ('array' !== $typeHint) {
+                $typeHint = $this->declareClass($typeHint);
+            }
+        }
+
         $script .= "
     /**
      * Get the [$clo] column value.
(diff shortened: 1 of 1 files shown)


nymo:
Mhm local tests are green. Looks like a general problem with the CI Tool?

marcj:
yeah, after the PR from travis everything broke although the PR showed that everything is green.

nymo:
@marcj since the CI tool is fixed again could you restart the ci test or do I have to add another commit?

PR #1043, opened 2015-11-23
Add timestamptz for postgres

### LINKED ISSUE #589: [Postgres] Allow 'timestamp with time zone' type to save time zone information
I have a number of tables with timestamp with time zone data type fields.

The code generated by propel for these fields in my Base classes is of the form:

function doInsert()
{
.....
   case 'ENTRY_DATE':
                        $stmt->bindValue($identifier, $this->entry_date ? $this->entry_date->format("Y-m-d H:i:s") : null, PDO::PARAM_STR);
                        break;
....
}

This is fine for fields of type 'timestamp', but not 'timestamp with time zone'.

I'm saving data as follows:
$obj=new Foo();
$obj->setEntryDate(new DateTime(null, new DateTimeZone('GMT'));
$obj->save();

I'm in the EST time zone (GMT -5).

The date that gets saved in the database is (assuming the date in the input was: 2014-03-26T20:15:42+00:00):

2014-03-26T20:15:42-05:00

Which is the GMT timestamp, but with the local time zone.  Postgres converts everything to GMT when it stores it and converts it back to local time when queried, but since the timezone info was missing when it went in, the time coming out is now incorrect.

There is a very simple fix which I think should be generally compatible - simply change the date format in the generated code from this:

$this->entry_date->format("Y-m-d H:i:s")

to this:

$this->entry_date->format("c")

Postgres understands ISO860 as a valid date format for both timestamp and timestamp with time zone data types.

bpicolo:
This basically fixes #589 by adding support for timestamptz. For pgsql, this means use timestamptz. For other databases, it should behave like timestamp (I may have gone overkill in ensuring that bit too, heh).

I don't consider this ready to ship yet. Looking for some feedback because I've never contributed to the code base (and PHP is not my forté). Also would like suggestions as to exactly which parts should get specific tests (An answer of `all of it` is totally groovy.)

### DIFF
--- a/src/Propel/Generator/Builder/Om/ObjectBuilder.php
+++ b/src/Propel/Generator/Builder/Om/ObjectBuilder.php
@@ -143,6 +143,8 @@ class ObjectBuilder extends AbstractObjectBuilder
             $fmt = $this->getPlatform()->getTimeFormatter();
         } elseif ($column->getType() === PropelTypes::TIMESTAMP) {
             $fmt = $this->getPlatform()->getTimestampFormatter();
+        } elseif ($column->getType() === PropelTypes::TIMESTAMPTZ) {
+            $fmt = $this->getPlatform()->getTimestampTzFormatter();
         }
 
         return $fmt;
@@ -778,7 +780,7 @@ abstract class ".$this->getUnqualifiedClassName().$parentClass." implements Acti
 
         $handleMysqlDate = false;
         if ($this->getPlatform() instanceof MysqlPlatform) {
-            if ($column->getType() === PropelTypes::TIMESTAMP) {
+            if ($column->getType() === PropelTypes::TIMESTAMP || $column->getType() === PropelTypes::TIMESTAMPTZ) {
                 $handleMysqlDate = true;
                 $mysqlInvalidDateString = '0000-00-00 00:00:00';
             } elseif ($column->getType() === PropelTypes::DATE) {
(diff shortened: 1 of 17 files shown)


src/Propel/Generator/Builder/Om/ObjectBuilder.php
```diff
@@ -778,7 +780,7 @@ public function addTemporalAccessorComment(&$script, Column $column)
 
         $handleMysqlDate = false;
         if ($this->getPlatform() instanceof MysqlPlatform) {
-            if ($column->getType() === PropelTypes::TIMESTAMP) {
+            if ($column->getType() === PropelTypes::TIMESTAMP || $column->getType() === PropelTypes::TIMESTAMPTZ) {
```
bpicolo: this might be unnecessary because of the platform initialize() bits? (Havent checked yet)

src/Propel/Generator/Builder/Om/ObjectBuilder.php
```diff
@@ -1616,16 +1622,7 @@ protected function addTemporalMutator(&$script, Column $col)
...
-                case 'DATE':
-                    $format = 'Y-m-d';
-                    break;
-                case 'TIME':
-                    $format = 'H:i:s';
-                    break;
-                default:
-                    $format = 'Y-m-d H:i:s';
-            }
+            $format =  $this->getTemporalFormatter($col);
```
bpicolo: Not sure if this is the right change, but it sure FEELS better

src/Propel/Generator/Platform/DefaultPlatform.php
```diff
@@ -1279,6 +1282,15 @@ public function getTimestampFormatter()
     }
 
     /**
+     * Gets the preferred timestamp with timezone formatter for setting date/time values.
+     * @return string
+     */
+    public function getTimestampTzFormatter()
```
bpicolo: would we prefer to explode on other platforms trying to use timestamptz?

bpicolo:
@marcj  mind taking a look at this?

marcj:
I'll take a look in some days. I'm unfortunately at the moment very busy and I want to test this first with my local postgres installations. :)

bpicolo:
Sweet. Thanks =]

bpicolo:
@marcj had some time to add a bit of relevant coverage. Let me know if there are some other important spots I missed.

eastebry:
Have there been any updates to this?

Issue #1023, opened 2015-10-28
Export only one column from collection? 


Alekc:
Am I blind or there is no quick and easy way of exporting only one column from collection as array?

Something similar to

 ```php
public function toColumnValue($keyColumn = 'PrimaryKey')
    {
        $ret = array();
        $keyGetterMethod = 'get' . ucfirst($keyColumn);
        foreach ($this as $obj) {
            $ret[] = $obj->$keyGetterMethod();
        }

        return $ret;
    }
```

Thanks.

**Comment by cristianoc72** [comment]
You can do it in your query
http://propelorm.org/documentation/reference/model-criteria.html
See paragraph "Getting columns instead of objects".
Il 28/ott/2015 17:40, "Alexander Chernov" <notifications@github.com> ha
scritto:

**Comment by Alekc** [comment]
Yes, I am aware of that way. I am thinking about a bit different type of usage. 

For example, in some cases I need to fetch list of authors, and later in the same code I need to extract their ids in order to use in another query or for different purpose. 

Right now, the only more or less straightforward way of dealing with it (without calling yet another query) would be to call 
```php
$authors = AuthorsQuery::create()->find();
$ids = array_keys($authors->toKeyIndex);
```
which is a bit of overkill for my tastes/.

Other times you need to extract foreign ids from 1:1 child table. So, to sum it up: a function which would permit you to extract column values from collection and still have full original objects which can be used later.

marcj:
Like this feature. :+1:

Alekc:
I took a liberty to add a pull request, feel free to rename the method :)

Big-Shark:
:+1:

PR #1045, opened 2015-11-24
Implemented method getColumnValues for export of column values.

Alekc:
From the issue https://github.com/propelorm/Propel2/issues/1023


marcj:
Very cool, thanks @Alekc !

PR #1047, opened 2015-12-02
Include identifierQuoting in the reverse-engineered schema.xml

motin:
I am working on a new SchemaParser using which the need for identifier quoting is detectable at the schema level. Without this PR however, the identifierQuoting attribute is not included in the reverse-engineered schema.xml

Let me know if there are any underlying reasons as to NOT include identifierQuoting.

Thanks


marcj:
great thanks!

Issue #1048, opened 2015-12-02
Feature: transaction support for migrations


eelkevdbos:
Wrapping a migration in a transaction improves the migration robustness. Would it be possible to add transaction statements at the beginning and end of the getUpSQL() and getDownSQL() methods for database drivers that support transactions natively?

I've been wrapping my migrations in "START TRANSACTION;" and "COMMIT;" statements (as a PostgreSQL user) for a while now and would recommend it to anyone having to do complex migrations that leave a mess when executed partially due to exceptions.

marcj:
Yeah like it, but unfortunately not supported by MySQL and SQLite. We should nonetheless go that way and include it per default for those platforms which support it.

eelkevdbos:
Great! Do you need a volunteer to implement it? If so, I would like to volunteer, but could use some 'best-practices' as to where to implement it. Should it be part of the MigrationMigrateCommand or would it be better to generate the raw SQL statements?

staabm:
IIRC we have a `transactional` helper which wrapps statments into a transaction

marcj:
Yeah would be pretty cool @eelkevdbos! :)
I'd place it directly in the platform class of postgres like MySQL it does with its foreign key DDL: https://github.com/propelorm/Propel2/blob/master/src/Propel/Generator/Platform/MysqlPlatform.php#L157-L172. ´getBeginDDL` is not defined in https://github.com/propelorm/Propel2/blob/master/src/Propel/Generator/Platform/PgsqlPlatform.php so you need to overwrite it there and return some transaction queries.

eelkevdbos:
Thanks for the leads @marcj + @staabm! I had to modify some of the unit tests, but the PostgreSQL implementation is finished. Should we keep this issue open for other platform implementations or just merge and close?

eelkevdbos:
Issue solved for PostgreSQL

PR #1049, opened 2015-12-02
Implemented migration transactions for PostgreSQL

eelkevdbos:
Single platform solution for #1048


marcj:
Looks good, thanks @eelkevdbos!

Issue #832, opened 2014-12-07
It would be awesome to have a way of getting empty migrations on the command line for data migrations.


kitsunde:
A common use case is that our client requests to have a list of things say `Company` with company `A`, `B`, `C` which gets added on table creation in the same migration, but then a day later they'd also like to have company `D` in there.

In Django we would create an empty migration  (https://docs.djangoproject.com/en/1.7/topics/migrations/#data-migrations) and add the additional change there.

It would be awesome if propel would also accept an `--empty` parameter to `propel diff` for data migrations.

PR #1055, opened 2015-12-05
Added migration create command for creating empty migration classes

motin:
Fixes https://github.com/propelorm/Propel2/issues/832

For semantic reasons, I chose to implement it as a separate command instead of an "--empty" command to migration:diff.

Usage and verbose output:
```
# vendor/bin/propel migration:create -vvv
"/app/dna/generated-migrations/PropelMigration_1449320135.php" file successfully created.
Now add SQL statements and data migration code as necessary.
Once the migration class is valid, call the "migrate" task to execute it.
```

### DIFF
--- a/src/Propel/Generator/Command/MigrationCreateCommand.php
+++ b/src/Propel/Generator/Command/MigrationCreateCommand.php
@@ -1,0 +1,100 @@
+<?php
+
+/**
+ * This file is part of the Propel package.
+ * For the full copyright and license information, please view the LICENSE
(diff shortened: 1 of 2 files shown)


motin:
Note: The travis builds are failing because of other reasons than this PR. Actually, accepting https://github.com/propelorm/Propel2/pull/1056 results in green travis builds again.

marcj:
Very cool, thanks! We should also add this to PropelBundle: https://github.com/propelorm/PropelBundle/issues/372

PR #1057, opened 2015-12-05
Additional iconv transliteration tests

motin:
This helps isolate odd iconv transliteration behavior, to make it clear when Propel2 is at fault and when failed tests stems from incorrect/unsupported locale configuration.

### DIFF
--- a/tests/Propel/Tests/Generator/Util/IconvTransliterationTest.php
+++ b/tests/Propel/Tests/Generator/Util/IconvTransliterationTest.php
@@ -1,0 +1,59 @@
+<?php
+
+/**
+ * This file is part of the Propel package.
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ *
+ * @license MIT License
+ */
+
+use Propel\Generator\Util\PhpParser;
+use \Propel\Tests\TestCase;
+
+/**
+ *
+ */
+class IconvTransliterationTest extends TestCase
+{
+
+    /**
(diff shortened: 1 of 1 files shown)


marcj:
Well, guess this is not complete since there is a PHP error in the tests

motin:
@marcj This is now complete with passing tests

dereuromark:
This looks nice!
Can we maybe rebase against master (or merge it) so we can check if it is still green and merge it?

Incognito:
PHP links to libiconv installed on the host, it doesn't ship specific versions (similar to how PHP uses cURL, etc...).

Beware: A lot of this behaviour is dependent on libiconv which has serious changes between 2015 and 2020. I ran into problems in 2016/2017 on PHP 7.x (but the version of php is irrelevant) because the base debian image on the host was built with one old version of libiconv and we were migrating to a new distro release which had a more modern libiconv release.

https://www.gnu.org/software/libiconv/

libiconv changelog: https://git.savannah.gnu.org/gitweb/?p=libiconv.git;a=blob_plain;f=ChangeLog;hb=refs/heads/master

Testing this could be

PhilinTv:
Add later to doc:
⚠️ Iconv transliteration tests may fail if the system locale is C or if libiconv version is inconsistent. To ensure stability, use en_US.UTF-8 locale when running tests.

PhilinTv:
Had to refactor a bit, but it finally made it, after years of waiting :) 
https://github.com/propelorm/Propel2/pull/2053

PR #1056, opened 2015-12-05
Fixes fatal error in test suites.

motin:
Original commit message: Fixed php7 test suites

Fixes fatal error described in https://github.com/propelorm/Propel2/issues/1052

The following warnings remain:

```
PHP Warning:  The use statement with non-compound name 'VersionableBehaviorTest8' has no effect in /tmp/propelQuickBuild-2.0.0-dev-7a1e019deb/VersionableBehaviorTest8_VersionableBehaviorTest8Version.php on line 3136
PHP Warning:  The use statement with non-compound name 'VersionableBehaviorTest8Foo' has no effect in /tmp/propelQuickBuild-2.0.0-dev-7a1e019deb/VersionableBehaviorTest8_VersionableBehaviorTest8Version.php on line 3281
```

Issue #1052, opened 2015-12-03
Fatal error when running tests on latest master - PHP Fatal error:  Cannot redeclare class Base\VersionableBehaviorTest8Query in /tmp/propelQuickBuild-2.0.0-dev-7a1e019deb/VersionableBehaviorTest8_VersionableBehaviorTest8Version.php on line 2996


motin:
Running tests against 5204a45 using php-fpm 5.6.14 on debian jessie.

I'd like to start contributing but would like a passing test suite locally first, so any hints as to what is required to run the tests successfully locally are much appreciated. Thanks :)

Steps to reproduce = follow the instructions on https://github.com/neam/docker-stack-for-propel2-unit-tests

```
root@6a89b8589387:/propel2# tests/bin/phpunit.agnostic.sh
Tests started in temp /tmp.
PHPUnit 5.0.10 by Sebastian Bergmann and contributors.

.....SS......................................................   61 / 3297 (  1%)
........................F.....F.....F.....F.....F............  122 / 3297 (  3%)
.............................................................  183 / 3297 (  5%)
.............................................................  244 / 3297 (  7%)
.............................................................  305 / 3297 (  9%)
.............................................................  366 / 3297 ( 11%)
...........................................................S.  427 / 3297 ( 12%)
.............................................................  488 / 3297 ( 14%)
.............................................................  549 / 3297 ( 16%)
...........................................F.................  610 / 3297 ( 18%)
S............................................................  671 / 3297 ( 20%)
.............................................................  732 / 3297 ( 22%)
......EPHP Fatal error:  Cannot redeclare class Base\VersionableBehaviorTest8Query in /tmp/propelQuickBuild-2.0.0-dev-7a1e019deb/VersionableBehaviorTest8_VersionableBehaviorTest8Version.php on line 2996
```

motin:
For information, the 5 failed tests are the following:

```
There were 5 failures:

1) Propel\Tests\Common\Config\Loader\IniFileLoaderTest::testIniFileNotReadableThrowsException
Failed asserting that exception of type "Propel\Common\Config\Exception\InputOutputException" is thrown.

2) Propel\Tests\Common\Config\Loader\JsonFileLoaderTest::testJsonFileNotReadableThrowsException
Failed asserting that exception of type "Propel\Common\Config\Exception\InputOutputException" is thrown.

3) Propel\Tests\Common\Config\Loader\PhpFileLoaderTest::testConfigFileNotReadableThrowsException
[...]

PR #1058, opened 2015-12-05
Fix warnings in issue 1052

motin:
Fixes warnings introduced by [the fix](https://github.com/propelorm/Propel2/pull/1056) to issue https://github.com/propelorm/Propel2/issues/1052


marcj:
Thanks @motin!

PR #1063, opened 2015-12-07
fixing serialization issues

### LINKED ISSUE #829: DateTime as Primary Key
Hi,

There is an issue when you define a primary key with a Date (datetime) type in the schema and then build the model, in the auto-generated code (in the Entity-Map.php), in the addInstanceToPool function, the DateTime attribute is casted to a String by doing: (string) dateTimePK. This is causing a Fatal Error exception, because the right way to cast a DateTime is using the ->format('Y-m-d') function. I did the change my self and everything worked fine. So I think the auto-generated code  should consider this right way to cast a DateTime into a String.

PD: By the way, excellent work you've done here, congrats and thank you very much! I love Propel.

Regards,

prgTW:
whenever a key that could be serialized couldn't be casted to a `string` at the same time (f.in. `\DateTime`) an exception was thrown

fixes #829

### DIFF
--- a/src/Propel/Generator/Builder/Om/TableMapBuilder.php
+++ b/src/Propel/Generator/Builder/Om/TableMapBuilder.php
@@ -646,14 +646,14 @@ class ".$this->getUnqualifiedClassName()." extends TableMap
         $pkphp = (array) $pkphp; // make it an array if it is not.
         $script = '';
         if (count($pkphp) > 1) {
-            $script .= "serialize(array(";
+            $script .= "serialize([";
             $i = 0;
             foreach ($pkphp as $pkvar) {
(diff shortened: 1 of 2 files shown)


staabm:
Please add a unit test

prgTW:
but for what? a unit test that would actually make sense would be for testing whole instance pooling imho because only change that this code changes is removing unnecessary casting everything to string

prgTW:
I've added test suite to check if adding to instance pool doesn't throw an exception if primary key is composed of `DATE` field

staabm:
@prgTW very nice, looks good to me :+1:

marcj:
Cool, thanks @prgTW!

PR #1064, opened 2015-12-08
Unit tests on Windows

oujesky:
I made few adjustments for running unit tests on Windows machine. 

Tests are passing green without DB, with MySQL and with SQLite on Windows 7, PHP 5.6. 

---

PostrgreSQL however for some weird reason fails few tests - mostly because different quoting - looks like the Windows version ignores binding values as integers and treats them as strings instead. 

That leads to following type of errors:
```
Expected :'INSERT INTO "group" ("id", "title", "as") VALUES (1, 'Test Group', 3)'
Actual   :'INSERT INTO "group" ("id", "title", "as") VALUES ('1', 'Test Group', 3)'
```

Any PostgreSQL expert who can give any insights? Is this some kind of known limitation or just misconfiguration? On linux, this works out of the box.

### DIFF
--- a/tests/Propel/Tests/Common/Config/ConfigurationManagerTest.php
+++ b/tests/Propel/Tests/Common/Config/ConfigurationManagerTest.php
@@ -37,9 +37,8 @@ class ConfigurationManagerTest extends ConfigTestCase
 
     public function tearDown()
     {
-        $this->getFileSystem()->remove($this->fixturesDir);
-
         chdir($this->currentDir);
+        $this->getFileSystem()->remove($this->fixturesDir);
     }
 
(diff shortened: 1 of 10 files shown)


marcj:
Oh, I guess we have very few people using Propel under Windows directly, so I believe there is nobody that can help you with that PostgreSQL issue. :) If you want, we could merge this and postpone window postgresql support into a later pull-request?

oujesky:
I share the same view, that there aren't many Propel and/or PostgreSQL users directly on Windows, so I guess we can leave it for some real use case (if there is any). I can make separate issue for this PostgreSQL/Windows problem.

marcj:
Well, alright, then let's do that. :) Thanks @oujesky!

PR #1067, opened 2015-12-10
Cache method information in PhpParser.

### LINKED ISSUE #1066: Model updating is slow
Hi,

When schema with multiple external schemas has very many tables the model updating is getting very slow. Would it be possible to create some kind of timestamp cache for the external schemas and update only the ones that have been changed? There might be issue with this when using custom behaviors but otherwise I think it would speed up the model updating a lot. What do you think?

staabm:
This speeds up generation process of large schemas considerably (~25%). refs #1066.

perf profiles (this numbers are not 100% accurate since more changes are included then this PR. See accurate numbers in https://github.com/propelorm/Propel2/pull/1067#issuecomment-163859751)
- before patch: https://blackfire.io/profiles/dd03dc97-cdc6-4f6f-bbcf-899cdc5b6918/graph (2 minutes and 13 seconds)
- after patch: https://blackfire.io/profiles/35114fdc-8aec-4ea2-ba31-e88d8bec865e/graph (1 minute and 40 seconds)

### DIFF
--- a/src/Propel/Generator/Util/PhpParser.php
+++ b/src/Propel/Generator/Util/PhpParser.php
@@ -36,6 +36,13 @@ class PhpParser
     protected $isAddPhp;
 
     /**
+     * methodName => methodCode
+     *
+     * @var string[]
+     */
+    private $knownMethodCache = array();
+
+    /**
      * Parser constructor
      *
      * @param string  $code     PHP code to parse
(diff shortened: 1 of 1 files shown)


staabm:
the last iteration of the patch is pending approval from @drodil

staabm:
@drodil confirmed the speed improvement in https://github.com/propelorm/Propel2/issues/1066#issuecomment-163852638

marcj:
ok very nice!

marcj:
We should keep track of the performance for every release :) to make sure we are not getting slower

staabm:
@marcj absolutely. continous perf would be great.

PR #1069, opened 2015-12-10
Speedup Database->getNextTableBehavior().

staabm:
This is more or less an experiment. could please someone double check I did not overlook something.

`GetNextTableBehaviour` is on the hot path of https://blackfire.io/profiles/35114fdc-8aec-4ea2-ba31-e88d8bec865e/graph?callname=Propel\Generator\Model\Database%3A%3AgetNextTableBehavior&selected=Propel\Generator\Model\Database%3A%3AgetNextTableBehavior&settings[dimension]=wt&settings[display]=focused&settings[tabPane]=nodes and this change tries to reduce the costs of it by getting rid of the `ksort`.

refs #1066

### DIFF
--- a/src/Propel/Generator/Model/Database.php
+++ b/src/Propel/Generator/Model/Database.php
@@ -753,16 +753,17 @@ class Database extends ScopedMappingModel
         // order the behaviors according to Behavior::$tableModificationOrder
         $behaviors = [];
         $nextBehavior = null;
+        $lowestModificationOrder = PHP_INT_MAX;
         foreach ($this->tables as $table) {
             foreach ($table->getBehaviors() as $behavior) {
                 if (!$behavior->isTableModified()) {
-                    $behaviors[$behavior->getTableModificationOrder()][] = $behavior;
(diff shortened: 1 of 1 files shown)


staabm:
@drodil could you please also try this change and profile it against your "big schema"

src/Propel/Generator/Model/Database.php
```diff
@@ -753,17 +753,18 @@ public function getNextTableBehavior()
         // order the behaviors according to Behavior::$tableModificationOrder
         $behaviors = [];
```
staabm: Todo: get rid of this now unused var

drodil:
Hi,

Did some profiling for you and here are the results.

Changes | Time | Profile
-------------|--------|----------
staabm@e87db60 | 1min, 43 seconds | https://blackfire.io/profiles/8e20e61f-7198-4c94-a347-5d8808bd7c74/graph
No changes | 1min, 43 seconds | https://blackfire.io/profiles/2da713fb-72f9-4aec-baa7-5db70b55246b/graph

No improvement at least in my case.

PR #1070, opened 2015-12-10
added a warning when xdebug is loaded in cli

staabm:
Xdebug being loaded reduces the runtime performance considerably (a factor up to 3). inform the user about it. refs #1066.

This was also a major problem when running composer, therefore we added this warning there like we do here.

see it in action: (Warning in the very first line after invoking the command)
xdebug

### DIFF
--- a/bin/propel.php
+++ b/bin/propel.php
@@ -8,10 +8,10 @@ if (!class_exists('\Symfony\Component\Console\Application')) {
     }
 }
 
-use Symfony\Component\Console\Application;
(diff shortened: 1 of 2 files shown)


staabm:
in case we are ok with linking a article which describes the issue further, we could add a link to https://getcomposer.org/xdebug

please tell me what you guys think

staabm:
@marcj any feedback?

marcj:
Yeah why not :)

PR #1077, opened 2015-12-11
optimize loop in propel.php

SwissalpS:
Don't need to redaclare $ns for every ReflectionClass.
Also avoid concat of '\\\\' in each iteration.


staabm:
this is just mirco and should be merged IMO

staabm:
args my last comment missed the "not" ;). nevermind.

marcj:
It shouldn't hurt :)

PR #1078, opened 2015-12-11
bug fix multiple 'XML Schema' with different 'database schema'

nibsirahsieu:
This PR fix bug when `xml schema` split into several files and contains different `database schema`. Before this patch, propel will set the same schema for all tables.


marcj:
Oh ok, didn't know about it, thanks :)

PR #1084, opened 2015-12-20
Fixed #1080. Made init command works in 2.3 up to 3.0.*

### LINKED ISSUE #1080: Symfony 3 issues
Fatal error: Class 'Symfony\Component\Console\Helper\DialogHelper' not found in ....vendor\propel\propel\src\Propel\Generator\Command\Helper\DialogHelper.php on line 8

Symfony 3 is missing the DialogHelper

SCIF:
Created layer between symfony's dialog (question in 3.0) helper and propel command. Should works in `2.3.*`-`3.0.*` fine. Symfony created similar layer called SymfonyStyle but it's not very good now and will replace QuestionHelper not soon.


marcj:
Thanks! @mpscholten could you please take a look at it? `init` command was primarily written by you :)

src/Propel/Generator/Command/InitCommand.php
```diff
@@ -50,17 +50,20 @@ protected function configure()
 
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        $this->getHelperSet()->set(new DialogHelper());
-
-        /** @var $dialog DialogHelper */
-        $dialog = $this->getHelper('dialog');
```
mpscholten: Maybe we can move this code block into a private `createDialogHelper` method? :)

src/Propel/Generator/Command/Helper/DialogHelper.php
```diff
@@ -49,13 +60,57 @@ private function formatQuestion($question, $default = null)
         }
     }
 
-    public function select(OutputInterface $output, $question, $choices, $default = null, $attempts = false, $errorMessage = 'Value "%s" is invalid', $multiselect = false)
+    public function select($question, $choices, $default = null, $attempts = false, $errorMessage = 'Value "%s" is invalid', $multiselect = false)
+    {
+        return parent::select(  $this->output,
```
mpscholten: Looks like this code has a wrong indentation :)

src/Propel/Generator/Command/Helper/DialogHelper3.php
```diff
@@ -0,0 +1,158 @@
...
+
+        return parent::ask($this->input, $this->output, $question);
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function askHiddenResponse($question, $fallback = true)
+    {
+        $question = new Question($question);
```
mpscholten: There's a `$this->formatQuestion` missing here. It should be `new Question($this->formatQuestion($question))` :) Otherwise the question is not followed by a `:` and also has no color.

mpscholten:
@SCIF otherwise everything looks good. Thanks for your contribution :)

src/Propel/Generator/Command/InitCommand.php
SCIF: @mpscholten and probably rename all classes/interface to ConsoleHelper* schema? It would be more correct.

mpscholten: Good suggestion :+1: Feel free to rename it :)

src/Propel/Generator/Command/Helper/DialogHelper.php
SCIF: @mpscholten , i aligned arguments because it's multiline constructor. Do you insist?

SCIF:
@mpscholten, check it now :)

src/Propel/Generator/Command/Helper/ConsoleHelper3.php
```diff
@@ -0,0 +1,158 @@
...
+        }
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function select($question, $choices, $default = null, $attempts = false, $errorMessage = 'Value "%s" is invalid', $multiselect = false)
+    {
+        $question = new ChoiceQuestion($this->formatQuestion($question, $default), $choices, $default);
+        $attempts && $question->setMaxAttempts($attempts);
```
mpscholten: Maybe we could use an `if`-statement here to be explicit? :)

SCIF: Using of it doesn't deny by coding standarts of propel afaik. It's not deprecated structure and not a bad practice. It's pretty short (therefore is good readable and understandable) and pretty beauty for me. Really don't want to renounce the use of it. I use it often. @marcj what do you think?

phlopsi: I don't mean to be rude, but there's a simple rule to contributing to open source projects:
coding style guide > consistency > good practice > personal opinion > bad practice

To me this falls under the category: "I have never seen this before in my 6-7 years of PHP developement experience. Is the CoffeScript inventor programming in PHP, now?", which, to me, indicates bad practice and just leads to confusion to anyone not accustomed to this kind of coding style.

mpscholten: @phlopsi thanks for a second opinion on this.

@SCIF as @phlopsi has pointed out it's not a common used practice in the php world, although the style is used in other programming languages. You will agree that code should be as readable as possible, and sometimes this means following conventions even when it not follows your personal style. I know this too well, sometimes following the convention sucks, but for open source software and software maintained by more than a single person in general, it is necessary to adapt ones coding style to the project. Remember: Your style is not wrong, it's just uncommon.

So let's just fix this and we're ready to merge :rocket:

marcj: I also love to use short-circuit evaluation and it's pretty to me as well, don't get me wrong, but especially in open-source I don't use such sugar because not all people are used to it, so I keep using it only in my personal projects. :)

SCIF:
I did it as people voted :)

mpscholten:
Great :) Then I guess we can merge @marcj :shipit:

src/Propel/Generator/Command/InitCommand.php
```diff
@@ -304,4 +283,23 @@ private function reverseEngineerSchema(OutputInterface $output, array $options)
...
+
+    /**
+     * @param InputInterface  $input
+     * @param OutputInterface $output
+     *
+     * @return ConsoleHelperInterface
+     */
+    protected function createConsoleHelper(InputInterface $input, OutputInterface $output)
+    {
+        if (class_exists('\Symfony\Component\Console\Helper\QuestionHelper', true)) {
```
marcj: second argument `true` is here default, no need to specify it.

marcj: A line comment above 'this class only exists in Symfony 3. So if it exists we use our ConsoleHelper3 to work with Symfony 3 correctly' or something.

SCIF:
@marcj, anything else? A nitpicking of last 3 days just wastes time :( One positive moment is that i worked on adapter (PropelLaravel) testing and found input stream issue, which keep in mind and fixed it.

Issue #1082, opened 2015-12-15
PostgreSQL and Windows


oujesky:
PostgreSQL for some unknown reason fails few tests on Windows - mostly because of different quoting - looks like the Windows version ignores binding values as integers and treats them as strings instead.

That leads to following type of errors:

```
Failed asserting that two strings are equal.
Expected :'INSERT INTO "group" ("id", "title", "as") VALUES (1, 'Test Group', 3)'
Actual   :'INSERT INTO "group" ("id", "title", "as") VALUES ('1', 'Test Group', 3)'
```

Discussed in #1064, that this is a low priority issue.

Tested on Windows 7, PHP 5.6

nibsirahsieu:
i'm just hitting this issue. My environment: Ubuntu 12.04, php 5.5.30, postgres 9.1.19.

Here are the steps i did:

- rm tests/Fixtures/fixtures_built
- DB=pgsql DB_HOSTNAME=localhost DB_USER=XXX DB_PW=XXX DB_NAME=propel2_test ./tests/bin/setup.pgsql.sh
- DB=pgsql DB_HOSTNAME=localhost DB_USER=XXX DB_PW=XXX DB_NAME=propel2_test ./vendor/bin/phpunit

```php

There were 7 failures:

1) Propel\Tests\Generator\Builder\Om\GeneratedObjectTest::testSetterOneToManyWithFkRequired
Failed asserting that 1 matches expected 3.

/xxx/Propel2/tests/Propel/Tests/Generator/Builder/Om/GeneratedObjectTest.php:1471

2) Propel\Tests\Generator\Builder\Om\GeneratedObjectWithFixturesTest::testEquals
Expected Book objects to be equal()
Failed asserting that false is true.

/xxx/Propel2/tests/Propel/Tests/Generator/Builder/Om/GeneratedObjectWithFixturesTest.php:230

3) Propel\Tests\Generator\Command\DatabaseReverseTest::testCommand
Failed asserting that actual size 0 matches expected size 1.

/xxx/Propel2/tests/Propel/Tests/Generator/Command/DatabaseReverseTest.php:53

4) Propel\Tests\Runtime\ActiveQuery\ModelCriteriaTest::testMagicFindByObject
findByXXX($value) is turned into findBy(XXX, $value)
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
[...]

PR #1085, opened 2015-12-20
[tests] fixed bug when propel trying to execute database migration

nibsirahsieu:
Rename `table1` and `table2` into `table1_ts` and `table2_ts` respectively, as `table1` already defined in another schema (`tests/Fixtures/migration-command/schema.xml`).

```
Executing statement "CREATE TABLE "table1"
(
    "id" serial NOT NULL,
    "title" VARCHAR NOT NULL,
    "table2_id" INTEGER,
    PRIMARY KEY ("id")
)"
                                                
  [Propel\Runtime\Exception\RuntimeException]   
  Failed to execute SQL "CREATE TABLE "table1"  
  (                                             
      "id" serial NOT NULL,                     
      "title" VARCHAR NOT NULL,                 
      "table2_id" INTEGER,                      
      PRIMARY KEY ("id")                        
  )". Aborting migration.

```
PS: This happen only when i'm running tests using postgres

### DIFF
--- a/tests/Fixtures/bookstore/behavior-timestampable-schema.xml
+++ b/tests/Fixtures/bookstore/behavior-timestampable-schema.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="ISO-8859-1" standalone="no"?>
 <database name="bookstore-behavior" defaultIdMethod="native" package="behavior.timestampable" namespace="Propel\Tests\Bookstore\Behavior">
 
-    <table name="table1">
+    <table name="table1_ts">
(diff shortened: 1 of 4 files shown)


marcj:
Isn't there then a deeper big in schema name usage inside Propel core? We have obviously not issues in our test suite. Maybe you have a newer/older PostgreSQL where some stuff with our schema naming isn't working correctly?

nibsirahsieu:
Here are My environment: Ubuntu 12.04, php 5.5.30, postgres 9.1.19. I don't know, may be there are something wrong with local environment, also i'm experienced issue as described here #1082.

nibsirahsieu:
closing this.. i'll investigate this issue later

PR #1088, opened 2015-12-24
Fix Sqlite persister and Behavior locator

cristianoc72:
SQLite keeps track of auto-increment fields in an internal table, called `sqlite_sequence`.
https://www.sqlite.org/fileformat2.html#seqtab

With this commit, `Propel\Runtime\Persister\SQL\SQLitePersister::readAutoIncrement` method
takes values from this table, preventing `FOREIGN KEY violation` errors.
This commit also provides a test, which failed with previous versions of the persister.
The previous algorithm, to read auto-increment fields, is still used if there's no `sqlite_sequence` table entry, for the given field.

### DIFF
--- a/src/Propel/Generator/Util/BehaviorLocator.php
+++ b/src/Propel/Generator/Util/BehaviorLocator.php
@@ -174,7 +174,7 @@ class BehaviorLocator
      */
     private function getCoreBehavior($name)
     {
-        $phpName = NamingTool::toCamelCase($name);
+        $phpName = ucfirst(NamingTool::toCamelCase($name));
 
         return sprintf('\\Propel\\Generator\\Behavior\\%s\\%sBehavior', $phpName, $phpName);
     }
--- a/src/Propel/Runtime/Persister/SQL/SQLitePersister.php
+++ b/src/Propel/Runtime/Persister/SQL/SQLitePersister.php
@@ -3,26 +3,68 @@
 namespace Propel\Runtime\Persister\SQL;
 
 use Propel\Runtime\Connection\ConnectionInterface;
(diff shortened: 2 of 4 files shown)


marcj:
Very nice, cool :)

src/Propel/Generator/Util/BehaviorLocator.php
```diff
@@ -174,7 +174,7 @@ public function getBehavior($name)
      */
     private function getCoreBehavior($name)
     {
-        $phpName = NamingTool::toCamelCase($name);
+        $phpName = ucfirst(NamingTool::toCamelCase($name));
```
staabm: shouldnt the naming tool already provide the correct name (otherwise it doesnt deserve its name)?

src/Propel/Runtime/Persister/SQL/SQLitePersister.php
```diff
@@ -3,32 +3,28 @@
...
-EOF;
 
         $autoIncrementField = current($this->entityMap->getPrimaryKeys());
-
-        $columnName = $autoIncrementField->getColumnName();
         $tableName = $autoIncrementField->getEntity()->getFQTableName();
-        $sql = sprintf($sql, $columnName, $tableName, $columnName);
 
-        $stmt = $connection->prepare($sql);
+        $stmt = $connection->prepare("SELECT seq FROM sqlite_sequence WHERE name = '$tableName'");
```
staabm: You are using a prepared statment, why dont you use a placeholder?

cristianoc72:
@marcj imho I should consider @staabm 's suggestions: do you prefer to revert this pr or I'll include them into my next one?

Issue #1087, opened 2015-12-23
[PgsqlSchemaParserTest] testParse


nibsirahsieu:
I found this oddities when i want to create test for `PgsqlSchemaParser.php`. When query on the line https://github.com/propelorm/Propel2/blob/master/tests/Propel/Tests/Generator/Reverse/PgsqlSchemaParserTest.php#L69 is executed, table `foo` will be created in the `public` schema. Now have a look at `parseTables` function of class `PgsqlSchemaParser`. When `$database` variable doesn't have `schema`, this line will be executed https://github.com/propelorm/Propel2/blob/master/src/Propel/Generator/Reverse/PgsqlSchemaParser.php#L150, the final sql query is 
```
SELECT c.oid, c.relname, n.nspname
          FROM pg_class c join pg_namespace n on (c.relnamespace=n.oid)
          WHERE c.relkind = 'r'
            AND n.nspname NOT IN ('information_schema','pg_catalog')
            AND n.nspname NOT LIKE 'pg_temp%'
            AND n.nspname NOT LIKE 'pg_toast%'
            AND n.nspname IN ('bookstore_schemas', 'contest', 'second_hand_books', 'migration')
          ORDER BY relname
```
When https://github.com/propelorm/Propel2/blob/master/tests/Propel/Tests/Generator/Reverse/PgsqlSchemaParserTest.php#L78 is executed, logically the value of variable `$table` must be NULL, because table `foo` was created in the `public` schema, and calling `$table->getColumns();` should throw an exception. But why the test pass in both  circleci and  travisci?.

marcj:
Good question :) I guess because `public` is the default namespace in those CI-environments.

nibsirahsieu:
i've just made PR here #1089

PR #1089, opened 2015-12-24
[PgsqlSchemaParserTest] set database schema to 'public'

nibsirahsieu:
Please see #1087 for detail information. Thanks


marcj:
Thanks! Strange bugfix, but if it fixes then ok :)

PR #1092, opened 2016-01-02
Fake PropelDateTime for filter like date

jeremydenoun:
proposal for add fallback on a fake DateTime object if value contain % symbol for LIKE query condition in filter


marcj:
I'm not sure why you need this? Why not just create a regular DateTime object?

jeremydenoun:
I try to make filter on a "datetime" field with this kind of syntax : `filterByDate("2016-%")` (valid in sql)

If we create regular DateTime not empty the check is :
`if (null === $timeZone) {
        // stupid DateTime constructor signature
        $dateTimeObject = new $dateTimeClass($value);
} else {
        $dateTimeObject = new $dateTimeClass($value, $timeZone);
}`

but in this case I have an exception (because this isn't a DateTime format)
so my naive workaround is to create a fake object implement only format method (that return constructor string) for be compatible with existing code and this work correctly.

On other approach could be to generate specific "filterBy" method for datetime field (like filterByDate_Month/Year/Day....) but this can be more work on not enough flexible.

If you think this kind of needs should be implement with ->where('Date = ?', '2016-%') you can refuse my PR, but I think this is more readable with my version.

rafaelnery:
I use extract with where condition

```PHP
$book = new Book();
$book->setField(new DateTime('2016-07-27'));
$book->save();

$query = BookQuery::create()
  ->where('extract(month from book.field) = ?', 7, PDO::PARAM_INT)
  ->find();
```
[Demo in Sandbox](http://sandbox.propelorm.org/69b3656)

---
**BUT, ONLY IF NECESSARY**

Why not change the type of filter then

```PHP
$query->filterByDateField(['month' => '12']);
```
or 

```PHP
$query->filterByDateField(new DateTime('NOW'), array('month'));
```

In both, propel could use [extract](http://dev.mysql.com/doc/refman/5.7/en/date-and-time-functions.html#function_extract)

jeremydenoun:
my use case is :
- I implement a full-text search from a single text input
- I use filterBy({field}, {input search}) _type independent_
- I'd like to be the nearest of ORM method (filterBy) for security/optimization _else I can use directly PDO request_

I can detect and translate with field type detection into query like :
```
->where('extract(month from book.field) = ?', 7, PDO::PARAM_INT)
->_or()
->where('extract(day from book.field) = ?', 7, PDO::PARAM_INT)
->_or()
->where('extract(year from book.field) = ?', 7, PDO::PARAM_INT)
```
but this isn't flexible and performant (try to implement 201_-1_-%)

So, I think Propel should allow to use database embed features,
maybe a better approach can be to add a parameter to filterBy function or add new function (more clear) for filter without type checking.

Last Thing, I try your first solution and I have find a bug :
```
$query = BookQuery::create()
  ->where('extract(month from book.field) = ?', 7, PDO::PARAM_INT)
  ->find();
echo $query->toString();
```;
=> Undefined index: column (src/Propel/Runtime/ActiveQuery/Criteria.php 1514)

After read code, I propose a fix on my last PR, I update my PR for support '_' too

PR #1099, opened 2016-01-11
Added: config unique constraint of slug_column

eelkevdbos:
Added the ability to configure whether or not to use a unique constraint on the slug_column of the table for the SluggableBehavior.

+ some additional docblock fixes


marcj:
Very nice, thanks!

PR #1103, opened 2016-01-14
Add identifierQuoting to archive behavior tables

thomaschaaf:
Archive behavior was not acknowledging the identifierQuoting parameter on the parent table.


marcj:
Beschde Danke!

PR #1104, opened 2016-01-14
removed php5.4 from travis build.

staabm:
5.4 is EOL.


staabm:
we can drop even more later on, but this one is a low hanging fruit.

Big-Shark:
+1

PR #1110, opened 2016-01-20
Use complex field in select with join if identifierQuoting="true"

jeremydenoun:
I work on propel with identifierQuoting="true" and I try to make a propel query with join field = A.b : 
$query->withColumn($field) or sorting by join field or where condition with join field

I have an issue on find() method because in Criteria::createSelectSql propel try to fetch columnName and tableName and try to link to TableMap column definition, in this case ColumnMap::normalizeName() for explode and fetch column name however with identifierQuoting true the field is convert to `A`.`b` so TableMap::getColumn() fail on hasColumn `b` because the column name is b (without quote)

My workaround is to remove all kind of escaping quote ` for mysql, ' for mssql with trim function (we shouldn't have space too) in normalize function, this can be more elegant with an "unquote" adapter method or change TableMap::getColumn() code but this can less performant and more harder..

### DIFF
--- a/src/Propel/Runtime/Map/ColumnMap.php
+++ b/src/Propel/Runtime/Map/ColumnMap.php
@@ -512,6 +512,7 @@ class ColumnMap
     {
         if (false !== ($pos = strrpos($name, '.'))) {
             $name = substr($name, $pos + 1);
+	     $name = trim($name," \t\n\r\0\x0B`'()\"[]~!-{}%^&.");
         }
 
         return strtoupper($name);


marcj:
Guess we should also include there other quote chars like `'`, `"` and `[`, `]`

src/Propel/Runtime/Map/ColumnMap.php
```diff
@@ -502,6 +502,7 @@ public static function normalizeName($name)
     {
         if (false !== ($pos = strrpos($name, '.'))) {
             $name = substr($name, $pos + 1);
+	     $name = trim($name,"\t\n\r\0\x0B`'");
```
staabm: in case we strip things like nullbytes etc. should we also strip other control chars?

staabm: -> I dont think we should strip nullbyte etc

jeremydenoun: in http://php.net/manual/fr/function.trim.php default value for trim is " \t\n\r\0\x0B" (so I extend it because space can't be used in column definition especially unquoted)

if I check main documentation page for column name unquoted, we can see : 
- **mysql** (http://dev.mysql.com/doc/refman/5.7/en/identifiers.html) : ASCII: [0-9,a-z,A-Z$_] (basic Latin letters, digits 0-9, dollar, underscore)
- **postgres** (http://www.postgresql.org/docs/9.3/static/sql-syntax-lexical.html): Subsequent characters in an identifier or key word can be letters, underscores, digits (0-9), or dollar signs ($)
- **sqlite** shouldn't be a problem (http://stackoverflow.com/questions/3694276/what-are-valid-table-names-in-sqlite / https://www.sqlite.org/lang_keywords.html)
- **oracle** (https://docs.oracle.com/cd/B19306_01/server.102/b14200/sql_elements008.htm _but Nonquoted identifiers can contain only alphanumeric characters from your database character set and the underscore (__)_, dollar sign ($), and pound sign (#). Database links can also contain periods (.) and "at" signs (@). Oracle strongly discourages you from using $ and # in nonquoted identifiers._)
- **mssql**  (https://msdn.microsoft.com/en-us/library/ms176027.aspx?f=255&MSPPError=-2147217396) support exotic delimiter ( "(" ")" ~  ! - { } % ^ ' & . )

so I propose 2 options :

1. we can trim only quoting char like @marcj propose (' " ` [ ]) and "(" + ")"  ~  ! - { } % ^ ' & . (for mssql)
2. we can remove all char non compliant with column name definition (in all databases supported)

I think 1. is more clear and easily to maintain, 2. respect better database syntax but can create side-effect

I can implement your prefer solution and PR

marcj: Then let's go the 1. way :)

marcj:
Would be awesome if you could squash your commits, so I can merge it =)

marcj:
Thanks :)

Issue #1115, opened 2016-01-31
Adding an index to a foreign-keyed MySQL column corrupts the entire table


ExplodingCabbage:
Building this schema on MySQL:

```
<?xml version="1.0" encoding="UTF-8"?>
<database name="exampledb">
    <table name="foo">
        <column name="bar_id" type="integer"/>
        <column name="other_column" type="integer"/>
        <foreign-key foreignTable="bar">
            <reference local="bar_id" foreign="id"/>
        </foreign-key>
    </table>
    <table name="bar">
        <column name="id" type="integer" primaryKey="true"/>
    </table>
</database>
```

... and then modifying it by adding an explicit key that includes `bar_id`, along with some other change that will cause an ALTER TABLE statement to be generated:

```
<?xml version="1.0" encoding="UTF-8"?>
<database name="exampledb">
    <table name="foo">
        <column name="bar_id" type="integer"/>
        <column name="other_column" type="integer"/>
        <column name="new_column" type="integer"/>
        <foreign-key foreignTable="bar">
            <reference local="bar_id" foreign="id"/>
        </foreign-key>
        <index>
            <index-column name="bar_id"/>
            <index-column name="other_column"/>
        </index>
    </table>
    <table name="bar">
        <column name="id" type="integer" primaryKey="true"/>
    </table>
</database>
```

... makes everything go boom, leaving the database schema in a corrupted state:

```
mark@lunchbox:~$ sudo Propel2/bin/propel diff
2 tables found in all databases.
Comparing models...
Structure of database was modified in datasource "exampledb": 1 modified tables
"/home/mark/generated-migrations/PropelMigration_1454198980.php" file successfully created.
Please review the generated SQL statements, and add data migration code if necessary.
Once the migration class is valid, call the "migrate" task to execute it.
mark@lunchbox:~$ sudo Propel2/bin/propel migrate
Executing migration PropelMigration_1454198980 up

                                                                         
  [Propel\Runtime\Exception\RuntimeException]                            
  Failed to execute SQL "ALTER TABLE `foo`                               
    ADD `new_column` INTEGER AFTER `other_column`". Aborting migration.  
                                                                         

                                                                                                                       
  [PDOException]                                                                                                       
  SQLSTATE[HY000]: General error: 1025 Error on rename of './exampledb/#sql-483_31' to './exampledb/foo' (errno: 150)  
                                                                                                                       

[...]

PR #1116, opened 2016-01-31
Make MySQL table modifications always use one statement to fix total table-corruption bug #1115

ExplodingCabbage:
There are three commits in here. The first adds a test that proves https://github.com/propelorm/Propel2/issues/1115. The second fixes it by changing the way that Propel generates DDL statements for MySQL to always modify a table with a single big statement instead of many. This breaks lots of tests that were simply asserting on the exact DDL output. The third commit changes all those tests so that they pass again.

### DIFF
--- a/src/Propel/Generator/Platform/MysqlPlatform.php
+++ b/src/Propel/Generator/Platform/MysqlPlatform.php
@@ -22,6 +22,7 @@ use Propel\Generator\Model\Table;
 use Propel\Generator\Model\Unique;
 use Propel\Generator\Model\Diff\ColumnDiff;
 use Propel\Generator\Model\Diff\DatabaseDiff;
+use Propel\Generator\Model\Diff\TableDiff;
 
 /**
  * MySql PlatformInterface implementation.
@@ -451,13 +452,13 @@ ALTER TABLE %s DROP PRIMARY KEY;
     public function getAddIndexDDL(Index $index)
     {
         $pattern = "
-CREATE %sINDEX %s ON %s (%s);
+ALTER TABLE %s ADD %sINDEX %s (%s);
 ";
 
         return sprintf($pattern,
+            $this->quoteIdentifier($index->getTable()->getName()),
             $this->getIndexType($index),
             $this->quoteIdentifier($index->getName()),
-            $this->quoteIdentifier($index->getTable()->getName()),
             $this->getIndexColumnListDDL($index)
         );
     }
@@ -471,12 +472,12 @@ CREATE %sINDEX %s ON %s (%s);
     public function getDropIndexDDL(Index $index)
     {
         $pattern = "
-DROP INDEX %s ON %s;
+ALTER TABLE %s DROP INDEX %s;
 ";
 
         return sprintf($pattern,
-            $this->quoteIdentifier($index->getName()),
-            $this->quoteIdentifier($index->getTable()->getName())
+            $this->quoteIdentifier($index->getTable()->getName()),
+            $this->quoteIdentifier($index->getName())
         );
     }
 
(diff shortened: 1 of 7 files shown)


ExplodingCabbage:
Oops, don't merge this yet. Need to fix some comments.

marcj:
Wow very nice, thanks! I'm waiting for travis and your go :)

ExplodingCabbage:
Cool, I've nuked a comment that I think was nonsense about needing to rerun the database setup scripts if `IssueTest1115.php` fails due to it corrupting the database. I did observe *some* weird behaviour while testing where I'd get into a state where nearly all the tests failed, but I'm not sure what I did to get there or whether it was unique to my test or not. (I fixed it when it happened by doing a `git clean -fdX`, rerunning `composer install`, and rerunning the setup script; just rerunning the setup script as my purged comment suggested wasn't sufficient.)

As far as I know, this is good to go, although since this is my first dive into Propel's internals an extra pair of eyes over it would certainly be welcome. :)

marcj:
Cool! MySQL fails in the test suite due to [one test](https://github.com/propelorm/Propel2/blob/master/tests/Propel/Tests/Issues/Issue617Test.php#L155-L163) which has been converted to the all-in-once ALTER TABLE statement.

ExplodingCabbage:
Oops! I'm simply running `./vendor/bin/phpunit` and don't see that failure; am I skipping that test by default?

ExplodingCabbage:
Just spotted another mistake - I haven't marked Issue1115 as being `@group mysql`. Sorry, will fix!

ExplodingCabbage:
Okay, I rebased to fix the group on Issue1115Test.php. Now, to fix the other test...

ExplodingCabbage:
Aha, so I understand why I wasn't seeing the failure before - I wasn't running the tests with `--group mysql`. D'oh.

ExplodingCabbage:
I think everything should pass, now. :)

marcj:
Here we got, very nice! Looks good to me, just want to wait for our chief review officer @staabm :raised_hands:

staabm:
> chief review officer

;-)

looks really good. having the statements combined into one alter also makes it "more atomar" which is great, in case no transactions are used.

Didnt test it in particular, but feels good to go.

src/Propel/Generator/Platform/MysqlPlatform.php
```diff
@@ -717,10 +718,96 @@ public function getAddColumnDDL(Column $column)
         );
     }
 
+    public function getModifyTableDDL(TableDiff $tableDiff)
+    {
+        $alterTableStatements = '';
+
+        $toTable = $tableDiff->getToTable();
```
staabm: this var is only used in one certain case below, therefore I would move it more to this location.

ExplodingCabbage: Sure. I think I did it this way to keep it similar to [the DefaultPlatform version](https://github.com/propelorm/Propel2/blob/master/src/Propel/Generator/Platform/DefaultPlatform.php#L765), but I've no real preference either way; I'll make the change tonight when I'm home from work.

(Alternatively, feel free to merge and make the change yourself.)

ExplodingCabbage: Just got home after a couple of beers and had a chance to take a look at this. On a closer glance I think I stand by my original placement; the variable is actually used twice lower down in the function, and the place where it's used is very similar to the three places where it's used in the `DefaultPlatform` implementation. I'll still move it if you want, *but* I think if we're going to move it we should move it in `DefaultPlatform`'s `getModifyTableDDL()` too; there's no reason I can see to use a different style in the two very similar functions.

marcj: don't let us be petty and ignore this one :)

marcj:
@ExplodingCabbage thanks very much, very nice contribution :) :raised_hands: :+1:

PR #1119, opened 2016-02-03
Fix --object-class option not being considered when building model

bcbrr:
Typo-ed option name when reading the option.


marcj:
Good catch thanks

PR #1120, opened 2016-02-05
Fix $getter mixup in sluggable behaviour generator

daniel-zahariev:
When a `scope_column` is present in the config of the behaviour the `$getter` gets overwritten which breaks things.

### DIFF
--- a/src/Propel/Generator/Behavior/Sluggable/SluggableBehavior.php
+++ b/src/Propel/Generator/Behavior/Sluggable/SluggableBehavior.php
@@ -330,9 +330,9 @@ protected function makeSlugUnique(\$slug, \$separator = '" . $this->getParameter
         ->prune(\$this)";
 
         if ($this->getParameter('scope_column')) {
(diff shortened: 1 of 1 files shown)


marcj:
This generates the exact same code, so this couldn't fix stuff you mentioned. I'm closing, but if you find some time to fix it we could reopen it.

daniel-zahariev:
The reason for this pull request is that when there's a `scope_column` setup for the `sluggable ` behaviour the `$getter` variable gets polluted and on line [366](https://github.com/propelorm/Propel2/blob/master/src/Propel/Generator/Behavior/Sluggable/SluggableBehavior.php#L366) it outputs something like (depending on the name of the scope column):
```
$slugNum = substr($object->getScope(), strlen($slug) + 1);
```
which obviously isn't good.

marcj:
Ah, now I got it. sorry for the misunderstanding! I though you submitted something weird. :P

marcj:
thanks!

PR #1132, opened 2016-02-29
Log function in model deprecated now

Big-Shark:
We should not loging anything in the model, and I think we need declared this function as deprecated, and will remove later.

### DIFF
--- a/src/Propel/Generator/Builder/Om/templates/baseObjectMethods.php
+++ b/src/Propel/Generator/Builder/Om/templates/baseObjectMethods.php
@@ -171,6 +171,8 @@
      * @param  string  $msg
      * @param  int     $priority One of the Propel::LOG_* logging levels
      * @return boolean
+     *
+     * @deprecated
      */
     protected function log($msg, $priority = Propel::LOG_INFO)
     {


SCIF:
Agreed. If PR will be merged — don't forget to do PR in docs to fix that [http://propelorm.org/documentation/07-logging.html#logging-messages](paragraph).

src/Propel/Generator/Builder/Om/templates/baseObjectMethods.php
```diff
@@ -171,6 +171,8 @@ public function setVirtualColumn($name, $value)
      * @param  string  $msg
      * @param  int     $priority One of the Propel::LOG_* logging levels
      * @return boolean
+     *
+     * @deprecated
```
SCIF: It's good practice to write version in which method will be deleted and message like `Consult http://propelorm.org/documentation/07-logging.html`

marcj:
I don't see the immediate need for this. Removing now some active methods would not make sense, so let us rather concentrate on `data-mapper` branch which does not have such stuff.

PR #1133, opened 2016-03-02
Problem with ObjectCollection::removeObject

kolah:
Hello. I ran into a problem while using ObjectCollection. Subsequential use of `ObjectCollection::removeObject` and `ObjectCollection::append` results with wrong index mapping, which causes that improper items are removed and left in the collection.

I have provided simple test case to reproduce this issue.

It seems it is caused by optimistic assumption in `ObjectCollection::append` that `$pos = count($this->data);`

### DIFF
--- a/tests/Propel/Tests/Issues/Issue1133Test.php
+++ b/tests/Propel/Tests/Issues/Issue1133Test.php
@@ -1,0 +1,84 @@
+<?php
+
+namespace Propel\Tests\Issues;
+
+use Propel\Tests\TestCase;
+use Propel\Runtime\Collection\ObjectCollection;
+
+class DummyObject
+{
+    private $id;
+
+    public function __construct($id)
+    {
+        $this->id = $id;
+    }
+
+    public function getId()
+    {
+        return $this->id;
+    }
+
+    public function hashCode()
+    {
+        return (string)$this->id;
+    }
+}
+
+/**
+ * This test proves the bug described in https://github.com/propelorm/Propel2/issues/1133.
+ *
+ * @group database
+ */
+class Issue1133Test extends TestCase
+{
+
+    public function testIssue1133Append()
+    {
+        $testCollection = new ObjectCollection;
+        $testCollection->setModel(DummyObject::class);
+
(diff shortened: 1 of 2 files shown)


tests/Propel/Tests/Issues/Issue1133Test.php
```diff
@@ -0,0 +1,60 @@
...
+
+class DummyObject 
+{
+    private $id;
+    
+    private static $counter = 0;
+    
+    public function __construct()
+    {
+        $this->id = self::$counter++;
```
prgTW: `new DummyObject($i)` would be much more readable ;)

kolah: Thanks for suggestion :)

staabm:
patch looks good to me. did you make sure that the unit test does not pass without your patch?

prgTW:
@staabm https://travis-ci.org/propelorm/Propel2/jobs/113011705 for commit bff5b99 (test case without the fix) doesn't pass so I assume yes :)

staabm:
I ask because you changed the test after this comit.

kolah:
@staabm The test was first, in both cases.

kolah:
I have separated commits with tests and fix - I had to change the way of calculating last inserted index value.

marcj:
Thanks everyone for the fix! I'm not sure about that yet, because it looks kinda like its a performance decrease to me since this `offsetSet` and `append` are heavily used and I'm not sure if `end()` and `key()` in combination are fast enough. I thought it could be faster by using array_keys() and then retrieve directly the last array item by its index.

kolah:
Do you mean calculating it on each append? I think it's faster using `key()` and 'end()' way, because it only operates on internal pointer of a data array. Here's a simple benchmark: [https://3v4l.org/JeATl](https://3v4l.org/JeATl)

marcj:
@kolah very nicely done, thanks! Then I guess we should merge this beauty :)

kolah:
Great! Thanks :)

PR #1142, opened 2016-03-13
Fix propel init fails when password has special chars

### LINKED ISSUE #1141: propel init fails when password or username has special characters like #
Usernames and passwords are urlencoded, but not decoded.
https://github.com/propelorm/Propel2/blob/master/src/Propel/Generator/Command/InitCommand.php#L267

https://github.com/propelorm/Propel2/blob/master/src/Propel/Generator/Command/AbstractCommand.php#L118

https://github.com/propelorm/Propel2/blob/master/src/Propel/Generator/Command/AbstractCommand.php#L151

benbankes:
Fixes #1141

### DIFF
--- a/src/Propel/Generator/Command/InitCommand.php
+++ b/src/Propel/Generator/Command/InitCommand.php
@@ -264,7 +264,7 @@ class InitCommand extends AbstractCommand
         $outputDir = sys_get_temp_dir();
 
         $this->getApplication()->setAutoExit(false);
-        $fullDsn = sprintf('%s;user=%s;password=%s', $options['dsn'], urlencode($options['user']), urlencode($options['password']));
+        $fullDsn = sprintf('%s;user=%s;password=%s', $options['dsn'], $options['user'], $options['password']);
 
         $input = new ArrayInput([
             'reverse',


src/Propel/Generator/Command/InitCommand.php
```diff
@@ -264,7 +264,7 @@ private function reverseEngineerSchema(OutputInterface $output, array $options)
         $outputDir = sys_get_temp_dir();
 
         $this->getApplication()->setAutoExit(false);
-        $fullDsn = sprintf('%s;user=%s;password=%s', $options['dsn'], urlencode($options['user']), urlencode($options['password']));
+        $fullDsn = sprintf('%s;user=%s;password=%s', $options['dsn'], $options['user'], $options['password']);
```
staabm: Do we need to encode the `;` somehow?

marcj: oh true, we need to urlencode the values still

benbankes: @marcj The `urlencode` function breaks authentication like this...  If I go through `propel init` using the password `H7{"Qj1n>P` The following command
`echo urlencode('H7{"Qj1n>P');` results in `H7%7B%22Qj1n%3EP`
which will not authenticate to a password to a database with the password `H7{"Qj1n>P`

If this is on purpose, then this only allows us to use the set of three special characters `.-_` when using propel init, and this should included in the documentation.  Would you like me to update the documentation?

@staabm We already use a non-encoded password in \Propel\Generator\Command\InitCommand.php::testConnection.
See https://github.com/propelorm/Propel2/blob/master/src/Propel/Generator/Command/InitCommand.php#L92
and
https://github.com/propelorm/Propel2/blob/master/src/Propel/Generator/Command/InitCommand.php#L95

Should I change the code to encoding that as well?  I will look into other encoding functions besides urlencode, since urlencode does not allow authentication with most special characters.  If you have any suggestions, please suggest.

benbankes: @marcj If we refactor some code, we can pull the username/password out of the dsn.  I would be happy to take a stab at that if that is the way you want to go.  That will be a better solution than simply removing urlencode from values to be placed in the dsn string.

benbankes: @staabm I was unable to find any encoding scheme that works with MySQL authentication.

staabm: @benbankes we only have this problem when we parse the data out of a dsn string or into one.

how does pdo handle passwords which contain a `;`?

benbankes: @staabm PDO handles a password with a `;` by not including the password in the dsn.  (I think this answers your question... tell me if it doesn't.)

The function signature is:
`PDO::__construct ( string $dsn [, string $username [, string $password [, array $options ]]] )`
See
http://php.net/manual/en/pdo.construct.php

In fact, the dsn specification for for MySQL at
http://php.net/manual/en/ref.pdo-mysql.connection.php
does not include the password in the dsn string at all.

This is how Propel uses PDO internally as well.
https://github.com/propelorm/Propel2/blob/master/src/Propel/Runtime/Connection/PdoConnection.php#L47

Does that answer your question?

staabm: oh I see. I guess we shouldn't do either then and pass user+pw as separate parameters.

@marcj wdyt?

marcj: Yes, if this is working as expected then we should go with the password not being in the DSN. :+1:

PR #1146, opened 2016-03-22
Added user and password options to the slaves configuration

dimitriyremerov:
This fixes a problem in which propel always tried to connect with empty username/password for all read queries


marcj:
Thanks! I'm not sure about that because user/password is not used in the code? Just defining it shouldn't fix it, should it?

dimitriyremerov:
It works because it uses the same configuration methods.
I also tested it. When I added manually two lines in generated-conf/config.php it started working for read. The problem was that YML configuration was invalid.
I also removed all Propel-generated files in my project and rebuild it again, this time the correct config was there automatically.

marcj:
Ah, cool! Thanks :)

PR #1147, opened 2016-03-22
Prevented function redeclaration error by skipping reserved methods

dimitriyremerov:
This fix prevents propel to generate duplicated methods such as isDeleted() or isNew()
The getters will still be available via getIsDeleted() etc.

### DIFF
--- a/src/Propel/Generator/Builder/Om/ObjectBuilder.php
+++ b/src/Propel/Generator/Builder/Om/ObjectBuilder.php
@@ -1015,6 +1015,10 @@ abstract class ".$this->getUnqualifiedClassName().$parentClass." implements Acti
      */
     protected function addBooleanAccessor(&$script, Column $column)
     {
+        $name = $column->getCamelCaseName();
+        if (in_array($name, ClassTools::getPropelReservedMethods())) {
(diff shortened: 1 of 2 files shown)


src/Propel/Generator/Builder/Om/ObjectBuilder.php
```diff
@@ -1015,6 +1015,10 @@ protected function addArrayAccessorBody(&$script, Column $column)
      */
     protected function addBooleanAccessor(&$script, Column $column)
     {
+        $name = $column->getCamelCaseName();
+        if (in_array($name, ClassTools::getPropelReservedMethods())) {
+            return; // Skip boolean accessors for reserved names
+        }
```
staabm: Instead of silently skipping should we print a error instead?

dimitriyremerov: We do not have a proper way of issuing warnings now in the builder, only throwing errors. I do believe that boolean accessors are not mandatory and a warning can be issued suggesting the user to use the normal accessor.

marcj:
Great, thanks :)

PR #1148, opened 2016-03-25
findPk should not use instance pool for non empty queries

### LINKED ISSUE #952: ObjectFormatter ignores the fields specified by the "select" method on a query if the instance is already in the instance pool
Say you use the ObjectFormatter everywhere. You run these two queries:

```
$query1 = BookQuery::create()->findPk(12);
var_dump($query1);

$query2 = BookQuery::create()->select(['Book.Name', 'Book.Summary'])->findPk(12);
var_dump($query2);
```

When running $query2 alone, you are supposed to get an `array('Book.Name' => ..., 'Book.Summary' => ...)`. However if a Book with id 12 is already in the instance pool of the formatter (i.e. you ran $query1 beforehand), you'll get an object back instead of the array.

This is not reproducible with other formatters AFAIK.

gharlan:
fixes #952

### DIFF
--- a/src/Propel/Generator/Builder/Om/QueryBuilder.php
+++ b/src/Propel/Generator/Builder/Om/QueryBuilder.php
@@ -264,7 +264,7 @@ abstract class ".$this->getUnqualifiedClassName()." extends " . $parentClass . "
                     $this->addFilterBySetCol($script, $col);
                 }
             }
-            
+
         }
         foreach ($this->getTable()->getForeignKeys() as $fk) {
             $this->addFilterByFK($script, $fk);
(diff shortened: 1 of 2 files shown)


gharlan:
I think the failures on travis are not related to this pr, but to an update of phpunit-mock-objects.

marcj:
Mh strange, we use composer.lock, do you know why it is using a new phpunit-mock-objects version?

gharlan:
> we use composer.lock

I don't think so. There is no composer.lock in this repo.

marcj:
Oh, oops :smile: Well, then we should start doing so :+1: Do you know how to fix that issue with phpunit-mock-objects?

gharlan:
> Do you know how to fix that issue with phpunit-mock-objects?

#1149

marcj:
Perfect, travis's green, thanks!

PR #1152, opened 2016-03-28
native ENUM column

jeremydenoun:
Following [PR 786](https://github.com/propelorm/Propel2/pull/786), I propose another native Enum implementation based on @oujesky work but limited to database structure (and not internal structure)

### DIFF
--- a/resources/dtd/database.dtd
+++ b/resources/dtd/database.dtd
@@ -92,7 +92,7 @@ PHP class or method name.
         | DATE   | TIME    | TIMESTAMP   | BINARY     | VARBINARY | LONGVARBINARY
         | NULL   | OTHER   | PHP_OBJECT  | DISTINCT   | STRUCT    | ARRAY
         | BLOB   | CLOB    | REF         | BOOLEANINT | BOOLEANCHAR
-        | DOUBLE | BOOLEAN | OBJECT      | ENUM       | SET
+        | DOUBLE | BOOLEAN | OBJECT      | ENUM       | NENUM     | SET
     ) "VARCHAR"
   phpType CDATA #IMPLIED
   sqlType CDATA #IMPLIED
--- a/resources/xsd/database.xsd
+++ b/resources/xsd/database.xsd
@@ -51,6 +51,7 @@
             <xs:enumeration value="OBJECT"/>
             <xs:enumeration value="ARRAY"/>
             <xs:enumeration value="ENUM"/>
+            <xs:enumeration value="NENUM"/>
             <xs:enumeration value="SET"/>
         </xs:restriction>
     </xs:simpleType>
--- a/src/Propel/Generator/Behavior/Sortable/SortableBehaviorObjectBuilderModifier.php
+++ b/src/Propel/Generator/Behavior/Sortable/SortableBehaviorObjectBuilderModifier.php
@@ -318,6 +318,11 @@ public function getScopeValue(\$returnNulls = true)
             $script .= "
     return array_search(\$this->{$this->getColumnGetter('scope_column')}(), {$this->tableMapClassName}::getValueSet({$this->tableMapClassName}::COL_{$columnConstant}));
             ";
+        } else if ($this->behavior->getColumnForParameter('scope_column')->isNativeEnumType()){
+            $columnConstant = strtoupper(preg_replace('/[^a-zA-Z0-9_\x7f-\xff]/', '_', $this->getColumnAttribute('scope_column')));
+            $script .= "
+    return \$this->{$this->getColumnGetter('scope_column')}();
+            ";
         } else if ($this->behavior->getColumnForParameter('scope_column')->isSetType()){
             $columnConstant = strtoupper(preg_replace('/[^a-zA-Z0-9_\x7f-\xff]/', '_', $this->getColumnAttribute('scope_column')));
             $script .= "
--- a/src/Propel/Generator/Builder/Om/AbstractObjectBuilder.php
+++ b/src/Propel/Generator/Builder/Om/AbstractObjectBuilder.php
@@ -48,7 +48,7 @@ abstract class AbstractObjectBuilder extends AbstractOMBuilder
                 if ($col->isNamePlural()) {
                     $this->addHasArrayElement($script, $col);
                 }
-            } elseif ($col->isEnumType()) {
+            } elseif ($col->isEnumType() || $col->isNativeEnumType()) {
                 $this->addEnumAccessor($script, $col);
             } elseif ($col->isSetType()) {
                 $this->addSetAccessor($script, $col);
@@ -93,7 +93,7 @@ abstract class AbstractObjectBuilder extends AbstractOMBuilder
                     $this->addAddArrayElement($script, $col);
                     $this->addRemoveArrayElement($script, $col);
                 }
-            } elseif ($col->isEnumType()) {
+            } elseif ($col->isEnumType() || $col->isNativeEnumType()) {
                 $this->addEnumMutator($script, $col);
             } elseif ($col->isSetType()) {
                 $this->addSetMutator($script, $col);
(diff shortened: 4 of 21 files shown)


jeremydenoun:
I'd like implement more test (based on GeneratedQueryEnumColumnTypeTest.php by example)
following @marcj comment but I don't find a way to use QuickBuilder for a specific platform...

I try to use with 
```
// (limit testing to mysql with @group mysql in class comment) +

            $platform = new MysqlPlatform();
            $configProp['propel']['database']['adapters']['mysql']['tableType'] = 'InnoDB';
            $config = new GeneratorConfig(__DIR__ . '/../../../../../Fixtures/bookstore', $configProp);
            $connection = $config->getBuildConnection();
            $platform->setGeneratorConfig($config);

            $builder = new QuickBuilder();
            $builder->setConfig($config);
            $builder->setPlatform($platform);
            $builder->buildSchema($schema, $connection['dsn'], isset($connection['user']) && $connection['user'] ? $connection['user'] : null,
            isset($connection['password']) && $connection['password'] ? $connection['password'] : null, new MysqlAdapter());
```

but it's seems QuickBuilder can work only with Sqlite (generated SQL use Sqlite platform syntax) 
or I don't find how use it... If you can help me, I can add similar test for Native implementation

Thx

marcj:
If you don't pass a config to the QuickBuilder, the QuickBuilder is automatically using SQLite `inmemory` :)

jeremydenoun:
yes I see that, but can I use "mysql" platform context, in my example the test fail because generated SQL (executed by mysql but with error) contain Sqlite syntax (by example autoincrement and not auto_increment)

marcj:
QuickBuilder should only be used for SQLite, or other platforms which aren't connected to real database (for unit tests only) If you need a connected platform the current test suite is running against, you should create new "real" tables in https://github.com/propelorm/Propel2/tree/master/tests/Fixtures/bookstore for example. After [rebuilding the fixtures](http://propelorm.org/documentation/cookbook/working-with-test-suite.html#rebuild-fixtures) you can use these tables/models then.

jeremydenoun:
thanks for your tips and if I understand how this work : 
- fixtures schema generate class connected to "real" tables" (for each platform tested) and we can use it on our own test

however it's seems fixture is platform independant, for my case I'd like test real platform only on platform where "ENUM" type is supported (sqlite not) should I create specific fixture with specific vendor ? or you recommend another way for do it ?

marcj:
I would go with the fixtures and exclude sqlite and skip the test class when you see that current platform is not supported like here https://github.com/propelorm/Propel2/blob/27bb4ec5dbe608ed9e8ff3f2f99c2ef43f03cc06/tests/Propel/Tests/Runtime/Util/TableMapTest.php#L282

jeremydenoun:
Ok perfect I work on it

Thanks

marcj:
Cool :) Feel free to ask further questions :+1:

jeremydenoun:
@marcj Thanks for tips, I add more test and I detect/fix missing code

I have a last issue where I need a little help

I patch correctly objectBuilder and queryBuilder for return value based on the real type enum platform (accessor, mutator, hydrate, ...) and my tests check common case (empty/default value/filter/count/...)

But I have an issue for ActiveQuery and `where("X = ?")` syntax on ModelCriteria.php:1797.
We make type checking for compute "values" used in SQL query, and for PropelTypes::ENUM type we use ColumnMap/getValueSetKey for array_match from value set and return array index (correct if ENUM is integer) 

however, to apply the same method as I implement into builders I'd like properly link the ColumnMap Object to a Column Object (where I can find the domain and sqlType)... It's seems voluntary to be lightweight and platform independent (following comment) but in my case without real sqlType I can't choose between array index or value for construct query.

Do you have a tip or a recommended approach ?

jeremydenoun:
PS: I remove and squash useless commit when all will be ready

marcj:
Since we can not rely on the sqlType (because maybe I want `INT` instead of `TINYINT` because I have so many sets) I would go with new type for the old behavior (that is using integer instead of real enums), so we have two: `ENUM` (which is real enum and generate automatically sqlType, so you dont have to specify it - your pull request + more) and `LENUM` for legacy ENUM, which uses the old approach with integers.

What do you think about that? This should make stuff easier to reason about. What needs to be done for this:

1. Change in all platforms: https://github.com/propelorm/Propel2/blob/master/src/Propel/Generator/Platform/MysqlPlatform.php#L54 and add LENUM
2. Change https://github.com/propelorm/Propel2/blob/master/src/Propel/Generator/Platform/DefaultPlatform.php#L367 so it returns the correct CREATE TABLE/ALTER TABLE stuff for Enums (with the stuff of `column::valueSet` attribute) 
3. change the check of `$col->getDomain()->getSqlType()) == "TINYINT" ?` to something more understandable and robust: `$col->isEnum()` and `$col->isLegacyEnum()`, where both need to be implemented like `function isEnum(){ return 'enum' === strtolower($this->getType());}`

jeremydenoun:
perfect I like your approach

jeremydenoun:
I implement your proposal and for the moment I have good result, I think I need more work/tests on real enum implementation especially on migration case (if user define enum empty the result is unexpected)

I think if native ENUM become "standard" ENUM, we should inform users about upgrade can impact unexpected behavior (if user query/update directly with PDO by example)

jeremydenoun:
I work on migration process and I find a point to cover : 
- actually ENUM are stored for mysql into TINYINT and start from 0
- if we alter type into ENUM we have an offset in value because it's seems value start from 1 in ENUM
so for fix it we must preUp migration increase by 1 value and postDown migration decrease by 1

There is an easy way to implement it ?

I work on Postgres Native Enum too, this is more harder than appear because we must create (and maintain type) before table creation... I continue to read source code to find a way to implement it

jeremydenoun:
@marcj I think our approach isn't good about ENUM and can break existing schema/database regarding migration step (with data corruption risk), I think we should reverse architecture and implement NENUM like Native Enum and keep legacy ENUM remains easier to abstract

marcj:
@jeremydenoun yes, you're right. Let us go with `NEMUM` then :+1:

jeremydenoun:
Done : 
- Add Types (PropelTypes, resources/schema)
- Default Abstract Implementation (abstract object/query builder, column case)
- Mysql implementation
- Tests

Todo : 
- Understand why CI integration fail (on my env no issue)
- Implement native ENUM types creation/migration on Oracle / Pgsql (this is a little tricky)
- Documentation about this new type - especially migration safe needs for convert from ENUM to NENUM : 
```
public function preUp($manager)
    {
        $sql = "UPDATE `table` set col = col+1;";
        $pdo = $manager->getAdapterConnection('default');
        $stmt = $pdo->prepare($sql);
        $stmt->execute();
        // add the pre-migration code here
     }

public function postDown($manager)
    {
        $sql = "UPDATE `table` set col = col-1;";
        $pdo = $manager->getAdapterConnection('default');
        $stmt = $pdo->prepare($sql);
        $stmt->execute();
        // add the post-migration code here
     }
```
_mysql native is defined like an array than contain : [ NULL, "", val1, val2 ] (increase offset)_

- Reverse from real database detection

marcj:
Awesome @jeremydenoun! Can't wait to merge this beauty whens its done :)

jeremydenoun:
I progress on Postgres & Oracle native enum implementation, it's seems propel is not really ready for abstract dynamic custom types (need additional step/query before table creation, inject dynamic type in PropelTypes, add capability to check validity/platform support/... in migration step and be compatible with existing code)

I think this can be a larger work that can include in the perimeter "specific sql type" implementation like JSON / geometric / xml... depends platform and SGDB version 

I propose to fallback to text field for the moment and improve this after more discussion in another PR

I prepare into another commit, documentation about my actual implementation

marcj:
I'm closing this since we implemented enums in our data-mapper branch, which will be merged into master soon. Thanks anyways!

jeremydenoun:
I will follow with great interest the data-mapper branch evolution and I hope breaking change will be minor.

I've always wanted finish the job and PR but I have find new issue about ENUM like value set with mixes string/integer value (with possible TableMap Collision Resolution) or I don't find automatic intelligent migration between INT -> ENUM.

For my side I use in my project `type="ENUM" valueSet="A,B,C,D" sqlType="enum('A','B','C','D')"` for be compatible with master branch and defined exactly what I want in my code and my database

Thanks for all

PR #1153, opened 2016-03-29
slug behavior and root user tests suite fix

jeremydenoun:
- slug behavior should force utf8 locale for assure iconv tranlit behavior (bootstrap.php not apply here)
- if user is root file "unreadable" checks should be skipped

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -2,6 +2,12 @@
 
 require_once __DIR__.'/../autoload.php.dist';
 
+// check if user is root and 
+if (function_exists("posix_getuid")) {
+    if (posix_getuid() == 0)
+        die("You must run tests suite with an unprivileged user.");
+}
+
 echo sprintf("Tests started in temp %s.\n", sys_get_temp_dir());
 /**
  * fix var_export behavior with floating number precision since PHP 5.4.22
@@ -11,4 +17,4 @@ echo sprintf("Tests started in temp %s.\n", sys_get_temp_dir());
  */
 ini_set('precision', 14);
 ini_set('serialize_precision', 14);
-setlocale(LC_ALL, 'en_GB.utf8'); //fixed issues with hhvm and iconv
+setlocale(LC_ALL, 'en_US.utf8'); //fixed issues with hhvm and iconv


tests/Propel/Tests/Generator/Behavior/Sluggable/SluggableBehaviorTest.php
```diff
@@ -111,6 +111,8 @@ public static function cleanupSlugProvider()
     public function testObjectCleanupSlugPart($in, $out)
     {
         $t = new TestableTable13();
+        // locale must be utf8 for be sure test work (iconv)
+        setlocale(LC_CTYPE, 'en_GB.utf8');
```
staabm: this line changes global state... we either need to reset the state after this test or we need to define this locale in a more global place so it works for all tests (and we dont have global state sideeffects after this particular test ran)

staabm:
> if user is root file "unreadable" checks should be skipped

why?

jeremydenoun:
for the first point (locale), I agree .. I push fix

for the second point (perms) :
```
#id
uid=0(root) gid=0(root) groups=0(root)
touch /tmp/test
chmod 200 /tmp/test
cat /tmp/test
```
(equivalent to `$this->getFilesystem()->chmod(sys_get_temp_dir() . '/notreadable.{ext}', 0200);`) and test wait error never happened for root user)

tests/Propel/Tests/Generator/Behavior/Sluggable/SluggableBehaviorTest.php
jeremydenoun: done

tests/Propel/Tests/Generator/Behavior/Sluggable/SluggableBehaviorTest.php
```diff
@@ -111,7 +111,12 @@ public static function cleanupSlugProvider()
     public function testObjectCleanupSlugPart($in, $out)
     {
         $t = new TestableTable13();
+        // locale must be utf8 for be sure test work (iconv)
+        $locale = setlocale(LC_CTYPE, null);
+        setlocale(LC_CTYPE, 'en_US.utf8');
         $this->assertEquals($out, $t->cleanupSlugPart($in), 'cleanupSlugPart() cleans up the slug part');
+        // restore to old
+        setlocale(LC_CTYPE, $locale);
```
marcj: This needs to be place in the tearDown method as assertEquals can fail and will stop the method.

marcj:
I really wouldn't have those root checks, as you really really shouldnt run test under root. (you should barely use root anyway, so why then for propel unit tests?) I see however the need for utf8 for slugable

tests/Propel/Tests/Generator/Behavior/Sluggable/SluggableBehaviorTest.php
jeremydenoun: ok I move this

staabm:
> I really wouldn't have those root checks, as you really really shouldnt run any test suite under root.

maybe we could instead check in https://github.com/propelorm/Propel2/blob/master/tests/bootstrap.php if the suite was started using root and error hard with a aproriate message.

staabm:
@jeremydenoun what the reasons that the `setlocale` call in the bootstrap.php doesnt work for this particular unit test?

jeremydenoun:
Regarding "root" checks, I use docker with OOB alpine for run unit testing (with standard user and root user for be complete) and only the "notreadable" checks fails with root user.

I agree with @staabm two options : 
- block root testing with global message
- support it on check

I can add "error message" if you prefer

marcj:
Yes let us throw a fatal when we see someone wants to use the test suite as root :+1:

jeremydenoun:
> @jeremydenoun what the reasons that the setlocale call in the bootstrap.php doesnt work for this particular unit test?

I check again we use bootstrap.php for this unit test but the problem is linked to en_GB.utf8 locale... if I switch to en_US.utf8 on bootstrap.php this work like expected (on debian jessie 8.0 / php7)

marcj:
Then let's see if `en_US.utf8` is also working on travis and circleci :)

jeremydenoun:
ok so this seems work with travis and circleci (fail on mariaDB starting)... I squash commit

marcj:
Perfect, thanks!

PR #1156, opened 2016-04-01
Urldecode values when parsing a DSN

### LINKED ISSUE #1141: propel init fails when password or username has special characters like #

benbankes:
Fixes #1141


marcj:
Thanks :)

PR #1157, opened 2016-04-01
Speedup travis builds

staabm:
Removing some defaults.. Try to make the travis composer cache work and no longer trigger api limits

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ env:
   - DB=sqlite
 
 install:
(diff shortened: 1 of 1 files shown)


staabm:
I think we could also drop php5.5..  It will be eol in a few months and I guess propel2 will not be released before...

http://php.net/supported-versions.php

marcj:
When it ends and we have no release, we can drop it then :+1:

staabm:
Ok, ready to land then

marcj:
Perfect, let us wait until travis is green again and mariadb is fixed, then I'm going to merge this :+1:

Issue #1144, opened 2016-03-14
mysql 5.7 does not supports adding and dropping a foreign key in the same ALTER TABLE so migrations are broken


stas-dovgodko:
For example https://dev.mysql.com/doc/refman/5.6/en/create-table-foreign-keys.html

> Prior to MySQL 5.6.6, adding and dropping a foreign key in the same ALTER TABLE statement may be problematic in some cases and is therefore unsupported. Separate statements should be used for each operation. As of MySQL 5.6.6, adding and dropping a foreign key in the same ALTER TABLE statement is supported for ALTER TABLE ... ALGORITHM=INPLACE but remains unsupported for ALTER TABLE ... ALGORITHM=COPY.

So for my table this migrtion:
`ALTER TABLE `asset`          
    DROP FOREIGN KEY `fk_assetkoatuu`,                
    DROP INDEX `asset_i_a79bf5`,                      
    ADD SPATIAL INDEX `asset_i_a79bf5` (`location`),  
    ADD CONSTRAINT `fk_assetkoatuu`                   
      FOREIGN KEY (`koatuu_id`)                       
      REFERENCES `koatuu` (`id`)                      
      ON DELETE RESTRICT`

working with 5.5 and not with 5.7.

PR #1161, opened 2016-04-06
Revert "Make MySQL table modifications always use one statement to fix total table-corruption bug #1115"

marcj:
Reverts propelorm/Propel2#1116

Because of https://github.com/propelorm/Propel2/issues/1144 and I discovered some strange effects with migrations after this PR as well, so revert for the moment (`Cannot drop index 'PRIMARY': needed in a foreign key constraint`).


gharlan:
:+1:

ExplodingCabbage:
Ah, what a pity. Just discovered this revert!

Not sure it's the right call; the bug https://github.com/propelorm/Propel2/pull/1116 fixes corrupts your database, whereas the bug this revert fixes just causes an illegal statement to be generated that will error out when you try to execute it. Still, it would be nice to figure out how to fix both simultaneously.

marcj:
Yes, in my daily use cases where I change a lot of tables during development I had several issues with this PR, so I reverted it. Maybe we get it working without `FOREIGN_KEY_CHECKS=0` so we get always a error messages instead of a corrupt database, however I'm not sure if this is even possible.

gharlan:
> Yes, in my daily use cases where I change a lot of tables during development I had several issues with this PR

Same here

> Still, it would be nice to figure out how to fix both simultaneously.

👍

Issue #943, opened 2015-06-01
Propel is trying to create and persist a new object instead of updating it


said-jomedia:
I retrieve a record from the database this way 

``` php
$error_messages = "";
 $filter = FiltersQuery::create()->findPk(1);
if ($filter->validate()) {
            $filter->save();
            echo "SUCCESS";
        } else {
            foreach ($filter->getValidationFailures() as $failure) {
                $error_messages .= $failure->getInvalidValue() . " " . $failure->getPropertyPath() . ": " . $failure->getMessage();
            }
            echo $error_messages;
        }
```

I get validation errors, like _the filter name already exists_ because it's thinking that I want to create a new object and persist it instead of updating an existing one.

marcj:
I rather think the validation is buggy and does not exclude itself from the unique constraint check. Please show us your concrete validations and the full stacktrace of the exception.

said-jomedia:
You're totally right, it was the validation that is causing the probleme of saving on update. But I don't understand why, the validation was working before and is also working for the create (save new record). I just disabled the validation and the update is working again. Here's my validation code in the schema: 
````XML
  <parameter name="nameUnique"
                       value="{column: name, validator: Unique, options: {message: This filter name already exists in the database}}"/>
````
So I don't know why the validation is buggy.

Here's the validation method generated by propel in my model:
````PHP
  public function validate(ValidatorInterface $validator = null)
    {
        if (null === $validator) {
            if(class_exists('Symfony\\Component\\Validator\\Validator\\LegacyValidator')){
                $validator = new LegacyValidator(
                            new ExecutionContextFactory(new DefaultTranslator()),
                            new ClassMetaDataFactory(new StaticMethodLoader()),
                            new ConstraintValidatorFactory()
                );
            }else{
                $validator = new Validator(
                            new ClassMetadataFactory(new StaticMethodLoader()),
                            new ConstraintValidatorFactory(),
                            new DefaultTranslator()
                );
            }
        }

        $failureMap = new ConstraintViolationList();

        if (!$this->alreadyInValidation) {
            $this->alreadyInValidation = true;
            $retval = null;


            $retval = $validator->validate($this);
            if (count($retval) > 0) {
                $failureMap->addAll($retval);
            }

            if (null !== $this->collDisplaySegments) {
                foreach ($this->collDisplaySegments as $referrerFK) {
                    if (method_exists($referrerFK, 'validate')) {
                        if (!$referrerFK->validate($validator)) {
[...]

PR #1169, opened 2016-04-16
Update UniqueValidator.php

fenric:
See https://github.com/propelorm/Propel2/issues/943

Sory I do not speak English :(


marcj:
Could you please try to explain what you're trying to fix?

alexdevid:
Why this is not merged?
The problem is when you try to **update** some model which has Unique validation constraint you got Error `This value is already stored in your database`
That fix solves this problem

marcj:
@alexdevid because of now answer and no tests. Feel free to add tests to it, so we can merge it

Issue #824, opened 2014-11-28
Versionable behavior duplicates tablePrefix in table name


gossi:
My schema in short:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<database defaultIdMethod="native" name="keeko"	namespace="gossi\trixionary\model" identifierQuoting="true"
	defaultPhpNamingMethod="phpname" xmlns:xsi="http://www.w3.org/2001/XMLSchema-inposition"
	xsi:noNamespaceSchemaLocation="http://xsd.propelorm.org/1.6/database.xsd"
	tablePrefix="kk_trixionary_">

	<table name="skill">
		<column name="id" type="INTEGER" primaryKey="true" required="true"
			autoIncrement="true" />
		...
		<behavior name="versionable">
			<parameter name="log_created_at" value="true"/>
			<parameter name="log_created_by" value="true"/>
			<parameter name="log_comment" value="true"/>
		</behavior>
	</table>
</database>
```

Results in the following table name for the versioned skill table: `kk_trixionary_kk_trixionary_version`.

Remark: There is no tablePrefix in `propel.yaml`.

PR #1171, opened 2016-04-21
Fix versionable behavior

rafaelnery:
As mentioned in issue #824  versionble behavior was generating the version table with duplicate prefix.

Table was changed the call method Table::getName() and replaced by Table::getOriginCommonName()

And finally been implemented tests for database with tablePrefix.


marcj:
Cool, thanks!

PR #1172, opened 2016-04-24
Allow PROPEL_ATTR_CACHE_PREPARES to be set via config

kitsunde:
It's documented as being settable in the config: http://propelorm.org/documentation/reference/configuration-file.html#attributes but it fails validation.

### DIFF
--- a/src/Propel/Common/Config/PropelConfiguration.php
+++ b/src/Propel/Common/Config/PropelConfiguration.php
@@ -124,6 +124,8 @@ class PropelConfiguration implements ConfigurationInterface
                                     ->end()
                                     ->arrayNode('attributes')
                                         ->children()
+                                            ->booleanNode('PROPEL_ATTR_CACHE_PREPARES')->defaultFalse()->end()
+                                            ->booleanNode('PropelPDO::PROPEL_ATTR_CACHE_PREPARES')->defaultFalse()->end()
                                             ->booleanNode('ATTR_EMULATE_PREPARES')->defaultFalse()->end()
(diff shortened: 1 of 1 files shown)


kitsunde:
Odd, this causes a regression on some issue I don't understand. https://github.com/propelorm/Propel2/blob/master/tests/Propel/Tests/Issues/Issue617Test.php

src/Propel/Common/Config/PropelConfiguration.php
```diff
@@ -124,6 +124,8 @@ protected function addDatabaseSection(ArrayNodeDefinition $node)
                                     ->end()
                                     ->arrayNode('attributes')
                                         ->children()
+                                            ->booleanNode('PROPEL_ATTR_CACHE_PREPARES')->defaultFalse()->end()
+                                            ->booleanNode('PropelPDO::PROPEL_ATTR_CACHE_PREPARES')->defaultFalse()->end()
```
marcj: is really both necessary?

marcj:
Well its seems this breaks a lot of things https://travis-ci.org/propelorm/Propel2/jobs/125438698
screen shot 2016-04-25 at 00 02 44

:D

src/Propel/Common/Config/PropelConfiguration.php
kitsunde: According to the documentation you can set it either with `PropelPDO::` or without.

marcj: oh, then we should allow that.

marcj: Maybe you should omit `defaultFalse` to get a green travis

PR #1174, opened 2016-04-30
add code enhancement on class NestedSetBehaviorObjectBuilderModifier

dmelo:
Instead of calculating `$objectClassName`, `$queryClassName` and `$useScope` on every method that uses those variables, it calcultes it at the __construct and setBuilder and use the stored values on various locations.

It still passing all tests on my machine and code GPA is slightly improved. https://codeclimate.com/github/dmelo/Propel2/compare/code-enhancement#ratings

### DIFF
--- a/src/Propel/Generator/Behavior/NestedSet/NestedSetBehaviorObjectBuilderModifier.php
+++ b/src/Propel/Generator/Behavior/NestedSet/NestedSetBehaviorObjectBuilderModifier.php
@@ -30,11 +30,14 @@ class NestedSetBehaviorObjectBuilderModifier
     /** @var ObjectBuilder */
     protected $builder;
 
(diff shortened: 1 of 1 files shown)


marcj:
I'm really not in favour of this change due to the fact that is only a cosmetic change with a huge diff. Refactoring all those huge builder is PITA and hard to track when you hunt down some changes that broke something. We're working already on the data-mapper branch to clean this mess up.

dmelo:
I understand the concern about big changes. But bear in mind that those changes make the code 51 lines lighter. It does it by taking away a block of code that was repeated on multiple places. Consider the benefits of DRY. Also, it makes fewer calls to "get.*" methods, by using class attributes instead. So the code tends to be more efficient as well.

PR #1186, opened 2016-05-18
Add phpdoc for "MigrationManager" parameter

LeoOnTheEarth:
It is easier to add a phpdoc to tell the `$manager` parameter's class name


marcj:
Why not typehint it directly in PHP? :)

LeoOnTheEarth:
> Why not typehint it directly in PHP? :)

That would be nice too :)

I will use `Propel\Generator\Manager\MigrationManager` for typehint

marcj:
Great thanks :)

PR #1202, opened 2016-05-30
Add ability to use specified namespace in propel init command

### LINKED ISSUE #1201: Namespace requested by the init command does not appear as in reverse engineered schema.xml
`propel init` asks "Which namespace should the generated php models use?", but the resulting models do not have the namespace applied because the namespace is never placed in the schema.xml file under the condition that the other question "Do you have an existing database you want to use with propel?" is answered affirmatively.

benbankes:
Fixes #1201 by placing the namespace (given by the user) into schema.xml when reverse engineering an existing database.  Also replaces the unused `schemaAutoNamespace` configuration option with the currently used `generator.schema.autoNamespace` propel configuration option.


mpscholten:
Good job 👍

marcj:
Nicely done, thanks!

PR #1204, opened 2016-05-31
PDODataFetcher can have null data object and throw exception

jeremydenoun:
PDODataFetcher can have null data object (after close by example) and rewinded (dump by example) 
=> tests avoid exception

### DIFF
--- a/src/Propel/Runtime/DataFetcher/PDODataFetcher.php
+++ b/src/Propel/Runtime/DataFetcher/PDODataFetcher.php
@@ -108,7 +108,8 @@ class PDODataFetcher extends AbstractDataFetcher
      */
     public function rewind()
     {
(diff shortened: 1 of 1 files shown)


marcj:
Can you point out under which circumstances this occurs?

jeremydenoun:
I use Kint (http://raveren.github.io/kint/) for debugging, Kint dumping method use object detection for transtyping correctly to array and detect PropelObject like Iterable (and use rewind method for try fetch more informations)[src](https://github.com/raveren/kint/blob/master/parsers/custom/objectiterateable.php) this should work but rewind after close throw exception

jeremydenoun:
this is linked to [iterator_to_array](http://php.net/manual/fr/function.iterator-to-array.php) function for be more global

marcj:
Alrighty, thanks!

PR #1207, opened 2016-06-03
Polymorphic relations allowed by DTD/XSD

prgTW:
Polymorphic relations are supported a long time now (https://github.com/propelorm/Propel2/pull/864) but DTD/XSD aren't aware of that


marcj:
Thanks!

prgTW:
I made mistake, please look at #1208

PR #1208, opened 2016-06-04
value part of reference should be optional

prgTW:
quick fix for #1207, sorry, my mistake :(


marcj:
No problem, thanks :)

PR #1215, opened 2016-06-10
Added proper Exception when encountering a table without PK.

JustusW:
During model:build time this error may be encountered when using imported tables without PK.


marcj:
Thanks!

PR #1217, opened 2016-06-10
Inconsistent behavior between join and joinWith leading to malformed on clause

JustusW:
$realFullColumnName may be either table.field or schema.table.field so the correct $realColumnName is always the last field of the exploded array and not the second.


JustusW:
Failing checks seem to be related to 5d5825b61d2432dbb03467d08db71fb116bd8c63 if I see this correctly?

PR #1220, opened 2016-06-13
Added some missing types for PGSQL

JustusW:
Not 100% sure about the character varying mapping to varchar although it seems appropriate.


marcj:
Looks good to me, thanks!

Issue #774, opened 2014-10-09
Versionable between tables


drodil:
Here is one fiddle that doesn't work. Versionable behaviour in two tables breaks Propel:

http://sandbox.propelorm.org/42d5d28

PR #1221, opened 2016-06-14
Fix Versionable behavior with FK and Custom Version Column

rafaelnery:
Fix versionable behavior when exists 2 tables with versionable behavior and custom version_column parameter.

Changed filterByVersion to filterBy[PHPNameOfField]

Issue #774


rafaelnery:
I will squash my commits :smiley:.

Sorry.

PR #1222, opened 2016-06-14
Fixed versionable behavior

rafaelnery:
Issue #774 

Fix versionable behavior when exists 2 connected tables, versionable behavior in both tables and parameter version_column is changed.

---
* Fix versionable when exists custom field on table with FK
* Implementing tests

---


marcj:
Perfect, thanks!

Issue #642, opened 2014-05-22
DatabaseMap::getTableByPhpName is really slow


drodil:
Hi,

I am using Propel for my project and I did some cache grinding with Xdebug. It seems that the DatabaseMap::getTableByPhpName function is eating almost 50% of the time of the request (which is around 800ms). 

I have custom class loader which seems to be called also during this function. Is there anyway this could be avoided or make this specific function faster?

PR #1224, opened 2016-06-15
Performance improvement for getTableByPhpName

drodil:
First try to look from \Map\ folder instead in model base folder which is most
of the time the correct case. Optimized composer autoloader now finds the
classes correctly from class map. Later on this should be configurable most
likely.

This relates to issue #642


staabm:
Did you try to use the new zend-opcache optimized composer autoloader? Maybe you can use the same technique to speedup your very own autoloader?

drodil:
@staabm I have only tried composer with --optimized flag enabled - it still misses the tablemap loadings in the first class_exists function here. What is zend-opcache optimized composer autoloader? I have opcache enabled but is there something to configure in composer to get it work? Maybe I will consider own autoloader for the tablemap loading..

staabm:
See https://github.com/composer/composer/pull/5174

I guess its only available in composer 1.1.x

PR #1228, opened 2016-06-22
Added tests that shows failures when trying queries that seem correct but apparently are not working

motin:
After some fiddling with queries I realized that in order for the query to actually work, I had to:
1. Supply relation aliases
2. Make sure that the relation aliases where in lowercase and started with the table name
3. Override the default join type in the use...Query() method

These tests shows the current behavior when not doing all of the above, as well as includes the working query for reference.


gechetspr:

PR #1229, opened 2016-06-23
Fix UniqueValidator error on update

alexdevid:
The problem is when you try to **update** some model which has Unique validation constraint you got `Error This value is already stored in your database`


alexdevid:
not actual

PR #1235, opened 2016-07-04
Fix doubly-namespaced classes on database:reverse command

### LINKED ISSUE #1233: Invoking database:reverse incorrectly duplicates namespace in schema.xml table element
The command `propel database:reverse` places the value from the `--namespace` option in the namespace attribute of both the `database` and `table` xml elements, causing the table's namespace.  When a namespace of `Foo` is used for the `--namespace` option of the command, the following xml results:

```
<database name="my_db" namespace="Foo">
  <table name="user" namespace="Foo">
  </table>
</database>
```

Per the namespace inheritance rules at http://propelorm.org/documentation/cookbook/namespaces.html, this would create a class `User` with the namespace `Foo\Foo`, which is not intended.

benbankes:
Fixes #1233, Replaces #1234

### DIFF
--- a/src/Propel/Generator/Schema/Dumper/XmlDumper.php
+++ b/src/Propel/Generator/Schema/Dumper/XmlDumper.php
@@ -214,8 +214,19 @@ class XmlDumper implements DumperInterface
             $tableNode->setAttribute('package', $package);
         }
 
-        if ($namespace = $table->getNamespace()) {
-            $tableNode->setAttribute('namespace', $namespace);
+        if ($table->getNamespace()) {
+            if ($table->getDatabase()->getNamespace()) {
+                $startsWithDbNamespace = '^' . preg_quote($table->getDatabase()->getNamespace());
+                $optionalBackslash = preg_quote('\\') . '?';
+
+                $regex = '/' . $startsWithDbNamespace . $optionalBackslash . '/';
+
+                $namespace = preg_replace($regex, '', $table->getNamespace());
+            }
+
+            if ($namespace) {
+                $tableNode->setAttribute('namespace', $namespace);
+            }
         }
 
         if ($table->isSkipSql()) {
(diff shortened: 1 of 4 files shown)


marcj:
Thanks for the contribution! I read through the https://github.com/propelorm/Propel2/commit/12cc008196a2f49aee25281bd3562151cb6ae008 commit which brought in the change. I guess the fix is not to fix XmlDumper but the actual internal model representation. Meaning, that the Reverse class/ReverseManager should not place the `--namespace` in Table::namespace. However, I'm currently not sure why Table::namespace is set, since [it actually only sets the Database::namespace](https://github.com/propelorm/Propel2/commit/12cc008196a2f49aee25281bd3562151cb6ae008#diff-fa0b081df8c12d75df36c1299325689bR217).

Incognito:
Rebasable and passes on recent CI. 

Conflicts with design goals of https://github.com/propelorm/Propel2/pull/1241

src/Propel/Generator/Schema/Dumper/XmlDumper.php
```diff
@@ -214,8 +214,19 @@ private function appendTableNode(Table $table, \DOMNode $parentNode)
...
+            if ($table->getDatabase()->getNamespace()) {
+                $startsWithDbNamespace = '^' . preg_quote($table->getDatabase()->getNamespace());
+                $optionalBackslash = preg_quote('\\') . '?';
+
+                $regex = '/' . $startsWithDbNamespace . $optionalBackslash . '/';
+
+                $namespace = preg_replace($regex, '', $table->getNamespace());
+            }
+
+            if ($namespace) {
```
Incognito, line 227: Validation error.

```

 ------ ---------------------------------------------- 
  Line   Propel/Generator/Schema/Dumper/XmlDumper.php  
 ------ ---------------------------------------------- 
  227    Variable $namespace might not be defined.     
 ------ ---------------------------------------------- 
```

dereuromark:
Is there any reason to have the `.../Model/` prefix?

Also:

    package="Acme.Blog" namespace="Blog" 

Why do we need the namespace at all, if the package seems to contain it. Or can those be separate?

dereuromark:
Replacement https://github.com/propelorm/Propel2/pull/1729

PR #1236, opened 2016-07-04
Run sql:build model:build config:convert post-init

benbankes:
Runs the next commands required to be able to use Propel

### DIFF
--- a/src/Propel/Generator/Command/InitCommand.php
+++ b/src/Propel/Generator/Command/InitCommand.php
@@ -37,7 +37,6 @@ class InitCommand extends AbstractCommand
         $this->defaultPhpDir = $this->detectDefaultPhpDir();
     }
 
-
     protected function configure()
     {
         parent::configure();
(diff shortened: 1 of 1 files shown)


benbankes:
@mpscholten Is there any reason we would not want to run the following commands directly after everything else in the `init` command?
`sql:build`
`model:build`
`config:convert`

Is there something lacking in this PR that needs to be written before it is mature enough to be pulled into the codebase?

src/Propel/Generator/Command/InitCommand.php
```diff
@@ -225,6 +226,27 @@ private function generateProject(OutputInterface $output, array $options)
...
         $this->writeFile($output, sprintf('%s/propel.%s.dist', getcwd(), $options['format']), $distConfig->render($options));
+
+        $this->buildSqlAndModelsAndConvertConfig();
+    }
+
+    private function buildSqlAndModelsAndConvertConfig()
+    {
+        $this->getApplication()->setAutoExit(false);
+
+        $followupCommands = [
```
mpscholten: Could we call this `followUpCommands` (I guess the english word for this is "follow up" and thus the camel case version would be "followUp" instead of "followup")

mpscholten:
Very cool idea :)

@marcj anything missing to get this merged?

marcj:
Well, no, I just waited for your feedback :)

marcj:
Thanks!

PR #1242, opened 2016-07-26
Locale-independent functionality of PropelDateTime::createHighPrecision

Musiksammler:
It came to my attention that PropelDateTime::createHighPrecision can return "false" when the numeric format for floats is changed by setting specific locales. For example in the locale "de_DE" floats will be formatted with a "," (comma) as decimal pointer. This will result in a "false" from DateTime::createFromFormat.

I submitted a patch, just replacing a "," with a "." in the value of "microtime(true)".


marcj:
Good catch, thanks!

PR #1243, opened 2016-07-27
specify timezone in PropelDateTime::createHighPrecision

niekdemelker:
just like in the newInstance
timezone must be explicitly specified and then changed
because of a DateTime bug: http://bugs.php.net/bug.php?id=43003


gharlan:
👍

mijgame:
image

gharlan:
@niekdemelker can you add a test case for this?

PR #1245, opened 2016-07-31
Fix schemaname for PostgreSQL 9.5

rafaelnery:
I discovered it was the migration, but in fact the problem is PgsqlSchemaParser. For up to version 9.4 the command  **show search_path** returns:

```SQL
# show search_path;
|    search_path |
|----------------|
| "$user",public |
```

From the postgreSQL version 9.5 we have the result.
```SQL
# show search_path;
|    search_path  |
|-----------------|
| "$user", public |
```
Just only one space and everything explodes.

So I removed the spaces at the time of reading the schemas :smiley: 

The **TravisCI** is not ready for **PostgreSQL9.5** why not add it in .travis.yml(travis-ci/travis-ci#4264). Not ready without sudo.

### DIFF
--- a/src/Propel/Generator/Reverse/PgsqlSchemaParser.php
+++ b/src/Propel/Generator/Reverse/PgsqlSchemaParser.php
@@ -229,7 +229,7 @@ class PgsqlSchemaParser extends AbstractSchemaParser
             $searchPath = explode(',', $searchPathString);
 
             foreach ($searchPath as &$path) {
-                $params[] = $path;
+                $params[] = trim($path);
                 $path = '?';
             }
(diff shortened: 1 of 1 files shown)


marcj:
Hey, good catch! Let's what our test suite says :)

> The TravisCI is not ready

`not` => `now` ? :P

rafaelnery:
He is not prepared natively ... If you want to add you should use SUDO, unlike our .travis.tml

dkpeall:
This change worked for me thanks! Hopefully gets included soon.

marcj:
Thanks!

PR #1249, opened 2016-08-09
Postgresql Error: DROP SEQUENCE needs IF EXISTS since PostgreSQL v8.2

remyw:
Since Propel 8.2:

IF EXISTS

```
 Do not throw an error if the sequence does not exist. A notice is issued in this case.
```

See: https://www.postgresql.org/docs/8.2/static/sql-dropsequence.html


marcj:
For what use case? The time a migration file is generated we know exactly whether the sequence did exist or not. I don't see a reason to use `if exists`. Can you explain your intention?

PhilinTv:
In Propel-generated migrations, the sequence presence is always known.
Time close :)

PR #1250, opened 2016-08-10
Update to use TIME & DATE types

mhaisley:
Previously these used DATETIME, but then propel tried to insert just a date or time into the database, resulting in a error "[S0001][241] Conversion failed when converting date and/or time from character string."


mhaisley:
@marcj Can you merge this as is, or do I need to do something else since the one test failed?

PR #1251, opened 2016-08-10
[data-mapper]Fix object builder tests

cristianoc72:
Fix object builder tests for entities having fields of array, boolean
and enum type.


marcj:
Yeah !

Issue #1257, opened 2016-08-28
PropelDateTime::createHighPrecision() leads to "Call to a member function setTimeZone() on boolean"


motin:
This happens after #1243 sometimes (randomly) when invoked from timestampable behavior. Different classes and item types. One of these lines are always in the backtrace:

```
$this->setCreated(\Propel\Runtime\Util\PropelDateTime::createHighPrecision());
```
```
$this->setModified(\Propel\Runtime\Util\PropelDateTime::createHighPrecision());
```

The error is thrown when`\DateTime::createFromFormat('U.u', self::getMicrotime());` returns false, which it should not be able to do, but seemingly does at random.

motin:
Found the explanation at https://bugs.php.net/bug.php?id=64414 and http://php.net/manual/en/datetime.createfromformat.php#119362

Reportedly, `(string) microtime(true)` may return a string timestamp number without a fractional part if the microseconds are exactly zero.  I.e., "1463772747" instead of the expected "1463772747.000000". 

number_format() can create a correct string representation of the microsecond timestamp every time, which can be useful for creating DateTime objects when used with DateTime::createFromFormat()

<?php
$now = DateTime::createFromFormat('U.u', number_format(microtime(true), 6, '.', ''));
var_dump($now->format('Y-m-d H:i:s.u')); // E.g., string(26) "2016-05-20 19:36:26.900794"

PR #1258, opened 2016-08-28
Ensuring that PropelDateTime::getMicrotime() includes the fractional part

motin:
... to match the expected 'U.u' format in createHighPrecision(). (#1257)


marcj:
Thanks!

Issue #675, opened 2014-07-27
Check incomplete foreign-keys


marcj:
See http://stackoverflow.com/questions/7947085/are-incomplete-key-references-in-propel-useful

When `foreign-key` points to a table with composite primary key but we have only defined one reference in our foreign key then we should throw a error.

crossan007:
It may be useful to update the documentation *somewhere* near this link: http://propelorm.org/documentation/04-relationships.html#more-than-two-foreign-keys

PR #1259, opened 2016-08-29
Check incomplete foreign-keys #675

atompulse:
Propel should not allow incomplete foreign key references when foreign
table has a composite primary key.

https://github.com/propelorm/Propel2/issues/675

### DIFF
--- a/src/Propel/Generator/Model/Table.php
+++ b/src/Propel/Generator/Model/Table.php
@@ -784,6 +784,25 @@ class Table extends ScopedMappingModel implements IdMethod
                 }
             }
 
+            // check for incomplete foreign key references when foreign table
+            // has a composite primary key
+            if ($foreignTable->hasCompositePrimaryKey()) {
+                // get composite foreign key's keys
+                $foreignPrimaryKeys = $foreignTable->getPrimaryKey();
+                // check all keys are referenced in foreign key
+                foreach ($foreignPrimaryKeys as $foreignPrimaryKey) {
+                    if (!$foreignPrimaryKey->hasReferrer($foreignKey) && $throwErrors) {
+                        // foreign primary key is not being referenced in foreign key
+                        throw new BuildException(sprintf(
+                            'Table "%s" contains a foreign key to table "%s" but does not have a reference to foreign primary key "%s"',
+                            $this->getName(),
+                            $foreignTable->getName(),
+                            $foreignPrimaryKey->getName()
+                        ));
+                    }
+                }
+            }
+
             if ($this->getPlatform() instanceof MysqlPlatform) {
                 $this->addExtraIndices();
             }
--- a/src/Propel/Runtime/Util/PropelDateTime.php
+++ b/src/Propel/Runtime/Util/PropelDateTime.php
@@ -76,8 +76,8 @@ class PropelDateTime extends \DateTime
     }
 
     /**
-     * Get the current microtime with milliseconds. Making sure that the decimal point separator is always "."
-     * and always including the fractional part. Otherwise self::createHighPrecision would return false.
+     * Get the current microtime with milliseconds. Making sure that the decimal point separator is always ".", ignoring
+     * what is set with the current locale. Otherwise self::createHighPrecision would return false.
      *
      * @return string
      */
@@ -85,7 +85,7 @@ class PropelDateTime extends \DateTime
     {
         $mtime = microtime(true);
 
-        return number_format($mtime, 6, '.', '');
+        return str_replace(',', '.', $mtime);
     }
 
     /**
--- a/tests/Propel/Tests/Issues/Issue675Test.php
+++ b/tests/Propel/Tests/Issues/Issue675Test.php
@@ -1,0 +1,60 @@
+<?php
+/**
+ * This file is part of the Propel package.
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ *
+ * @license MIT License
+ */
+namespace Propel\Tests\Issues;
+
+use Propel\Generator\Util\QuickBuilder;
+use Propel\Tests\TestCaseFixtures;
+
+/**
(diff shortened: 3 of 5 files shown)


tests/Propel/Tests/Issues/Issue675Test.php
```diff
@@ -0,0 +1,60 @@
...
+    </table>
+</database>
+EOF;
+
+        $builder = new QuickBuilder();
+        $builder->setSchema($schema);
+        $builder->buildClasses(null, true);
+
+        // Propel should disallow incomplete foreign reference
+        $this->expectException('\Propel\Generator\Exception\BuildException');
```
marcj: you have this already in the annotation "@expectedException"

src/Propel/Generator/Model/Table.php
```diff
@@ -784,6 +784,25 @@ public function setupReferrers($throwErrors = false)
...
             }
 
+            // check for incomplete foreign key references when foreign table
+            // has a composite primary key
+            if ($foreignTable->hasCompositePrimaryKey()) {
+                // get composite foreign key's keys
+                $foreignPrimaryKeys = $foreignTable->getPrimaryKey();
+                // check all keys are referenced in foreign key
+                foreach ($foreignPrimaryKeys as $foreignPrimaryKey) {
+                    if (!$foreignPrimaryKey->hasReferrer($foreignKey) && $throwErrors) {
```
marcj: What if my foreign key refers to an unique key of $foreignTable instead of its primary key? This should be supported as well.

atompulse: Do we have test case for this situation?

marcj: No, I don't think we have somewhere a schema.xml with that scenario

atompulse: Do you think we could cover this use case with this schema ?

`<database name="test" defaultIdMethod="native">
    <table name="event">
        <column name="id" type="integer" required="true" primaryKey="true" autoIncrement="true" />
        <column name="name" type="varchar" size="50" required="true" />
        <column name="organiser_id" type="integer" required="true" />
        
        <foreign-key foreignTable="organiser">
            <reference local="organiser_id" foreign="id" />
        </foreign-key>
    </table>

    <table name="organiser">
        <column name="id" type="integer" required="true" primaryKey="true" autoIncrement="true" />
        <column name="code" type="integer" required="true"/>
        <column name="name" type="varchar" size="50" required="true" />
        
        <unique name="u_id_code">
          <unique-column name="id" />
          <unique-column name="code" />
        </unique>
    </table>
</database>`

marcj: Rather with this

```xml
<table name="organiser">
        <column name="id" type="integer" required="true" primaryKey="true" autoIncrement="true" />
        <column name="code" type="integer" required="true"/>
        <column name="name" type="varchar" size="50" required="true" />
        <!-- Has unique -->
        <unique>
          <unique-column name="code" />
        </unique>
    </table>
```

atompulse: @marcj could you please elaborate the exact situation we are trying to catch here related to unique's ?

According to #675 
> See http://stackoverflow.com/questions/7947085/are-incomplete-key-references-in-propel-useful
> 
> When foreign-key points to a table with composite primary key but we have only defined one reference in our foreign key then we should throw a error.
> 
this pull request fixes exactly this...

I tried to understand/find your unique key issue but I fail to do so :(

marcj: Ok, then let's ignore this use case for the moment.

atompulse: Sounds good.

marcj:
Perfect, good job, thanks!

src/Propel/Runtime/Util/PropelDateTime.php
```diff
@@ -76,16 +76,16 @@ public static function createHighPrecision($time = null)
...
+     * what is set with the current locale. Otherwise self::createHighPrecision would return false.
      *
      * @return string
      */
     public static function getMicrotime()
     {
         $mtime = microtime(true);
 
-        return number_format($mtime, 6, '.', '');
+        return str_replace(',', '.', $mtime);
```
gharlan: @atompulse Sometimes now `PropelDateTime::createHighPrecision()` throws an error.
It happens, when the microtime has 0 micro seconds (e.g. 1496476204.0).
You convert the float value to string here (by `str_replace`). Result will be `'1496476204'`.
Then `$dateTime = \DateTime::createFromFormat('U.u', ...)` in line 71 will fail.

marcj: Fixed https://github.com/propelorm/Propel2/commit/efafd0960ada68914b2a03f97151da98226c0f75

gharlan: thanks!

src/Propel/Generator/Model/Table.php
```diff
@@ -784,6 +784,25 @@ public function setupReferrers($throwErrors = false)
                 }
             }
 
+            // check for incomplete foreign key references when foreign table
+            // has a composite primary key
+            if ($foreignTable->hasCompositePrimaryKey()) {
```
prgTW: `$foreignTable` can be `null` here. I reproduced it with a table defined in schema that had a foreign key to a table not defined in schema but still created with a behavior

atompulse: I believe that is an edge-case. If you have a solution please contribute with a merge-request.

atompulse: @prgTW I have a merge request that infact solves this.

prgTW: actually my first solution I was testing was to compare against `null` as You already did in Your PR. I didn't come up with a better solution so far

PR #1262, opened 2016-09-02
Support more complex counts

motin:
Full backwards compatibility. The LogicException with message "Propel cannot create a COUNT query when using HAVING and  duplicate column names in the SELECT part" has been removed.


PhilinTv:
Good potential for complex analytics/reporting use case.

gechetspr:

PR #1270, opened 2016-09-09
Fixed versionable behavior

gabor-kormany:
Versioning from N:1 side of the relation was not working properly due to the generator adding the foreign column instead of the primary key. This was effecting "addVersion"  and "populateFromVersion" calls. Exchanged the $fk usage with the use of "$fk->getTable()->getFirstPrimaryKey()". This means that the N side of the relation should use the primary key instead of using the foreign key.

### DIFF
--- a/src/Propel/Generator/Behavior/Versionable/VersionableBehaviorObjectBuilderModifier.php
+++ b/src/Propel/Generator/Behavior/Versionable/VersionableBehaviorObjectBuilderModifier.php
@@ -457,7 +457,7 @@ public function populateFromVersion(\$version, \$con = null, &\$loadedObjects = 
             $relatedVersionQueryClassName = $this->builder->getClassNameFromBuilder($this->builder->getNewStubQueryBuilder($foreignVersionTable));
             $relatedVersionTableMapClassName = $this->builder->getClassNameFromBuilder($this->builder->getNewTableMapBuilder($foreignVersionTable));
             $relatedClassName = $this->builder->getClassNameFromBuilder($this->builder->getNewStubObjectBuilder($foreignTable));
-            $fkColumn = $fk->getLocalColumn();
+            $fkColumn = $foreignTable->getFirstPrimaryKeyColumn();
             $fkVersionColumn = $foreignVersionTable->getColumn($this->behavior->getParameter('version_column'));
 
             $script .= "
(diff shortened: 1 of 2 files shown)


gabor-kormany:
I've faced the following problem when using versionable behavior:
For versioned objects in N:1 relation the generated class on the N side contains invalid constants in addVersion and populateFromVersion calls. 

I've fixed the issue by not referencing to the foreign key but to the foreign object's primary key. 
One restriction applies, that the primary key should be a simple one (one field).

Please review and apply the fix.
Thanks!

marcj:
I'm not sure about that change. Can you push a test case that fails without the change? (so travis keeps green with your fix, but when I undo the changes outside of the test files, it will be red)

gabor-kormany:
Hi, 

Not sure, I might need your help on it. How can I test if the generated class have invalid code? If you tell me that I'm happy to push a test case.

marcj:
A test case covers your changes. Means, if you revert your changes, this new test case should fail. Introducing your changes back in, makes the test case green. This helps other people to understand what the actual bug was and proves that your changes actually fixes that thing.

gabor-kormany:
I have the impression that we're not on the same page.
My problem is that I don't really know how to test if the output of the generator contains invalid constants. And by invalid I mean, constants that are not defined within the class or its ancestor(s), depending on how it is referenced. 
So If you can help me with that I'd happy to provide the test for you.

marcj:
> My problem is that I don't really know how to test if the output of the generator contains invalid constants 

The generator contains invalid constants? I don't see in the diff that you change any constants?

gabor-kormany:
If you fold out to line 469 you should see:
`\$c1 = \$query->getNewCriterion({$this->builder->getColumnConstant($fkColumn, $relatedVersionTableMapClassName)}, \$value);`

and one of the changes (in line 460):
`-$fkColumn = $fk->getForeignColumn();`
`+$fkColumn = $foreignTable->getFirstPrimaryKeyColumn();`

The original version was generating code with reference to a constant that is defined in the other side class of the relation. So I'm changing it to generate the proper constant. And yes I'm not explicitly changing constants in the code. I'm changing which constant is used in the generated code.

marcj:
Alright, I see it now, thanks for explanation! :) I'm going to merge it, thanks!

PR #1277, opened 2016-09-15
Add makeNew parameter to copy() function in ObjectBuilder

shaymargolis:


src/Propel/Generator/Builder/Om/ObjectBuilder.php
```diff
@@ -6212,15 +6212,16 @@ protected function addCopy(&$script)
      * objects.
      *
      * @param  boolean \$deepCopy Whether to also copy all rows that refer (by fkey) to the current row.
+     * @param  boolean \$makeNew Whether to reset autoincrement PKs and make the object new.
      * @return ".$this->getObjectClassName(true)." Clone of current object.
      * @throws PropelException
      */
-    public function copy(\$deepCopy = false)
+    public function copy(\$deepCopy = false, \$makeNew = true)
```
marcj: New arguments should behave like the old one, so default should be false.

Incognito:
```suggestion
    public function copy(\$deepCopy = false, \$makeNew = false)
```

This would improve backwards compatibility and prevent the need for a major.

Incognito (changes requested):
This would also require tests that make use of this new feature.

The concept seems great on paper, it would help a lot with situations where you want idempotent mutation.

Something else to consider: Is there a way to do this without adding a new parameter to the method? Boolean params are usually red-flags for Single Responsibility Principle violations.

gechetspr:

Issue #1213, opened 2016-06-09
Only a few PDO attributes can be configured in propel.database.connections.default.attributes, in contrast to what is mentioned in docs


motin:
On http://propelorm.org/documentation/reference/configuration-file.html#attributes it says that all standard PDO attributes can be set as well as a Propel-specific attribute PropelPDO::PROPEL_ATTR_CACHE_PREPARES

It further mentions that attributes in the configuration file can be specified with or without the PDO:: (or PropelPDO::) constant prefix.

However, [PropelConfiguration.php restricts the the possible attributes to a few values (ATTR_EMULATE_PREPARES and ATTR_TIMEOUT)](https://github.com/propelorm/Propel2/blob/df9277744796c39c3f8dde8da57fb4304be18838/src/Propel/Common/Config/PropelConfiguration.php#L130-L131). Setting other attributes results in the following exception when for instance creating a migration (`vendor/bin/propel -vvv migration:create`):

```
  [Symfony\Component\Config\Definition\Exception\InvalidConfigurationException]
  Unrecognized options "ATTR_ERRMODE, PROPEL_ATTR_CACHE_PREPARES" under "propel.database.connections.default.attributes" 
```

Note that neither the PDO:: nor PropelPDO:: prefixes are allowed neither.

Either the docs should be updated or Propel should be configured to allow setting the mentioned attributes.

marcj:
Yeah probably we should not limit those children. @cristianoc72 do you have an idea how to solve that elegantly?

marcj:
Fixed via #1279.

Issue #1278, opened 2016-09-28
PDO::ATTR_PERSISTENT not recognized by config attributes


sleenen:
When adding 

'attributes' => [
    PDO::ATTR_PERSISTENT => true,
],

to connection settings, I get the following error:

[Symfony\Component\Config\Definition\Exception\InvalidConfigurationException]
  Unrecognized option "12" under "propel.database.connections.default.attributes"

The PropelConfiguration class only has ATTR_EMULATE_PREPARES has a valid child. 

Why are lots of the other PDO attributes not added? Is there a reason for this?

The options part does have a 'ATTR_PERSISTENT', but what does that actually do? Does that apply the PDO::ATTR_PERSISTENT => true ?

sleenen:
#1213 , hadn't seen that one, my bad.

marcj:
Fixed via #1279.

PR #1279, opened 2016-10-04
Compatibility of numeric and string keys

Userator:
Compatibility of numeric and string keys


Userator:
this pull request possibly fix #1213 #1278

PR #1286, opened 2016-10-17
[WIP][data-mapper]Naming strategy

cristianoc72:
Propel object-model classes come with a pletora of name properties: `name`, `phpName`, `tableName`, `columnName`, `camelCaseName`, etc.
All those classes, using object model, has the responsibility to correctly format those names.
With this PR we introduce a new naming strategy, simpler and more coherent.
This is a big refactor but it involves only Propel internals, not the api (except a little difference in the `schema.xml`).
## Innovations
1. Now we have only `name` and `sqlName` properties.
2. Accessor methods have the responsibility to correctly format the returning strings.
3. No more `phpName` into the `schema.xml`.
4. Introduce `sqlName` into the `schema.xml`.
### In details
- **name** property is the name of the object to use into the PHP code. It's _forced_ to _camelCase_. The name property of Entity objects is forced to _CamelCase_ (first letter capital).Eg.

``` xml
<!-- schema.xml -->
<entity name="my-entity">
    <field name="my-field"/>
</entity>
```

The name property of the Entity object will be `MyEntity`. 
The name property of the Field object will be `myField`.
- **sqlName** property is the name of the object to use into SQL statements. It's not forced to a predefined format, to avoid problems working with already existent databases.
  If it's not defined by the user, it's automatically set to the _underscore_ version of the `name` property. It can be set into the schema.xml file by using `sqlName` attribute.

``` xml
............. 
<field name="my-field" sqlName="MYFIELD"/>
```

The name property of the Field object will be `myField` and the sqlName property of the Field object will be `MYFIELD`.

``` xml
............. 
<field name="my-field" />
```

The name property of the Field object will be `myField` and the sqlName property of the Field object now is `my_field`.
- **accessors** Every model class (implementing `MappingModelInterface`) must have a `getName`
  and a `getSqlName` methods. Those methods return the correctly formatted strings (camelCase or underscore). In some cases, we can pass a boolean argument, to return a fully qualified name or a fully qualified sql name. 
- **phpName** attribute removed from `schema.xml`. The `name` property is almost the same as the old phpName.
- **sqlName** We introduce this attribute in `schema.xml` to define the name of a database element, when we work with an already existent database, or when the user whish to define his own naming strategy for database.
## Steps to go
- [x] Implement the new naming strategy
- [x] Enjoy no regression (except a couple of skipped or incomplete tests)
- [x] `Propel\Generator\Model` tests are green
- [x] `Propel\Generator\Platform` tests are green
- [x] Fix skipped or incomplete tests
## Future improvements

Think about a sort of abstraction, to let the user choose his own naming strategy. Maybe 
via configuration file.


marcj:
Good work! What if we not transform the `name` value, so we need no naming strategy and everybody can use what they want (I prefer camelCase, someone else underscore). I actually don't like to be forced to write now kebab-case. Also, when we have later annotations like in Doctrine we need to use the actual class property as name (same for the entity:name) and can not modify it.

cristianoc72:
Well, I can see your point and this strategy is too strict for me, too. Anyway, I think we should take a well defined decision. Let me explain better. 

Of course, Propel object model classes need two kind of names: a name representing the class and a name representing the real database element, mapped by the class.
Traditionally, Propel has a `name` property, taken directly from the xml schema, representing the database element, and a `phpName`, to be used in php code.
Into the *data-mapper* branch, the `name` property is considered to be the class name and, obviously, phpName was removed.
But this approch causes the failure of many test cases, working with schema:
```php
//Propel\Generator\Model\Database
$database = new Database();

//Propel\Generator\Model\Entity
$entity = new Entity('entity');
$database->addEntity($entity);

$entity1 = new Entity('entity');
$entity1->setSchema('bis');

//this fails
$database->addEntity($entity1);
```
This test case fails, because addEntity method checks if the added entity is already into the database, searching on the `name` property. But if this property represents the class, it can't distinguish from different schemas, so an exception is thrown because we have two entities with the same name.

I can see two solutions:
1. continue with the traditional naming and resume `phpName` property (I'm trying this way, now)
2. embrace this refactor

If we continue with this refactor, imho the `name` property should be forced to camelCase, because it should follow PSR-1 standard.

Of course, any other suggestion is wellcome! 😄

marcj:
Actually, it's ok that this test fails because you can not have two of the same entity class in a database but with different schemas. The ID of an entity is its php class name. Guess there's no need for a phpName attribute.

cristianoc72:
Ok. Let's sum up:

1. leave the `name` value as written into `schema.xml`
2. definitively remove phpName
3. given 1, we should transfor "dinamically" (via accessors or NamingTrait methods) the name of generated classes an properties to follow PSR-1, without save them into the `name` property.

If it's ok, I'm going to close this PR and restart from the last commit of  `data-mapper` branch.

marcj:
@cristianoc72 I would not transform stuff from field::name or relation::field. just use it as is, so people are free to name their classes however they want :)

marcj:
ok, I pushed my latest changes.

marcj:
screen shot 2016-11-11 at 14 41 16

only ~600 left :)

cristianoc72:
Great!! :clap:

marcj:
@cristianoc72 time to rebase 😁 we have only 100 errors left in agnostic test suite 🎉

cristianoc72:
Yeah!! Can't wait!! Next week I'm going to work again on Propel: I'ld like
to finish versionable behavior port. It's one of the most used behaviors
and new Propel version can't miss it!

Il 19/nov/2016 03:32, "Marc J. Schmidt" <notifications@github.com> ha
scritto:

marcj:
@cristianoc72 nice! Worked the whole day on it. 62 errors left for agnostic tests 👍  Btw, I fixed your Sortable issue with scope filtering by 'NULL' (as string, instead of null). Was actually my issue in the persister :)

PR #1288, opened 2016-10-18
Static Code Analysis: minor code changes

kalessil:
- usage of elvis operator;
- usage of boolean casting instead of double negation;
- usage of prefix increment;
- removed unnecessary semicolons;
- usage of newer random API functions;
- usage of original functions instead of aliases;
- fixed array_push miss-usage;
- usage of type casting instead of functions;


marcj:
Thanks, but I don't think we are going to merge this.

kalessil:
No problems, thanks for the feedback.

PR #1292, opened 2016-11-05
removing leading space within the sqlite dsn

nickdietel:
the extra space was causing the below error when using an absolute path like /home/my.app.sq3
Unable to connect to the specific sql server: SQLSTATE[HY000] [14] unable to open database file


marcj:
Good catch, thanks!

nickdietel:
you're welcome ☺

PR #1293, opened 2016-11-07
Use DATETIME2 to allow high precision timestamps

jhaisley:
DATETIME timestamps are limited to 3 digit precision, where DATETTIME2 allows 7 digits. The timestampable behavior needs this precision. 

Example of types: 
datetime	2007-05-08 12:35:29.123
datetime2	2007-05-08 12:35:29. 1234567


MDT-FGI:
I can confirm that this change fixes the error. Merge suggested.

marcj:
Thanks!

Issue #1225, opened 2016-06-16
Error with persistent connections to MySQL


emicks:
After updating to the latest dev-master (via composer on 2016-06-16), I'm facing the following exception when connecting to a MySQL (5.6.25) database:

```
PDOException: SQLSTATE[HY000]: General error: PDO::ATTR_STATEMENT_CLASS cannot be used with persistent PDO instances

in vendor\propel\propel\src\Propel\Runtime\Connection\PdoConnection.php at line 85
at PDO ->setAttribute ('13', array('\Propel\Runtime\Adapter\Pdo\PdoStatement', array())) 
in vendor\propel\propel\src\Propel\Runtime\Connection\PdoConnection.php at line 85
at PdoConnection ->setAttribute ('13', array('\Propel\Runtime\Adapter\Pdo\PdoStatement', array())) 
in vendor\propel\propel\src\Propel\Runtime\Connection\PdoConnection.php at line 60 
at PdoConnection ->__construct ('mysql:host=127.0.0.1;dbname=xxxx;charset=utf8mb4', 'xxxx', 'xxxx', array('ATTR_PERSISTENT' => true)) 
in vendor\propel\propel\src\Propel\Runtime\Adapter\Pdo\PdoAdapter.php at line 70 
at PdoAdapter ->getConnection (array('adapter' => 'mysql', 'user' => 'xxxx', 'password' => 'xxxx', 'dsn' => 'mysql:host=127.0.0.1;dbname=xxxx;charset=utf8mb4', 'options' => array('ATTR_PERSISTENT' => true), 'attributes' => array('ATTR_EMULATE_PREPARES' => true), 'settings' => array('charset' => 'utf8mb4', 'queries' => array()), 'classname' => '\Propel\Runtime\Connection\DebugPDO', 'slaves' => array())) 
in vendor\propel\propel\src\Propel\Runtime\Connection\ConnectionFactory.php at line 41  
[...]

PR #1295, opened 2016-11-11
persistent connections working again

prgTW:
Persistent connections are available again by not overriding PDO statement class

fixes this error:
```
Warning: PDO::setAttribute(): SQLSTATE[HY000]: General error: PDO::ATTR_STATEMENT_CLASS cannot be used with persistent PDO instances
```

ref: propelorm/Propel2#1225


marcj:
We should rather completely drop Propel\Runtime\Adapter\Pdo\PdoStatement instead of using it only for HHVM.

staabm:
alternatively we should drop HHVM support. not sure this runtime is relevant for somebody.

marcj:
or both :P

prgTW:
With this PR I wanted only to make persistent connections available again and didn't want to alter HHVM in any way cause I don't use it and don't have it and didn't want to make any backward-incompatible changes. Droping HHVM support could be a separate PR though

prgTW:
so ... any other thoughts about merging this one and restoring persistent connections in propel?

marcj:
What do you mean by dropping hhvm support? This drops basically the overwritten class support, for whatever that was used. Let us just fix it the right way and remove `Propel\Runtime\Adapter\Pdo\PdoStatement` completely, instead of using it only for hhvm.

prgTW:
@marcj I removed `Propel\Runtime\Adapter\Pdo\PdoStatement` class as You suggested

src/Propel/Runtime/Adapter/Pdo/PdoStatement.php
```diff
@@ -1,101 +0,0 @@
...
- * This file is part of the Propel package.
- * For the full copyright and license information, please view the LICENSE
- * file that was distributed with this source code.
- *
- * @license MIT License
- */
-
-namespace Propel\Runtime\Adapter\Pdo;
-
-use Propel\Runtime\Connection\StatementInterface;
```
prgTW: actually it's used in quite few places, shouldn't it's usages be converted to usages of `\PdoStatement` as well, am I getting You right?

prgTW: removed

marcj:
Nice, good work, thanks!

prgTW:
great news, thanks you too :)

marcj:
Backported to data-mapper branch.

denormalizer:
Hi guys. 

This pull request broke something that used to work: #1413

In composer.json, I've currently had to make sure propel use the commit prior to this:

    "propel/propel": "dev-master#38f1044e9b57cb7f68b59224f1830bde7009673f"

Does anyone have a workaround so I don't have to be stuck on this commit?

PR #1301, opened 2016-11-25
Check for null key in InstancePoolTrait

drodil:
Sometimes getInstanceKey returns null and it messes the instance pool. I think it has something to do with the joinWith populating the related objects.


rabidus:
This happens when you are saving your entity while alreadyInSave, now that not saved entity will go into instance pool.

Now InstancePoolTrait::getInstanceKey will call your entity getPrimaryKey() which returns null, and will be casted to empty string, and saved into instance pool with empty string key. Messes things up.

PR #1313, opened 2016-12-13
Fix collision of class names between namespaces

Almamu:
Fixes an use case like this one:
```
<database name="default" defaultIdMethod="native" defaultPhpNamingMethod="underscore" namespace="Database">
  <table name="arclabs_accounts" idMethod="native" phpName="Account" namespace="ArcLaboratories">
    [...]
  </table>
  <table name="mlabs_accounts" idMethod="native" phpName="Account" namespace="MinecraftLaboratories">
    [...]
  </table>
</database>
```

### DIFF
--- a/src/Propel/Generator/Builder/Om/AbstractOMBuilder.php
+++ b/src/Propel/Generator/Builder/Om/AbstractOMBuilder.php
@@ -328,7 +328,9 @@ abstract class AbstractOMBuilder extends DataModelBuilder
                 //so use the fqcn
                 return ($namespace ? '\\' . $namespace : '') .  '\\' . $class;
             } else {
-                $autoAliasName = 'Child' . $class;
+                // $autoAliasName = 'Child' . $class;
+                // FIXME: USE A DIFFERENT SYSTEM TO GENERATE AN UNIQUE NAME FOR THIS CLASS?
+                $autoAliasName = $class . md5($namespace);
             }
 
             return $this->declareClassNamespace($class, $namespace, $autoAliasName);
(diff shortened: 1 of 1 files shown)


src/Propel/Generator/Builder/Om/AbstractOMBuilder.php
```diff
@@ -328,7 +328,9 @@ public function declareClassNamespace($class, $namespace = '', $alias = false)
                 //so use the fqcn
                 return ($namespace ? '\\' . $namespace : '') .  '\\' . $class;
             } else {
-                $autoAliasName = 'Child' . $class;
+                // $autoAliasName = 'Child' . $class;
+                // FIXME: USE A DIFFERENT SYSTEM TO GENERATE AN UNIQUE NAME FOR THIS CLASS?
+                $autoAliasName = $class . md5($namespace);
```
marcj: Oh, well, no md5 in names please :P

Almamu: That just change the aliases generated for class names (for the use in autogenerated classes, doesnt change any class name) I do agree that MD5 is not the best way, but there are not many ways of doing it. At first I though to generate a name based on the namespace and that could probably be better... for example, get the last namespace and concat it to the class name so a case like the one on the PR generates valid use statements that do not collide. That would generate aliases like MinecraftLaboratoriesAccount and ArcLaboratoriesAccount in case of a collision in the class name. Would that be a better approach then?

PhilinTv:
```
$nsSuffix = str_replace('\\', '', implode('', array_slice(explode('\\', $namespace), -2)));
$autoAliasName = $nsSuffix . $class;
```

+ test

gechetspr:

PR #1315, opened 2016-12-14
Reverse engineering

flapy2b:
Put automaticaly the isCrossRef attribute on the tables if needed.
Works only for MySQL.

### DIFF
--- a/src/Propel/Generator/Reverse/MysqlSchemaParser.php
+++ b/src/Propel/Generator/Reverse/MysqlSchemaParser.php
@@ -115,15 +115,15 @@ class MysqlSchemaParser extends AbstractSchemaParser
 
             $this->addTableVendorInfo($table);
             
-            //Check if the table is the crosstab of a many-to-many relation
-            $nbPks = count($table->getPrimaryKey());
+         //Check if the table is the crosstab of a many-to-many relation
+        	$nbPks = count($table->getPrimaryKey());
             $nbFks = count($table->getForeignKeys());
             if($nbPks > 1 && $nbFks > 1 && $nbPks == $nbFks) {
-            	$flag = false;
-            	for ($i = 0; $i < count($table->getPrimaryKey()); $i++){
-            		$flag = $table->getPrimaryKey()[$i]->getName() == $table->getForeignKeys()[$i]->getLocalColumnName();
+            	$isCrossRef = true;
+            	foreach ($table->getForeignKeys() as $foreignKey) {
+            		$isCrossRef &= $foreignKey->isForeignPrimaryKey();
             	}
-            	$table->setCrossRef($flag);
+            	$table->setCrossRef($isCrossRef);
             }
         }
 
--- a/src/Propel/Generator/Reverse/MysqlSchemaParser.php
+++ b/src/Propel/Generator/Reverse/MysqlSchemaParser.php
@@ -114,6 +114,18 @@ class MysqlSchemaParser extends AbstractSchemaParser
             $this->addPrimaryKey($table);
 
             $this->addTableVendorInfo($table);
+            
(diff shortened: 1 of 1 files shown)


src/Propel/Generator/Reverse/MysqlSchemaParser.php
```diff
@@ -114,6 +114,18 @@ public function parse(Database $database, array $additionalTables = [])
             $this->addPrimaryKey($table);
 
             $this->addTableVendorInfo($table);
+            
+            //Check if the table is the crosstab of a many-to-many relation
+            if(count($table->getPrimaryKey()) > 1 &&
```
staabm: please use a local var instead of counting again and again

marcj:
Why does it only works for MySQL?

flapy2b:
@marcj because I have only tested with MySQL for the moment and I worked only on MysqlSchemaParser.php

marcj:
This check is not very elegant. You could just check if all foreign keys are primary keys using 

```php
$isCrossRef = count($table->getForeignKeys()) > 0;
foreach ($table->getForeignKeys() as $foreignKey()) {
    $isCrossRef &= $foreignKey->isForeignPrimaryKey();
}
```

src/Propel/Generator/Reverse/MysqlSchemaParser.php
```diff
@@ -116,15 +116,14 @@ public function parse(Database $database, array $additionalTables = [])
...
-            			$flag = false;
-            			for ($i = 0; $i < count($table->getPrimaryKey()); $i++){
-            				$flag = $table->getPrimaryKey()[$i]->getName() == $table->getForeignKeys()[$i]->getLocalColumnName();
-            			}
-            			$table->setCrossRef($flag);
+            $nbPks = count($table->getPrimaryKey());
+            $nbFks = count($table->getForeignKeys());
+            if($nbPks > 1 && $nbFks > 1 && $nbPks == $nbFks) {
+            	$flag = false;
+            	for ($i = 0; $i < count($table->getPrimaryKey()); $i++){
```
staabm: $nbPks

src/Propel/Generator/Reverse/MysqlSchemaParser.php
```diff
@@ -114,6 +114,17 @@ public function parse(Database $database, array $additionalTables = [])
             $this->addPrimaryKey($table);
 
             $this->addTableVendorInfo($table);
+            
+         //Check if the table is the crosstab of a many-to-many relation
```
Incognito, line 118: Method is increasing in complexity with each new detail added, it would be great if someone could jump in and break up the jobs done in this method.

Incognito: Something with the indentation is awkward in this patch.

src/Propel/Generator/Reverse/MysqlSchemaParser.php
```diff
@@ -114,6 +114,17 @@ public function parse(Database $database, array $additionalTables = [])
...
+        	$nbPks = count($table->getPrimaryKey());
+            $nbFks = count($table->getForeignKeys());
+            if($nbPks > 1 && $nbFks > 1 && $nbPks == $nbFks) {
```
Incognito, line 121: This logic can be simplified, for example:

```suggestion
            if($nbPks === $nbFks && $nbPks > 1) {
```

Incognito: Does this ensure that the table is always what is expected 100% of the time in all possible cases, or does it simply tend to be this way in some specific/common usecases?

dereuromark:
Is there any follow up or shall we close this?

Incognito:
+1 close

PR #1319, opened 2016-12-30
Added polymorphic many-to-many relationships.

wleona3:
Allows for polymorphic many-to-many relationships. Unfortunately at this time does not allow additional static FKs. The builder almost supports it, but since each polymorphic sibling requires a unique relation, the builder tries to also create a new relation for the other static FKs. Example: I have a table that has two static FK relationships A and B and two polymorphic siblings, C1 and C2. When creating object A the builder has to create properties and methods for the unique C1 and C2 relationships, but then also creates two B relationships, creating duplicate properties and methods. Not sure exactly how to implement that properly. But if that caveat is fixed, this would support any combination of static FKs and Polymorphic FKs.

### DIFF
--- a/src/Propel/Common/Util/CartesianProduct.php
+++ b/src/Propel/Common/Util/CartesianProduct.php
@@ -1,0 +1,43 @@
+<?php namespace Propel\Common\Util;
+
+class CartesianProduct
+{
+    public static function get($input) {
+        $result = array();
+
+        while (list($key, $values) = each($input)) {
+            if (empty($values)) {
+                continue;
+            }
+
+            if (empty($result)) {
+                if (!is_array($values)) {
+                    $values = [$values];
+                }
+                foreach($values as $value) {
+                    $result[] = array($key => $value);
+                }
+            }
+            else {
+                $append = array();
+
+                foreach($result as &$product) {
+                    $product[$key] = array_shift($values);
+
+                    $copy = $product;
+
+                    foreach($values as $item) {
+                        $copy[$key] = $item;
+                        $append[] = $copy;
+                    }
+
+                    array_unshift($values, $product[$key]);
+                }
+
+                $result = array_merge($result, $append);
+            }
+        }
+
+        return $result;
+    }
+}
--- a/src/Propel/Generator/Builder/Om/ObjectBuilder.php
+++ b/src/Propel/Generator/Builder/Om/ObjectBuilder.php
@@ -4458,10 +4458,16 @@ abstract class ".$this->getUnqualifiedClassName().$parentClass." implements Acti
             foreach ($crossFKs->getIncomingForeignKey()->getColumnObjectsMapping() as $reference) {
                 $local   = $reference['local'];
                 $foreign = $reference['foreign'];
+                $value   = $reference['value'];
 
                 $idx = array_search($local, $crossPks, true);
-                $script .= "
+                if ($value) {
+                    $script .= "
+                        \$entryPk[$idx] = '$value';";
+                } else {
+                    $script .= "
                         \$entryPk[$idx] = \$this->get{$foreign->getPhpName()}();";
+                }
             }
 
             $combinationIdx = 0;
@@ -4469,10 +4475,16 @@ abstract class ".$this->getUnqualifiedClassName().$parentClass." implements Acti
                 foreach ($crossFK->getColumnObjectsMapping() as $reference) {
                     $local   = $reference['local'];
                     $foreign = $reference['foreign'];
+                    $value   = $reference['value'];
 
                     $idx = array_search($local, $crossPks, true);
-                    $script .= "
+                    if ($value) {
+                        $script .= "
+                        \$entryPk[$idx] = '$value';";
+                    } else {
+                        $script .= "
                         \$entryPk[$idx] = \$combination[$combinationIdx]->get{$foreign->getPhpName()}();";
+                    }
                 }
                 $combinationIdx++;
             }
@@ -4506,20 +4518,32 @@ abstract class ".$this->getUnqualifiedClassName().$parentClass." implements Acti
             foreach ($crossFKs->getIncomingForeignKey()->getColumnObjectsMapping() as $reference) {
                 $local   = $reference['local'];
                 $foreign = $reference['foreign'];
+                $value   = $reference['value'];
 
                 $idx = array_search($local, $crossPks, true);
-                $script .= "
+                if ($value) {
+                    $script .= "
+                        \$entryPk[$idx] = '$value';";
+                } else {
+                    $script .= "
                         \$entryPk[$idx] = \$this->get{$foreign->getPhpName()}();";
+                }
             }
 
             $crossFK = $crossFKs->getCrossForeignKeys()[0];
             foreach ($crossFK->getColumnObjectsMapping() as $reference) {
                 $local   = $reference['local'];
                 $foreign = $reference['foreign'];
+                $value   = $reference['value'];
 
                 $idx = array_search($local, $crossPks, true);
-                $script .= "
+                if ($value) {
+                    $script .= "
+                        \$entryPk[$idx] = '$value';";
+                } else {
+                    $script .= "
                         \$entryPk[$idx] = \$entry->get{$foreign->getPhpName()}();";
+                }
             }
 
             $script .= "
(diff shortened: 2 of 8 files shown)


marcj:
Wow, that's a rather big change, thanks dude! I'd rather want to see that beauty in https://github.com/propelorm/Propel3 :P

wleona3:
Thanks for the kind words marcj, unfortunately our company will be stuck with PHP 5.6 for the foreseeable future so we'll be sticking with Propel2

src/Propel/Generator/Builder/Om/ObjectBuilder.php
```diff
@@ -4423,21 +4423,33 @@ protected function addCrossFkScheduledForDeletion(&$script, CrossForeignKeys $cr
             foreach ($crossFKs->getIncomingForeignKey()->getColumnObjectsMapping() as $reference) {
                 $local   = $reference['local'];
                 $foreign = $reference['foreign'];
+				$value   = $reference['value'];
```
marcj: tabs vs space

wleona3:
Whitespace corrected!

Incognito (commented):
This branch passes CI when rebased on latest master. The new tests also pass when rebased.

Because it's been open for a while we should give this a proper read-over to check how the feature works and see if there's any code changes needed to meet the new standards on the project.

It looks like a really cool feature to add to Propel.

tests/Propel/Tests/Generator/Builder/Om/GeneratedObjectM2MRelationPolymorphicWithSelfReference.php
```diff
@@ -0,0 +1,133 @@
...
+     * set    |    9            10            11           12
+     *
+     */
+
+
+
+    /*
+     * ####################################
+     * 1. add, add
+     */
```
Incognito, line 77:
```suggestion
    /**
     * |        | Add | Remove | Set | Get |
     * |--------|-----|--------|-----|-----|
     * | Add    |  1  |    2   |  3  |  4  |
     * | Remove |  5  |    6   |  7  |  8  |
     * | Set    |  9  |   10   |  11 |  12 |
     *
     * Test the intersection of `Add` and `Add`.
     */
```

tests/Propel/Tests/Generator/Builder/Om/GeneratedObjectM2MRelationPolymorphicWithSelfReference.php
```diff
@@ -0,0 +1,133 @@
...
+    public function test1()
```
Incognito, line 78:
```suggestion
    public function testCombinations()
```

I suspect this test should be broke into three tests, based on the cyclical appearance of setup + assert code in here.

tests/Propel/Tests/Generator/Builder/Om/GeneratedObjectM2MRelationPolymorphicWithSelfReference.php
```diff
@@ -0,0 +1,133 @@
...
+use Propel\Generator\Util\QuickBuilder;
+use Propel\Runtime\Collection\ObjectCollection;
+use Propel\Runtime\Collection\ObjectCombinationCollection;
+use Propel\Tests\Helpers\PlatformDatabaseBuildTimeBase;
+
+
+class GeneratedObjectM2MRelationPolymorphic extends PlatformDatabaseBuildTimeBase
+{
+    /**
+     * Tests for a M2M relation with three pks where each is a FK.
```
Incognito, line 13:
```suggestion
     * Tests for a Many-to-Many relation with three PKs where each has its own FK.
```

tests/Propel/Tests/Generator/Builder/Om/GeneratedObjectM2MRelationPolymorphicWithSelfReference.php
Incognito: I'm not 100% sure if these comments are meaningful or should be removed, I just wanted to improve the presentation of it.

src/Propel/Common/Util/CartesianProduct.php
```diff
@@ -0,0 +1,43 @@
+<?php namespace Propel\Common\Util;
+
+class CartesianProduct
```
Incognito, line 3: I would like to see some way to limit the size of this. Cartesian products end up being expensive for both CPU and RAM in easily process-crashing-ways.

src/Propel/Common/Util/CartesianProduct.php
```diff
@@ -0,0 +1,43 @@
...
+{
+    public static function get($input) {
```
Incognito, line 5: There's a good opportunity to decrease the logical nesting here. If I get a moment I might propose an alternative that uses generators (since propel is now 7.1+).

src/Propel/Generator/Builder/Om/ObjectBuilder.php
```diff
@@ -4458,21 +4458,33 @@ protected function addCrossFkScheduledForDeletion(&$script, CrossForeignKeys $cr
...
 
                     $idx = array_search($local, $crossPks, true);
-                    $script .= "
+                    if ($value) {
+                        $script .= "
+                        \$entryPk[$idx] = '$value';";
+                    } else {
+                        $script .= "
                         \$entryPk[$idx] = \$combination[$combinationIdx]->get{$foreign->getPhpName()}();";
+                    }
```
Incognito, line 4487: Because `addCrossFkScheduledForDeletion` is already quite complex we should do our best to keep reducing it whenever possible. I see an opportunity here to simply put this additional logic into a protected method.

It could also help reduce duplication, as the other two code hunks in this patch look very similar to me.

src/Propel/Generator/Model/ForeignKey.php
```diff
@@ -1019,4 +1019,30 @@ public function isAtLeastOneLocalPrimaryKey()
 
         return 0 !== count($cols);
     }
+
+    public function isTheSamePolymorphicSignature(ForeignKey $fk)
```
Incognito, line 1023: There should be a discussion around the names introduced in this model as they're going to become highly visible.

src/Propel/Generator/Model/ForeignKey.php
```diff
@@ -1019,4 +1019,30 @@ public function isAtLeastOneLocalPrimaryKey()
...
+    public function getPolymorphicSignature()
+    {
+        if (!$this->isPolymorphic()) {
+            return false;
+        }
+
+        $sortedCols = $this->getLocalColumns();
+        sort($sortedCols);
+
+        return sha1(implode('|||', $sortedCols));
```
Incognito, line 1046: I don't know about everyone else, but any time I've wanted to generate something like a "unique signature" it's always come back to haunt me in 2 ways:

1. It ends up being slow when called hundreds of times in a loop.
2. It ends up being really hard to change in a BC way because it isn't abstracted out of the class itself.

It might be a good feature but we should look at improving these two points. Perhaps the signature generation could be even simpler than it is, and/or it could also be moved out of the class.

src/Propel/Generator/Model/Table.php
```diff
@@ -819,15 +820,39 @@ public function getCrossFks()
         $crossFks = [];
         foreach ($this->referrers as $refFK) {
             if ($refFK->getTable()->isCrossRef()) {
-                $crossFK = new CrossForeignKeys($refFK, $this);
-                foreach ($refFK->getOtherFks() as $fk) {
-                    if ($fk->isAtLeastOneLocalPrimaryKeyIsRequired() &&
-                        $crossFK->isAtLeastOneLocalPrimaryKeyNotCovered($fk)) {
-                        $crossFK->addCrossForeignKey($fk);
+                // group all polymorphic fks
```
Incognito, line 823: Let's find a way to reduce the complexity of this class. It was already 2 nested foreach loops, both nested inside if statements, this has now increased a fair bit.

I'd also like to understand more about the performance impact of this change, and what cases users of Propel will need to be cautious about getting into.

Incognito (changes requested):
This is a really cool feature and I think it will be useful to anyone with polymorphic needs.

PhilinTv:
Justification for Beta6.
- An advanced ORM feature that is useful in complex data modeling (e.g., tagging, permission systems).
- Extensive test coverage is included (including polymorphic + self-referencing cases).
- Well-isolated implementation via ForeignKey signature logic and cartesian combinator.
- DX complexity is manageable, and performance tradeoffs are acceptable.

gechetspr:

Issue #1321, opened 2017-01-01
Support of XML Inclusions for the XML Configuration files


hakre:
The XML standard(s) come with a nice feature called *XML Includes* which allows to use (parts) from other XML files to be included into a document.

For example, you can create mixins for the configuration files. E.g. define the database structure, but use a different database name in multi-tenant application.

dereuromark:
Looks like PR was merged.

PR #1322, opened 2017-01-01
Support XML Inclusions in configuration, adds #1321

hakre:
Implemented via the existing functionality in PHP's DOMDocument object.

Additionally libxml_disable_entity_loader(true) is not necessary any
longer (it prevented opening XML files from disc earlier) since Oct 2012
with the release of libxml2 version 2.9.0.

The old code looked influenced by the (earlier to that date) hotfix in
the Symfony project however the XXE attack Symfony was prone to, did not
apply to the XmlToArrayConverter (no validate on parse).

Refs:

- #1321

- https://www.w3.org/TR/xinclude/

- https://mail.gnome.org/archives/xml/2012-October/msg00045.html


marcj:
Perfect, thanks!

PR #1326, opened 2017-01-05
set foreignSchema when schema is public

domcar:
Since Propel ignore the public schema of Postgres, there is a problem when in the schema.xml there is a foreign-key with reference to a foreignTable that is located in the public schema. 

In this example the table `approvals` is in `schema1` while the foreignTable `orders` is located in `public`, this generates a problem because the reverse-engineer will look for the foreignTable in `schema1`:
```
<table name="approvals" schema="schema1" idMethod="native" >
    <column name="orderid" phpName="Orderid" required="true"/>
    <foreign-key foreignTable="orders" name="approvals_orderid_fk">
      ..........
    </foreign-key>
```

With this fix it is possible to reference a foreignTable when it is in the public schema:
```
<table name="approvals" schema="schema1" idMethod="native" >
    <column name="orderid" phpName="Orderid" required="true"/>
    <foreign-key foreignTable="orders" foreignSchema="public" name="approvals_orderid_fk">
      ..........
    </foreign-key>

### DIFF
--- a/src/Propel/Generator/Reverse/PgsqlSchemaParser.php
+++ b/src/Propel/Generator/Reverse/PgsqlSchemaParser.php
@@ -417,9 +417,9 @@ class PgsqlSchemaParser extends AbstractSchemaParser
                 $fk = new ForeignKey($name);
                 $fk->setForeignTableCommonName($foreignTable->getCommonName());
                 if ($table->guessSchemaName() != $foreignTable->guessSchemaName()) {
-                     if($foreignTable->getSchema())
+                        if($foreignTable->getSchema())
                             $fk->setForeignSchemaName($foreignTable->getSchema());
-                     elseif($foreignTable->getSchema() == 'public')
+                        else
                             $fk->setForeignSchemaName("public");
                 }
                 $fk->setOnDelete($ondelete);
--- a/src/Propel/Generator/Reverse/PgsqlSchemaParser.php
+++ b/src/Propel/Generator/Reverse/PgsqlSchemaParser.php
@@ -417,7 +417,10 @@ class PgsqlSchemaParser extends AbstractSchemaParser
                 $fk = new ForeignKey($name);
                 $fk->setForeignTableCommonName($foreignTable->getCommonName());
                 if ($table->guessSchemaName() != $foreignTable->guessSchemaName()) {
-                    $fk->setForeignSchemaName($foreignTable->getSchema());
+                     if($foreignTable->getSchema())
+                            $fk->setForeignSchemaName($foreignTable->getSchema());
+                     elseif($foreignTable->getSchema() == 'public')
+                            $fk->setForeignSchemaName("public");
                 }
                 $fk->setOnDelete($ondelete);
                 $fk->setOnUpdate($onupdate);

### CODE REVIEW

File: src/Propel/Generator/Reverse/PgsqlSchemaParser.php [hunk 1]
```diff
@@ -417,7 +417,10 @@ protected function addForeignKeys(Table $table, $oid)
                 $fk = new ForeignKey($name);
                 $fk->setForeignTableCommonName($foreignTable->getCommonName());
                 if ($table->guessSchemaName() != $foreignTable->guessSchemaName()) {
-                    $fk->setForeignSchemaName($foreignTable->getSchema());
+                     if($foreignTable->getSchema())
+                            $fk->setForeignSchemaName($foreignTable->getSchema());
+                     elseif($foreignTable->getSchema() == 'public')
```
staabm: what is "the public" schema in pg exactly? something internal?

domcar: As stated by the Postgres documentation:

_tables created without specifying any schema names, by default such tables (and other objects) are automatically put into a schema named "public". Every new database contains such a schema_

Here is the link: https://www.postgresql.org/docs/9.1/static/ddl-schemas.html Section 5.7.2

marcj: This code can not work. When First condition `if($foreignTable->getSchema())` evaluates to false it never can be true for `$foreignTable->getSchema() == 'public'`.

domcar: Why shouldn't it work? If the condition is `false` then we have no `schema` for the foreignTable, that's it (although there should always be a schema).
But if it evaluates to `true` AND it is equal to public, then the `foreignSchemaName` will be set to public. (Maybe instead of `elseif` I should have used only `if`).
Without it, the `public` schema will not be set.

marcj: ``php
if($foreignTable->getSchema())
    $fk->setForeignSchemaName($foreignTable->getSchema());
```

When getSchema() returns a value it will be set. If it does not  return a value you check via `elseif($foreignTable->getSchema() == 'public')`, which can never be truely because you checked already in the previous if that `$foreignTable->getSchema()` returns falsely. Falsely means it can never be `== 'public'`. If you would use instead of `elseif` simple `if`, you would set it twice. Too illustrate it:

```
$schema = 'public';
if ($schema)
    $fk->setForeignSchema($schema);
if ($schema == 'public')
    $fk->setForeignSchema('public');
```

As you see, the second check is redundant.

domcar: No, because the public schema doesn't get set, this is the problem. I don't know why but it doesn't get set!

marcj: That is weird. It should be set via the first `setForeignSchema()` call. If that is not the case, then the bug hides in that method. Calling however the same method twice with the same arguments directly after each other is really not a bugfix, and it is currently not logical at all. I just don't see how this fixes anything. Maybe you can provide a unit-test that proves that this codes fixes it.

domcar: Hi Marc, sorry you were right, i uploaded the wrong commit, please have a look at it now

marcj:
Mh, the thing is: Setting the schema to "public" per default can be wrong, since "public" is not always the default schema in Postgres. We should actually find out in which schema the target table is located and place it in "schema" attribute if it differs to the default schema (which we need to find out by querying the database using `SHOW search_path;`).

PhilinTv:
Complex story, not clearly fixable. It's time to close it.

PR #1328, opened 2017-01-09
WIP - Mssql fixes

cedriclombardot:


src/Propel/Runtime/Connection/ConnectionFactory.php
```diff
@@ -55,6 +55,11 @@ public static function create(array $configuration = [], AdapterInterface $adapt
                     }
                     $value = constant($value);
                 }
+                
+                if ($option == "ATTR_EMULATE_PREPARES" || $option == "ATTR_TIMEOUT") {
```
cedriclombardot: How set it per db type ?

marcj: Why would you exclude those?

cedriclombardot: Beacause when we use it in mssql it fail (Eg : http://stackoverflow.com/questions/16366709/php-pdo-sqlsrv-cant-set-logintimeout)

marcj: well, then the developer can simply not use these options?

cedriclombardot: They was added automatically in propel https://github.com/propelorm/Propel2/blob/97dcdae4aff3ae6ae5f5bcf356a72babaa9f293d/src/Propel/Common/Config/PropelConfiguration.php#L132

marcj: Oh, then we should change that :)

cedriclombardot:
Yes, but i dont know how do this properly without breaking mysql feature

src/Propel/Runtime/Connection/ConnectionFactory.php
marcj: ok, then maybe the check is better placed in Connection::setAttribute() which has the DSN and knows whether its MSSQL or not

marcj: This needs to be moved first until I can merge it :)

src/Propel/Runtime/Adapter/Pdo/MssqlAdapter.php
```diff
@@ -300,4 +300,25 @@ public function cleanupSQL(&$sql, array &$params, Criteria $values, DatabaseMap
...
     }
+    
+    /**
+     * Returns timestamp formatter string for use in date() function.
+     *
+     * @return string
+     */
+    public function getTimestampFormatter()
+    {
+        return 'Y-m-d H:i:s:000';
```
Incognito, line 311: Can MSSSQL support `Y-m-d H:i:s.u` or is `:000` required?

DarkAxi0m: Just finished some testing and was going to create a PR, when I noticed this.

DateTime   can use `Y-m-d H:i:s.u`;
But DateTime2 needs to use `Y-m-d H:i:s`;

_(not sure how to proceeded from here)_

dereuromark:
Replaced by https://github.com/propelorm/Propel2/pull/1881

PR #1330, opened 2017-01-09
Fix regression in ObjectCollection of Collections.

peter17:
The test I add was failing with:
```
Propel\Runtime\Exception\BadMethodCallException: Call to undefined method: hashCode

/home/peter/dev/Propel2/src/Propel/Runtime/Collection/Collection.php:558
/home/peter/dev/Propel2/src/Propel/Runtime/Collection/ObjectCollection.php:496
/home/peter/dev/Propel2/src/Propel/Runtime/Collection/ObjectCollection.php:496
/home/peter/dev/Propel2/src/Propel/Runtime/Collection/ObjectCollection.php:485
/home/peter/dev/Propel2/tests/Propel/Tests/Runtime/Collection/ObjectCollectionTest.php:277
```
because `ObjectCollection.php:496` calls `hashCode()` on itself with:
```
        if (is_object($object) && is_callable([$object, 'hashCode'])) {
            return $object->hashCode();
        }
```
The test returns `true` because `Collection.php` has a `__call()` method. It then fails because that method does not handle `hashCode`.

So I added a `hashCode` method in `Collection.php`.

Collections of collections worked in Propel1 (with class `PropelCollection`). They happen when you create Symfony forms with `'multiple' => true` when the subform returns a Collection.

### DIFF
--- a/src/Propel/Runtime/Collection/Collection.php
+++ b/src/Propel/Runtime/Collection/Collection.php
@@ -164,10 +164,10 @@ class Collection implements \ArrayAccess, \IteratorAggregate, \Countable, \Seria
     {
         return new CollectionIterator($this);
     }
-    
+
     /**
      * Count elements in the collection
-     * 
+     *
      * @return int
      */
     public function count()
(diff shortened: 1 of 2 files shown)


marcj:
Good find, thanks!

PR #1339, opened 2017-02-01
fix TableMapTrait.php

CrayD:
Fix translateFieldNames method in Propel/Runtime/Map/TableMapTrait.php.
Bug: if key == 0(id in my case) field name does not translated.


gulaandrij:
Any feedback on this? Can this be merged?

Issue #1347, opened 2017-02-12
Validator 3.2.2


angelk:
Discussion: https://github.com/propelorm/Propel2/commit/b7c8ac9052c972838ae74d6ad0f6fa9ea5f2a100

Fix that we are waiting: https://github.com/symfony/symfony/pull/21592
This must be fixed before I (or someone else) continue to investigate the problem with validator v3.2.2.

I'm creating this issue because the problem is bigger than I tough and to keep every one updated with the investigation :)

Explanation of the problem:

Constraint result with 3.2.1
```
array(2) {
  [0] =>
  class Symfony\Component\Validator\Constraints\Regex#91 (6) {
    public $message =>
    string(25) "Please enter a valid ISBN"
    public $pattern =>
    string(9) "/[^\d-]+/"
    public $htmlPattern =>
    NULL
    public $match =>
    bool(false)
    public $payload =>
    NULL
    public $groups =>
    array(3) {
      [0] =>
      string(7) "Default"
      [1] =>
      string(19) "ValidateTriggerBook"
      [2] =>
      string(22) "ValidateTriggerFiction"
    }
  }
  [1] =>
  class Symfony\Component\Validator\Constraints\Regex#93 (6) {
    public $message =>
    string(25) "Please enter a valid ISBN"
    public $pattern =>
    string(9) "/[^\d-]+/"
    public $htmlPattern =>
    NULL
    public $match =>
    bool(false)
    public $payload =>
    NULL
    public $groups =>
    array(2) {
      [0] =>
      string(7) "Default"
      [1] =>
      string(22) "ValidateTriggerFiction"
    }
  }
}
```

Constraints with 3.2.2
```
array(1) {
  [0] =>
  class Symfony\Component\Validator\Constraints\Regex#78 (6) {
    public $message =>
    string(25) "Please enter a valid ISBN"
    public $pattern =>
    string(9) "/[^\d-]+/"
    public $htmlPattern =>
    NULL
    public $match =>
    bool(false)
    public $payload =>
    NULL
    public $groups =>
    array(2) {
      [0] =>
      string(7) "Default"
      [1] =>
      string(22) "ValidateTriggerFiction"
    }
  }
}
```

# fix
> Potential idea. I will re-investigate after changes in symfony!
[...]

PR #1350, opened 2017-02-19
Validator fix

angelk:
This fix https://github.com/propelorm/Propel2/issues/1347

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "symfony/yaml": "~2.3||~3.0",
         "symfony/console": "~2.3||~3.0",
         "symfony/finder": "~2.3||~3.0",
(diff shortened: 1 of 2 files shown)


gulaandrij:
nice!

composer.json
```diff
@@ -20,7 +20,7 @@
         "symfony/yaml": "~2.3||~3.0",
         "symfony/console": "~2.3||~3.0",
         "symfony/finder": "~2.3||~3.0",
-        "symfony/validator": "~2.3||~3.0 <=3.2.1",
+        "symfony/validator": "~2.3||~3.0.0||~3.1.0||>=3.2.4",
```
gharlan: We should not use `>=`. 
4.0, 5.0, 6.0... matches `>=3.2.4`.

angelk: tnx, fixed

Issue #1355, opened 2017-03-12
Cannot use booleans in ini config files


gboddin:
Hello,

I'm using *master*, and I was trying to configure the following in ```propel.properties``` (or ```propel.ini```) : 

```ini
propel.generator.recursive = true
propel.generator.namespaceAutoPackage = false
propel.generator.packageObjectModel = true
```

but it triggers the following : 
```sh
  [Symfony\Component\Config\Definition\Exception\InvalidTypeException]                    
  Invalid type for path "namespaceAutoPackage". Expected boolean, but got integer.
```

Tested with PHP 5.6.21 and php 7.0.14.

PR #1356, opened 2017-03-12
Fix broken boolean types on IniFileLoader (propelorm/Propel2#1355)

gboddin:
Hello,

I'm using *master*, and I was trying to configure the following in ```propel.properties``` (or ```propel.ini```) : 

```ini
propel.generator.recursive = true
propel.generator.namespaceAutoPackage = false
propel.generator.packageObjectModel = true
```

but it triggers the following : 
```sh
  [Symfony\Component\Config\Definition\Exception\InvalidTypeException]                    
  Invalid type for path "namespaceAutoPackage". Expected boolean, but got integer.
```

Tested with PHP 5.6.21 and php 7.0.14.

### DIFF
--- a/src/Propel/Common/Config/Loader/IniFileLoader.php
+++ b/src/Propel/Common/Config/Loader/IniFileLoader.php
@@ -23,6 +23,7 @@ use Propel\Common\Config\Exception\IniParseException;
  */
 class IniFileLoader extends FileLoader
 {
+
     /**
      * Separator for nesting levels of configuration data identifiers.
      *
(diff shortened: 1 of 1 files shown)


src/Propel/Common/Config/Loader/IniFileLoader.php
```diff
@@ -58,7 +59,7 @@ public function supports($resource, $type = null)
      */
     public function load($file, $type = null)
     {
-        $ini = parse_ini_file($this->getPath($file), true);
+        $ini = parse_ini_file($this->getPath($file), true, INI_SCANNER_RAW);
```
gboddin: It could have been INI_SCANNER_TYPED to avoid additional logic in parseKey() but it would have broke compatibility with PHP 5.5

marcj:
Good catch, thanks! :)

PR #1359, opened 2017-03-29
Fixed Base\Object::toArray when using DateTimeImmutable from configuration

mxcoder:
When using [configuration](http://propelorm.org/documentation/reference/configuration-file.html#datetime-settings) like:
```
      dateTime:
          useDateTimeClass: true
          dateTimeClass: DateTimeImmutable
```
The base objects `toArray` method is unable to process the `DateTimeImmutable`, because its checking directly for `\DateTime`, the current patch makes it use `DateTimeInterface` instead.


marcj:
Thanks, good job!

PR #1360, opened 2017-03-29
Allows the use of Unsigned="true|false" as a MySQL vendor column parameter

TrogloGeek:
(stupid) example:
```xml
        <column name="metricIdStart"      phpName="MetricIdStart" type="integer"            required="true" expose="true">
            <vendor type="mysql">
                <parameter name="Unsigned" value="true"/>
            </vendor>
        </column>
        <column name="metricIdEnd"        phpName="MetricIdEnd"   type="integer"            required="true" expose="true">
            <vendor type="mysql">
                <parameter name="Unsigned" value="false"/>
            </vendor>
        </column>
```

Will yield the following SQL

```mysql
    `metricIdStart` INTEGER UNSIGNED NOT NULL,
    `metricIdEnd` INTEGER NOT NULL,
```

I did no check upon column type when processing Unsigned attribute to keep the code simple, I wonder if I should test it and raise an exception if not applicable to current column type?

### DIFF
--- a/src/Propel/Generator/Platform/MysqlPlatform.php
+++ b/src/Propel/Generator/Platform/MysqlPlatform.php
@@ -366,6 +366,17 @@ DROP TABLE IF EXISTS " . $this->quoteIdentifier($table->getName()) . ";
             $ddl[] = $sqlType;
         }
         $colinfo = $col->getVendorInfoForType($this->getDatabaseType());
+        if ($colinfo->hasParameter('Unsigned')) {
+            $unsigned = $colinfo->getParameter('Unsigned');
+            switch (strtoupper($unsigned)) {
+                case 'FALSE': break;
+                case 'TRUE':
+                    $ddl[] = 'UNSIGNED';
+                    break;
+                default:
+                    throw new EngineException('Unexpected value "'.$unsigned.'" for MySQL vendor column parameter "Unsigned", expecting "true" or "false".');
+            }
+        }
         if ($colinfo->hasParameter('Charset')) {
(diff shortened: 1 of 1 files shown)


src/Propel/Generator/Platform/MysqlPlatform.php
```diff
@@ -366,6 +366,18 @@ public function getColumnDDL(Column $col)
...
+            $unsigned = $colinfo->getParameter('Unsigned');
+            switch (strtoupper($unsigned)) {
+                case 'FALSE': break;
+                case 'TRUE':
+                    $ddl[] = 'UNSIGNED';
+                    break;
+                default:
+                    throw new EngineException('Unexpected value "'.$unsigned.'" for MySQL vendor column parameter "Unsigned", expecting "true" or "false".');
+            }
+            unset($unsigned);
```
marcj: that unset is not necessary :)

TrogloGeek: Yeah sorry, I have the habit of unsetting as soon as I am done with a local var so my IDE stops proposing it to me when autocompleting outside of the codeblock I'm using it.
Would you like me to update the PR without this line?

marcj: Yes please

TrogloGeek: Ok, here it comes

marcj:
Thanks :)

TrogloGeek:
no problem, il allows me not to maintain a fork, so thank you ;-)

PR #1366, opened 2017-04-18
Set timestamp size to 6 to account for high precision timestamps

lmanzke:
The Propel timestampable behavior was recently updated to allow high precision timestamps. In theory, this is good, but unfortunately, MySQL discards the extra precision when saving, causing some problems when you want to compare two objects (one that was "saved" but not "reloaded from database" vs. one that is read from the database). For tests with fixtures, this is quite inconvenient, so I added a size for the timestamp field to make MySQL store the extra precision. 

From https://dev.mysql.com/doc/refman/5.6/en/fractional-seconds.html:

```TIMESTAMP(N) was permitted in old MySQL versions, but N was a display width rather than fractional seconds precision. Support for this behavior was removed in MySQL 5.5.3, so applications that are reasonably up to date should not be subject to this issue. Otherwise, code must be rewritten.```

This means that there might be some side effects here before the mentioned version. Which minimal version does Propel require?

### DIFF
--- a/src/Propel/Generator/Platform/MysqlPlatform.php
+++ b/src/Propel/Generator/Platform/MysqlPlatform.php
@@ -47,7 +47,7 @@ class MysqlPlatform extends DefaultPlatform
         $this->setSchemaDomainMapping(new Domain(PropelTypes::VARBINARY, 'MEDIUMBLOB'));
         $this->setSchemaDomainMapping(new Domain(PropelTypes::LONGVARBINARY, 'LONGBLOB'));
         $this->setSchemaDomainMapping(new Domain(PropelTypes::CLOB, 'LONGTEXT'));
-        $this->setSchemaDomainMapping(new Domain(PropelTypes::TIMESTAMP, 'DATETIME'));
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by lmanzke** [comment]
I see this causes some test failures related to unrecognized syntax. Is there some way to ensure backwards compatibility?

**Comment by marcj** [comment]
It says it fails at

```sql
CREATE TABLE `table3`   
  (                                          
      `id` INTEGER NOT NULL AUTO_INCREMENT,  
      `title` VARCHAR(100),                  
      `test` DATETIME(6),                    
      PRIMARY KEY (`id`)                     
  ) ENGINE=InnoDB
```

but I'm not sure why. Maybe the MySQL version in Travis is too old?

lmanzke:
Not sure, is it older than 5.5.3?

marcj:
I'm not sure. Should work with older version as well, shouldn't it?

marcj:
Well, I can't merge it until it's green in Travis :(

lmanzke:
I know, it's hard, I'll think of something else and close this for now.

PR #1372, opened 2017-05-19
json type column for mysql 5.7

cedriclombardot:


src/Propel/Generator/Builder/Om/ObjectBuilder.php
```diff
@@ -968,6 +968,21 @@ protected function addObjectAccessorBody(&$script, Column $column)
...
+        $this->addDefaultAccessorOpen($script, $column);
+        $this->addJsonAccessorBody($script, $column);
+        $this->addDefaultAccessorClose($script);
+    }
+
+    protected function addJsonAccessorBody(&$script, Column $column)
+    {
+        $clo = $column->getLowercasedName();
+        $script .= "
+        return json_decode(\$this->$clo);";
```
marcj: What if the column is empty or null?

cedriclombardot: Same :

````
php > var_dump(json_decode(json_encode("")));
string(0) ""
php > var_dump(json_decode(json_encode(NULL)));
NULL
`````

marcj: Well, $this->$clo is null and you call thus `json_decode(null);`. I don't see a json_encode in that line.

cedriclombardot: For me it also working 

`````
php > echo json_decode(null);
php > var_dump(json_decode(null));
php shell code:1:
NULL
`````

But if u prefer i can add one if

marcj: Ok, great, then this is not an issue :)

marcj:
Can you also add a little test that covers this feature?

marcj:
Great stuff, thanks :)

PR #1373, opened 2017-05-21
Allowing file loader to accept empty env vars (e.g. DB_PW)

AviBueno:
getenv() returns a string response, which may be an empty string.
This is a valid response and may represent, for example, and empty db password variable (DB_PW).
The patch is to check if getenv() returned false in order to determine if an error occurred or not.
* The same check is used in [TestCaseFixtures.php (line 101)](https://github.com/propelorm/Propel2/blob/618c192272b5bc9636412659bf32bd572910b819/tests/Propel/Tests/TestCaseFixtures.php#L101)

### DIFF
--- a/src/Propel/Common/Config/Loader/FileLoader.php
+++ b/src/Propel/Common/Config/Loader/FileLoader.php
@@ -320,7 +320,7 @@ abstract class FileLoader extends BaseFileLoader
         $env = explode('.', $value);
         if ('env' === $env[0]) {
             $envParam = getenv($env[1]);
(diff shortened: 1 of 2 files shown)


AviBueno:
Hi, the 2nd commit (unit test) failed on a single Travis test. Not sure if it's the code or Travis.

marcj:
I restarted it. Travis issue again. Green now :)

AviBueno:
Great, thanks!

marcj:
Alright, thanks :)

PR #1375, opened 2017-05-26
MSSQL `applyLimit` Greedy Regex Fix

McManning:
Fixes an issue with the `applyLimit` in the MssqlAdapter incorrectly splitting the query string when inner select statements are included. 

Example use case:
```php
$projects = ProjectQuery::create()
    ->where('Project.ProjectID IN (
      SELECT InstanceID FROM RBAC.dbo.InstanceTemp
    )')
    ->offset(10)
    ->limit(10)
    ->find();
```


marcj:
Nice, thanks!

Issue #1381, opened 2017-05-30
MSSQL `applyLimit` Regex patch breaking virtual columns


McManning:
I seemed to have been a bit gung-ho about resolving #1376 

While it did fix the case of the regex not parsing out the appropriate WHERE clause, it introduces a bug for virtual columns that are based on select statements. 

E.g. the following use case generates a malformed SQL statement (Same base sample from #1376):
```php
$projects = ProjectQuery::create()
    ->withColumn('(SELECT Email FROM User WHERE ID = Project.OwnerID)', 'Email')
    ->offset(10)
    ->limit(10)
    ->find();
```

In this case, the `FROM` in the virtual column is the split point in `MssqlAdapter::applyLimit`. 

If you want me to do a PR to reverse PR #1375 in the meantime while I iron out these new issues, I can do that. 

Thanks,
Chase

McManning:
Update on this; I have a potential patch ready to go. I'm adding a handful of regression tests, just to give myself (and probably others) peace of mind on this one. 

I'm seeing that there isn't a `MssqlAdapterTest` class present - would it be fine if I added one alongside the other Pdo adapter tests? It seems like you guys aren't doing active MSSQL testing in your CI environment, but the tests I want to write wouldn't need an active connection anyway. They'd be very similar to `OracleAdapterTest`.

marcj:
Yeah sure, go for it. :)

dereuromark:
PR seems merged.

PR #1382, opened 2017-05-31
Fix detection of subquery virtual columns in MssqlAdapter::applyLimit

McManning:
Corrections for #1381, as well as a new regression test class `MssqlAdapterTest` to cover them. 

I'm not sure if temporarily changing the default adapter for the tests was the right approach, but I couldn't find another way to effectively test using `ModelCriteria`. I'm open to alternative suggestions.

### DIFF
--- a/src/Propel/Runtime/Adapter/Pdo/MssqlAdapter.php
+++ b/src/Propel/Runtime/Adapter/Pdo/MssqlAdapter.php
@@ -140,15 +140,31 @@ class MssqlAdapter extends PdoAdapter implements SqlAdapterInterface
         }
 
         // split the select and from clauses out of the original query
-        $selectSegment = [];
-
+        $selectStatement = '';
+        $fromStatement = '';
         $selectText = 'SELECT ';
+        $selectTextLen = strlen($selectText);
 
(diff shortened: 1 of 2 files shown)


marcj:
Great, thanks!

chris-chr:
i think the changes in MssqlAdapter class broke queries where these is a column name that contains the word from.
e.g
```
SELECT TOP 1 repair.id, repair.date_from, repair.date_to
FROM repair WHERE repair.id=:p1

SELECT TOP 1 repair.id, repair.date_ FROM , repair.date_to
FROM repair WHERE repair.id=:p1
```

marcj:
@chris-chr can you provide a fix or should we revert?

McManning:
If it's not incredibly urgent, I can be available later today to take a look at doing another PR on top of this. I assume it might be related to that `stripos($sql, 'from', $i)` line in there.

marcj:
that would be awesome :)

McManning:
Opened a PR that should correct this. @chris-chr if you could double check this against your use case, that'd be super helpful. Thanks!

chris-chr:
Just checked, it works properly. Thanks!

marcj:
Wow, that was fast! :)

Issue #1345, opened 2017-02-10
Corrupt data returned: many-to-many without primary key


ronisaacson:
Consider a many-to-many relationship with a mapping table like this:

```
create table customer_order (
  customer_id int not null,
  order_id int not null,
  foreign key (customer_id) references customers(id),
  foreign key (order_id) references orders(id)
  -- note: no primary key
);
```

If you call $customer->getOrders() and the customer has 5 orders, you'll get a collection with 5 objects. But all 5 objects will be the same, and they'll all be for the first order! This is because CustomerOrderTableMap::getPrimaryKeyHashFromRow() always returns null. So it fetches 5 rows... populateObject() turns the first into an object, caches it with a key of null, then turns the next 4 rows into the exact same object.

The fix is to add a primary key to this table. But it should still return the correct results (or at least throw an error) if you don't have one.

lcsdms:
I had the same problem before, it's referenced on this issue : #797

dereuromark:
Is someone able to make a PR here with suggested changes?
Or did those referenced commits reach master?

PR #1387, opened 2017-06-05
Fix issue #1345: Corrupt data returned: many-to-many without primary key

AlexanderBliznuk:
InstancePoolTrait::addInstanceToPool didn't check if the primary key of instance was empty string. And since addInstanceToPool assigned each of instances without primary constraint to empty string key, InstancePoolTrait::getInstanceFromPool could return only the last written one.


marcj:
Good find! Unfortunately it broke some unit tests. I re-run the build to make sure it wasn't a mistake by travis. Let's see.

Issue #1388, opened 2017-06-12
commit #1259 breaks model-only relationships


AlexanderBliznuk:
skipSql="true" on foreign-key is being ignored

PR #1389, opened 2017-06-12
Fix issue #1388: broken model-only relationships

Alexander-Bliznyuk:


tests/Propel/Tests/Issues/Issue1388Test.php
```diff
@@ -0,0 +1,58 @@
...
+        $builder->setSchema($schema);
+
+        try {
+            $builder->buildClasses(null, true);
+            $noExceptionThrown = true;
+        } catch (\Exception $e) {
+            $noExceptionThrown = false;
+        }
+
+        $this->assertTrue($noExceptionThrown);
```
marcj: When a exception is thrown this test is automatically marked as failed, so no need for that assertTrue :)

src/Propel/Generator/Model/Table.php
```diff
@@ -791,7 +791,7 @@ public function setupReferrers($throwErrors = false)
                 $foreignPrimaryKeys = $foreignTable->getPrimaryKey();
                 // check all keys are referenced in foreign key
                 foreach ($foreignPrimaryKeys as $foreignPrimaryKey) {
-                    if (!$foreignPrimaryKey->hasReferrer($foreignKey) && $throwErrors) {
+                    if (!$foreignPrimaryKey->hasReferrer($foreignKey) && !$foreignKey->isSkipSql() && $throwErrors) {
```
marcj: well, generating SQL has nothing to do with checking for valid foreign keys. So this check is a misuse of the `skip-sql` attribute to allow models generation with invalid foreign keys.

marcj (changes requested):

src/Propel/Generator/Model/Table.php
AlexanderBliznuk: Propel used to allow flexible models tailored for application's needs. This checking for foreign keys merely ties hands. For example, if referent table has composite key, it appears to be impossible to join it by a column outside the key.
Please also consider fully justified and common scenario: i18n tables having primary key(id, locale). Of course in many cases other tables would refer to i18n tables only by id. Another similar common scenario is versioning of entities. 
Maybe it's ok to be strict and thus guarantee scheme validity between different RDBMS and to guide amateurs.
But  [model only relationships](http://propelorm.org/documentation/04-relationships.html#model-only-relationships) feature shouldn't be broken. It doesn't work now as described in docs and my pull request fixes this regression.

gechetspr:

PR #1395, opened 2017-07-15
Fix MssqlAdapter::applyLimit from generating malformed queries when `from` is included as a non-keyword

McManning:
Fixes regression issue found after PR #1382 was merged in. Now correctly identifies `FROM` as a keyword versus `FROM` as part of a column name.


marcj:
that was fast :) good job, thanks!

PR #1396, opened 2017-07-17
Fix json equality check

kriks57:
Previously, when using JSON Type for a column, every setXXX() on this column would result in a modified column.
This PR adds a check between old JSON and new value, as objects.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,8 +43,7 @@
     ],
     "extra": {
         "branch-alias": {
-            "dev-master": "2.0-dev",
-            "dev-fix-json-equality": "2.0-dev"
+            "dev-master": "2.0-dev"
         }
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,8 @@
     ],
     "extra": {
         "branch-alias": {
-            "dev-master": "2.0-dev"
+            "dev-master": "2.0-dev",
+            "dev-fix-json-equality": "2.0-dev"
         }
     }
 }
--- a/src/Propel/Generator/Builder/Om/ObjectBuilder.php
+++ b/src/Propel/Generator/Builder/Om/ObjectBuilder.php
@@ -170,7 +170,7 @@ class ObjectBuilder extends AbstractObjectBuilder
             $fmt = $this->getTemporalFormatter($column);
             try {
                 if (!($this->getPlatform() instanceof MysqlPlatform &&
-                ($val === '0000-00-00 00:00:00' || $val === '0000-00-00'))) {
+                    ($val === '0000-00-00 00:00:00' || $val === '0000-00-00'))) {
                     // while technically this is not a default value of NULL,
                     // this seems to be closest in meaning.
                     $defDt = new \DateTime($val);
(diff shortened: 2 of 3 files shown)


src/Propel/Generator/Builder/Om/ObjectBuilder.php
```diff
@@ -999,11 +1001,32 @@ public function addJsonAccessorComment(&$script, Column $column)
...
+     * @param string &$script
+     * @param Column $column
+     */
+    public function addJsonAccessorOpen(&$script, Column $column)
+    {
+        $cfc = $column->getPhpName();
+        $visibility = $column->getAccessorVisibility();
+
+        $script .= "
+    ".$visibility." function get$cfc(\$asArray = false";
```
marcj: I'd go per default asArray=true

kriks57: I would too, but it would change previous behavior

marcj: do it :)

kriks57: BTW, it's done :)

src/Propel/Generator/Builder/Om/ObjectBuilder.php
```diff
@@ -1804,11 +1881,19 @@ protected function addJsonMutator(&$script, Column $col)
...
         $clo = $col->getLowercasedName();
 
-        $this->addMutatorOpen($script, $col);
+        $this->addJsonMutatorOpen($script, $col);
 
         $script .= "
-        \$this->$clo = json_encode(\$v);
-        \$this->modifiedColumns[".$this->getColumnConstant($col)."] = true;
+        if (is_array(\$v)) {
+            \$v = json_decode(json_encode(\$v));
```
marcj: what purpose has this line?

composer.json
```diff
@@ -43,7 +43,8 @@
     ],
     "extra": {
         "branch-alias": {
-            "dev-master": "2.0-dev"
+            "dev-master": "2.0-dev",
+            "dev-fix-json-equality": "2.0-dev"
```
marcj: we can remove that again

src/Propel/Generator/Builder/Om/ObjectBuilder.php
kriks57: transform array to object, for camparison

composer.json
kriks57: just before merging, we currently need it :)

src/Propel/Generator/Builder/Om/ObjectBuilder.php
staabm: couldnt you just cast it to `(object)`?

kriks57: it wouldn't work recursively

marcj: I don't understand. Why don't compare the actual object, but the json-string of it. Where is a object comparison happen?

kriks57: 5 lines below :)
`if (\$v != json_decode(\$this->$clo)) {`

marcj: Then we should change that to `if (json_encode($v) != $this->$clo)`, it's safer and probably faster.

kriks57: Let's try

kriks57: done, what do you think ?

kriks57:
Could we merge ?

composer.json
marcj: Then remove it ;P

kriks57: done :)

kriks57:
Now, can we? :)

marcj:
Alrighty, thanks!

PR #1397, opened 2017-07-26
Fix getPrimaryKeyFromRow for custom php types

### LINKED ISSUE #926: Typecast on Om/TableMapBuilder.php#L849 causes issues with user defined phpType
## Synopsis

When using a custom userspace column php datatype, the generator creates a typecast that doesn't work.
### Example schema entry

```
<table name="_operator" idMethod="native" phpName="Operator">
    <column name="id" phpName="Id" phpType="\api\utility\ShaId" type="VARCHAR" size="20" sqlType="binary(20)" primaryKey="true" required="true"/>
    <column name="name" phpName="Name" type="VARCHAR" size="120" require="true"  />
</table>
```
### Yields (OperatorTableMap.php ~@L189)

```
    return (\api\utility\ShaId) $row[  // <-- This part. 
        $indexType == TableMap::TYPE_NUM
            ? 0 + $offset
            : self::translateFieldName('Id', TableMap::TYPE_PHPNAME, $indexType)
    ];
```

jaspervdm:
In the generated TableMap, the `getPrimaryKeyFromRow` method gave an error when using a custom `phpType`. This is fixed by only casting to the php type if its not an custom object type. In that case, use the contstructor. Fixes #926

### DIFF
--- a/src/Propel/Generator/Builder/Om/TableMapBuilder.php
+++ b/src/Propel/Generator/Builder/Om/TableMapBuilder.php
@@ -847,11 +847,11 @@ class ".$this->getUnqualifiedClassName()." extends TableMap
                 if (!$col->isLazyLoad()) {
                     if ($col->isPrimaryKey()) {
                         $script .= '
-        $pks[] = (' . $col->getPhpType() . ') ' . "\$row[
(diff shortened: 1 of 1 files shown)


marcj:
Cool, thanks!

Issue #1406, opened 2017-08-25
propel diff issue with CURRENT_TIMESTAMP on maria 10.2


wollanup:
## Expected Behavior
`propel diff` should not generate diffs when nothing has changed

## Current Behavior
`propel diff` find diff with `TIMESTAMP` columns with `defaultExpr="CURRENT_TIMESTAMP"`
**From what I can remember, issue started with mariadb 10.2, and was not present in 10.1**

## Steps to Reproduce
1. change/create schema
2. run `propel diff`
3. run `propel migrate`
4. run `propel diff` **(should not not find diff)**

## Detailed Description
Given this XML schema :
```
<?xml version="1.0" encoding="utf-8"?>
<database name="test_date_db">
	<table name="test_date">
		<column name="col_date" type="TIMESTAMP" defaultExpr="CURRENT_TIMESTAMP"/>
	</table>
</database>
```
I run `propel diff` which generate CREATE code :
```
CREATE TABLE `test_date`
(
    `col_date` TIMESTAMP NULL DEFAULT CURRENT_TIMESTAMP
) ENGINE=InnoDB;
```
`propel migration` creates the table.

`mysqldump` give me this result :
```
CREATE TABLE `test_date` (
  `col_date` timestamp NULL DEFAULT current_timestamp()
) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```

If a run `propel diff` again, I get this :
```
ALTER TABLE `test_date`

  CHANGE `col_date` `col_date` TIMESTAMP NULL DEFAULT CURRENT_TIMESTAMP;
```

## Environment
Application runs in a Debian docker container with :
- php 7.0
- mariadb 10.2.8
- propel 2 (6c3d6364fd237231519298911055292107a8b3f1)

wollanup:
After some investigations on Diff code, in [ColumnComparator.php](https://github.com/propelorm/Propel2/blob/master/src/Propel/Generator/Model/Diff/ColumnComparator.php#L75) we have :

**$fromDefaultValue**
```
NULL
```
**$toDefaultValue**
```
class Propel\Generator\Model\ColumnDefaultValue#716 (2) {
  private $value =>
  string(17) "CURRENT_TIMESTAMP"
  private $type =>
  string(4) "expr"
}
```

I wil try to understand how Domain object is populated as we use it in Column::getDefaultValue() ...

marcj:
[...]

PR #1407, opened 2017-08-25
Fix issue #1406 : diff issue with CURRENT_TIMESTAMP on maria 10.2

wollanup:
In mariadb 10.2 `CURRENT_TIMESTAMP` is replaced by`current_timestamp()` in `SHOW COLUMNS FROM my_table`


marcj:
Good catch, thanks!

PR #1425, opened 2017-10-23
address countable issue in php7.2

### LINKED ISSUE #1424: upcoming: php 7.2 countable issue
All:

Looking into php7.2 and seeing some upcoming warning(s) that will need to be addressed due to the countable updates - I'll try to look into creating a PR for it but wanted to get it into the issues bucket:

Warning: count(): Parameter must be an array or an object that implements Countable in src/Propel/Runtime/ActiveQuery/InstancePoolTrait.php on line 33

Thanks!

cbschuld:
address countable issue in php7.2 and update if/else structure for unnecessary else

fixes #1424

**NOTE:** _I am not new to using propel but new to the internals of her; it will need someone with some local knowledge_

### DIFF
--- a/src/Propel/Runtime/ActiveQuery/InstancePoolTrait.php
+++ b/src/Propel/Runtime/ActiveQuery/InstancePoolTrait.php
@@ -30,12 +30,14 @@ trait InstancePoolTrait
     public static function getInstanceKey($value)
     {
         if (!($value instanceof Criteria) && is_object($value)) {
-            if (count($pk = $value->getPrimaryKey()) > 1 || is_object($value->getPrimaryKey())) {
+            $pk = $value->getPrimaryKey();
+            if(is_object($pk) || (is_numeric($pk) && $pk > 1) || (is_array($pk) && count($pk) > 1)) {
                 $pk = serialize($pk);
             }
-
             return (string) $pk;
-        } elseif (is_scalar($value)) {
+        }
+
+        else if (is_scalar($value)) {
             // assume we've been passed a primary key
             return (string) $value;
         }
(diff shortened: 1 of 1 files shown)


src/Propel/Runtime/ActiveQuery/InstancePoolTrait.php
```diff
@@ -30,12 +30,14 @@ public static function addInstanceToPool($object, $key = null)
     public static function getInstanceKey($value)
     {
         if (!($value instanceof Criteria) && is_object($value)) {
-            if (count($pk = $value->getPrimaryKey()) > 1 || is_object($value->getPrimaryKey())) {
+            $pk = $value->getPrimaryKey();
```
denormalizer: Sorry, I'm just snooping around. Should:

    (is_numeric($pk) && $pk > 1)

be

    (is_numeric($pk) && count($pk) > 1)

cbschuld: @denormalizer - a numeric value won't be "countable" which is the source of the issue; since we know it is numeric on the conditional it would just be 

```(is_numeric($pk) && $pk > 1)```

denormalizer: True. I don't know what I was thinking :)

miraclebg:
Why not simply:

` $pk = $value->getPrimaryKey();

            if (($pk && (is_array($pk) || ($pk instanceof Countable)) && count($pk) > 1) || is_object($value->getPrimaryKey())) {
                $pk = serialize($pk);
            }`

miraclebg:
Well - the is_array check here is clearly not necessary but still :)

src/Propel/Runtime/ActiveQuery/InstancePoolTrait.php
guoshaoxia: Thanks that worked well!

marcj:
Thanks!

dereuromark:
@marcj 
We need this released for our framework to be 7.2+ compatible. Thank you.

cristianoc72:
@dereuromark this commit contains a regression, please wait #1452

PR #1434, opened 2017-12-08
Symfony 4 compatibility

cristianoc72:
Symfony 4 compatibility, useful for https://github.com/propelorm/PropelBundle/pull/476
Since we never use it, I've removed Behat, because it causes composer failures, while switching from Symfony components 2.x to 3.4+


gharlan:
I would like to use Propel 2 with SF4, so would be happy to get this merged.

marcj:
Thanks!

PR #1439, opened 2017-12-29
Added referenceOnly attribute to external-schema

JonasRudolph:
I use this xsd file in my IDE and it always complains, that the referenceOnly-attribute 'is not allowed here', even though it is [0]. Thats why I would like this to be merged :)

[0]: http://propelorm.org/documentation/reference/schema.html#external-schema-element


marcj:
Great thanks!

PR #1446, opened 2018-01-15
Make offsetGet compatible with the parent's offsetGet method

timoschwarzer:
This fixes fatal errors with PHP 7.2.


marcj:
Thanks!

PR #1449, opened 2018-01-17
Address failed tests after applying fix for issue #1425

Gomes81:
Address failed tests after applying fix for issue #1425 which is ralated to the countable issue in php 7.2.

I think the issue is related to the fact that the fix added the condition: is_numeric($pk) && $pk > 1 to the if, which is not present in the previous releases, which then caused some tests to start failing.

### DIFF
--- a/src/Propel/Runtime/ActiveQuery/InstancePoolTrait.php
+++ b/src/Propel/Runtime/ActiveQuery/InstancePoolTrait.php
@@ -31,9 +31,11 @@ trait InstancePoolTrait
     {
         if (!($value instanceof Criteria) && is_object($value)) {
             $pk = $value->getPrimaryKey();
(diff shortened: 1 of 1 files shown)


dereuromark:
The php7.2+ compatibility becomes a bit of a critical issue lately as a lot of service providers are rolling out already 7.2 as their default PHP env, and people expect PHP applications to be running on those just fine.

Gomes81:
It seems that the issue was already solved by another commit and this fix is not needed anymore... However, I still prefer this implementation to the one currently implemented.

dereuromark:
What commit/fix do you refer to? You got the example to compare to your solution?

Gomes81:
Sorry for not mention it @dereuromark, it was by this pull request: #1452 (which in fact did such a good job, fixing all the issues with tests in 7.2)

marcj:
Thanks!

Issue #1447, opened 2018-01-16
TimestampableBehavior PropelDateTime::createHighPrecision


eugenekurasov:
composer show | grep propel
```
propel/propel                      2.0.0-alpha6 Propel2 is an open-source Object-Relational Mapping (ORM) for PHP 5.5 and up.
propel/propel-bundle               3.0.0-alpha1 Integration of Propel in Symfony2
```

\Propel\Generator\Behavior\Timestampable\TimestampableBehavior::preInsert
Update time is different then created time on sqlite

PR #1457, opened 2018-01-28
 #1447. On preInsert object in TimestampableBehavior we get different…

eugenekurasov:
… dataTime.

For solve this situation we need available set same time on create and update field.

### DIFF
--- a/src/Propel/Generator/Behavior/Timestampable/TimestampableBehavior.php
+++ b/src/Propel/Generator/Behavior/Timestampable/TimestampableBehavior.php
@@ -102,12 +102,13 @@ class TimestampableBehavior extends Behavior
      */
     public function preInsert($builder)
     {
-        $script = '';
+        $script = '$time = time();
+$highPrecision = \\Propel\\Runtime\\Util\\PropelDateTime::createHighPrecision();';
 
         if ($this->withCreatedAt()) {
             $valueSource = strtoupper($this->getTable()->getColumn($this->getParameter('create_column'))->getType()) === 'INTEGER'
-                ? 'time()'
-                : '\\Propel\\Runtime\\Util\\PropelDateTime::createHighPrecision()'
+                ? '$time'
+                : '$highPrecision'
             ;
             $script .= "
 if (!\$this->isColumnModified(" . $this->getColumnConstant('create_column', $builder) . ")) {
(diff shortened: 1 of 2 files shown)


src/Propel/Runtime/Util/PropelDateTime.php
```diff
@@ -75,6 +80,23 @@ public static function createHighPrecision($time = null)
         return $dateTime;
     }
 
+    /**
+     * Get a DateTime object with milliseconds resolution.
+     *
+     * @param null $time
```
dereuromark: whats the alternative value? should be stated here, too.

eugenekurasov:
Yeah, tests passed

eugenekurasov:
@marcj can you merge this PR?
Fix for #1447

src/Propel/Generator/Behavior/Timestampable/TimestampableBehavior.php
```diff
@@ -85,7 +85,7 @@ public function preUpdate($builder)
         if ($this->withUpdatedAt()) {
             $valueSource = strtoupper($this->getTable()->getColumn($this->getParameter('update_column'))->getType()) === 'INTEGER'
                 ? 'time()'
-                : '\\Propel\\Runtime\\Util\\PropelDateTime::createHighPrecision()'
+                : '\\Propel\\Runtime\\Util\\PropelDateTime::createSameHighPrecision(1)'
```
marcj: createSameHighPrecision has as first argument a bool, not a integer.

src/Propel/Generator/Behavior/Timestampable/TimestampableBehavior.php
```diff
@@ -105,20 +105,20 @@ public function preInsert($builder)
         $script = '';
 
         if ($this->withCreatedAt()) {
-            $valueSource = strtoupper($this->getTable()->getColumn($this->getParameter('create_column'))->getType()) === 'INTEGER'
+            $createdValueSource = strtoupper($this->getTable()->getColumn($this->getParameter('create_column'))->getType()) === 'INTEGER'
                 ? 'time()'
-                : '\\Propel\\Runtime\\Util\\PropelDateTime::createHighPrecision()'
+                : '\\Propel\\Runtime\\Util\\PropelDateTime::createSameHighPrecision(1)'
```
marcj: same here

marcj:
Can you talk a bit about that solution and why you've chosen to use a static global stored timestamp in the PropelDateTime class? I don't think that is very elegant, maybe you can show me what the actual error is so we can find a more sexy solution.

eugenekurasov:
Hi,
I try found is more easy solution for preInsert action.
So I need to have same value for insert database on preInsert.
createHighPrecision - give me different values each time, and this is like object is was modify, but is not true.
So I need HighPrecision with save state for Insert and update. And I think **private static** is transparent solution and is small evil.

Other solution - is change generated script with additional conditions for check if created date is exist and update exist - generate once time.
But of this solution I need to check what fields type, and if they different I need to convert value.  I think it is incomprehensible solution solution.

If you have any idea how I can do it for easy understand and without static - I will be try do it.

I will be change number to boolean.

eugenekurasov:
@marcj I set boolean type instead integer.
For preUpdate method I roll back change.

eugenekurasov:
@marcj I have found solution without using static variable.
Can you check PR again please.

marcj:
That looks way better, thanks!

PR #1464, opened 2018-02-16
SQL injection fix: Coerce offset and limit values to integers for MySQL LIMIT clause

### LINKED ISSUE #1463: SQL injection possible with limit() on MySQL
**The `limit()` query method is susceptible to catastrophic SQL injection with MySQL.**

For example, given a model `User` for a table `users`:
```php
UserQuery::create()->limit('1;DROP TABLE users')->find();
```
This will drop the `users` table!

The cause appears to be a lack of integer casting of the limit input in either [`Propel\Runtime\ActiveQuery\Criteria::setLimit()`](https://github.com/propelorm/Propel2/blob/master/src/Propel/Runtime/ActiveQuery/Criteria.php#L1260-L1266) or in [`Propel\Runtime\Adapter\Pdo\MysqlAdapter::applyLimit()`](https://github.com/propelorm/Propel2/blob/master/src/Propel/Runtime/Adapter/Pdo/MysqlAdapter.php#L122-L129). The code comments there seem to imply that casting was avoided due to overflow issues with 32-bit integers.

This is surprising behavior since one of the primary purposes of an ORM is to prevent basic SQL injection.

**This affects all versions of Propel: 1.x, 2.x, and 3.**

mpetrovich:
When constructing a MySQL LIMIT clause, values for the offset and limit are coerced to integers. This prevents arbitrary SQL from being injected via a query limit. Example:
```php
UserQuery::create()->limit('1;DROP TABLE users')->find();
```
Previously, this would have injected `DROP TABLE users` into the generated SQL. Now, the limit value would be coerced to the integer `1`.

Fixes #1463

### DIFF
--- a/src/Propel/Runtime/Adapter/Pdo/MysqlAdapter.php
+++ b/src/Propel/Runtime/Adapter/Pdo/MysqlAdapter.php
@@ -121,6 +121,9 @@ class MysqlAdapter extends PdoAdapter implements SqlAdapterInterface
      */
     public function applyLimit(&$sql, $offset, $limit)
     {
+        $offset = (int) $offset;
+        $limit = (int) $limit;
+
         if ($limit >= 0) {
             $sql .= ' LIMIT ' . ($offset > 0 ? $offset . ', ' : '') . $limit;
         } elseif ($offset > 0) {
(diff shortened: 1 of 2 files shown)


mpetrovich:
If these changes look good, I'll port them to Propel versions 1 and 3 as well.

marcj:
Perfect, thanks!

mpetrovich:
Followup PR to prevent similar vulnerabilities in other database adapters: https://github.com/propelorm/Propel2/pull/1465

PR #1465, opened 2018-02-16
SQL injection fix: Cast limit to integer when setting via Criteria::setLimit()

mpetrovich:
This is a followup to a [fix for SQL injections with LIMIT clauses in MySQL](https://github.com/propelorm/Propel2/pull/1464). That fix only applied to the MySQL adapter, and other existing or future adapters could still be at risk.

By coercing limit inputs to integers upon setting them, we can avoid SQL injection vulnerabilities with `limit()` across all database adapters.

The original code comments implied that integer coercion could be problematic with 32-bit integers, but unit tests in this PR prove otherwise. Even 64-bit integers seem to work fine.

### DIFF
--- a/src/Propel/Runtime/ActiveQuery/Criteria.php
+++ b/src/Propel/Runtime/ActiveQuery/Criteria.php
@@ -1259,8 +1259,7 @@ class Criteria
      */
     public function setLimit($limit)
     {
-        // TODO: do we enforce int here? 32bit issue if we do
-        $this->limit = $limit;
+        $this->limit = (int) $limit;
 
         return $this;
     }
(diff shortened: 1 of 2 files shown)


src/Propel/Runtime/ActiveQuery/Criteria.php
```diff
@@ -1259,8 +1259,7 @@ public function isSingleRecord()
      */
     public function setLimit($limit)
     {
-        // TODO: do we enforce int here? 32bit issue if we do
-        $this->limit = $limit;
+        $this->limit = (int) $limit;
```
mpetrovich: _Maintainers:_
Can anyone provide some context around this comment? My unit tests show that no truncation occurs for large values like 64-bit integers.

marcj: looks fine to me.

Issue #1471, opened 2018-03-17
ObjectCollection::toArray() will not pass on $includeForeignObjects parameter to objects


ioannis-papikas:
Although each generated object that implements `ActiveRecordInterface` has a parameter `$includeForeignObjects` to include any foreign objects, `ObjectCollection` does not make use of this parameter and passes on always `true`.

The issue is here: `\Propel\Runtime\Collection\ObjectCollection::toArray($keyColumn = null, $usePrefix = false, $keyType = TableMap::TYPE_PHPNAME, $includeLazyLoadColumns = true, $alreadyDumpedObjects = [])`

dereuromark:
PR open

PR #1472, opened 2018-03-17
Add $includeForeignObjects in ObjectCollection::toArray() parameters

ioannis-papikas:
This Pull Request solves the issue #1471


dereuromark:
Can you also rebase on top of latest master?
Then we should be able to proceed.

Incognito (changes requested):
Added feature should get new tests.

PR #1484, opened 2018-06-23
Archivable behavior with FK customization.

ppatrik:
Added two boolan parameters:
* **archive_foreign_keys** (default: false) - on true FK will be copied to Archive table class, so API for archive and non-archive class is the same.
* **archive_foreign_keys_skip_sql** (default: true) - same API is enought so database FK will be skiped by default.

### DIFF
--- a/src/Propel/Generator/Behavior/Archivable/ArchivableBehavior.php
+++ b/src/Propel/Generator/Behavior/Archivable/ArchivableBehavior.php
@@ -28,6 +28,8 @@ class ArchivableBehavior extends Behavior
         'archive_table'       => '',
         'archive_phpname'     => null,
         'archive_class'       => '',
+        'archive_foreign_keys' => 'false',
+        'archive_foreign_keys_skip_sql' => 'true',
         'log_archived_at'     => 'true',
         'archived_at_column'  => 'archived_at',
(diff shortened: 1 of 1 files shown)


dereuromark:
@ppatrik or others: Any update on this?

Incognito:
@dereuromark It's a useful improvement but I don't think ppatrik is coming back to add a test after 4 years. Also, I'm not sure if anyone is actively using Archive functionality in the wild or not. If you think end-users aren't using it: maybe close it with a flag "icebox" or something to indicate it can be pulled back if there is a need for it.

mringler:
@Incognito I think Archive is used, at least there was a question on the issue tracker from someone using it: [#1919#issuecomment-1340073429](https://github.com/propelorm/Propel2/issues/1919#issuecomment-1340073429)
I think the problem even was about missing FKs

Incognito:
Sounds relevant IMO, that issue could probably be fixed with this MR. I'm no longer a maintainer but I know at the time the code LGTM in concept, but it was missing a test to ensure it doesn't break in the future.

Migrating from old sf components to modern ones was a massive pain and comprehensive tests was the only way it was possible without breaking things.

PR #1491, opened 2018-08-07
Fix relative path for external-schema

jzaplet:


src/Propel/Generator/Manager/AbstractManager.php
```diff
@@ -364,6 +364,12 @@ protected function includeExternalSchemas(\DOMDocument $dom, $srcDir)
 
             $externalSchema->parentNode->removeChild($externalSchema);
 
+            $relativePath = realpath($srcDir. '/'. $include);
+
+            if (!file_exists($include) && file_exists($relativePath)) {
```
marcj: `!file_exists($include)` should be removed tho

jzaplet: Can I use is_readable function in this condition?

jzaplet: Or do you think I should remove the first part of the condition?

marcj: `if (file_exists($relativePath)) {` should be enough

dereuromark:
Is there any test case for this maybe?

jzaplet:
> Is there any test case for this maybe?

If you mean test case of "relative path for external schema".

There is example https://github.com/wakerscz/cms-sandbox/blob/master/app/schema/schema.xml

dereuromark:
Yes. A unit test here would prove that the change is valid and also protect it from accidental regressions in the future.

jzaplet:
But I think, Propel ORM is no longer maintained.

dereuromark:
You can check the recent history and you will see that is very much maintained again. Mainly by Spryker and friends.

dereuromark:
Would be awesome if you also rebased on top of latest master.
As we have now a very strong test suite running for all kind of versions and DBs.

mringler:
Hi @jzaplet,

found this PR when I was struggling with something around external schemas. I am happy to help you write a test case, if you are still interested, and find that this is a valuable addition to Propel.
Looking at the code, I have some questions, maybe we can sort them out and get this going?

src/Propel/Generator/Manager/AbstractManager.php
```diff
@@ -364,6 +364,12 @@ protected function includeExternalSchemas(\DOMDocument $dom, $srcDir)
 
             $externalSchema->parentNode->removeChild($externalSchema);
 
+            $relativePath = realpath($srcDir. '/'. $include);
```
dereuromark, line 367: can we use DS constant (DIR... SEP) instead of / maybe?

mringler:
Added a test and some fixes in #1722

PR #1492, opened 2018-08-09
 Fix path to auto-generated Base classes (replaced dot before .Base)

jzaplet:
For using package attribute in xml schema


dereuromark:
Is there any test case for this maybe?

gechetspr:
@jzaplet We are closing this PR due to the fact that there is no movement for couple of years already.

Please re-open it if the issue is still valid

PR #1507, opened 2019-02-13
Add column and table descriptions to MySQL reverse-engineered schemas

ikeyan:
```XML
<table name="test" ... description="Test table">
  <column name="id" ... description="The ID">
</table>
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,6 +23,7 @@
         "symfony/validator": "~2.3||~3.0.0||~3.1.0||^3.2.4||~4.0",
         "symfony/filesystem": "~2.3||~3.0||~4.0",
         "symfony/config": "~2.3||~3.0||~4.0",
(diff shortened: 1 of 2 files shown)


dereuromark:
This requires a rebase now.
Is it possible to add a test case along with the changes you propose? To outline why this changes?

composer.json
```diff
@@ -23,6 +23,7 @@
         "symfony/validator": "~2.3||~3.0.0||~3.1.0||^3.2.4||~4.0",
         "symfony/filesystem": "~2.3||~3.0||~4.0",
         "symfony/config": "~2.3||~3.0||~4.0",
+        "symfony/translation": "~2.3||~3.0||~4.0",
```
Incognito, line 26: We've accomplished this in the latest build so it can be removed from the changeset after this work is rebased.

PR #1508, opened 2019-02-14
faster SqlParser

ikeyan:


dereuromark:
This requires a rebase now.
Please try to keep the changes isolated per topic
If you also want to fix up other topics, those should be separate PRs for quicker review and being able to merge them.

PR #1513, opened 2019-05-16
Fixed relative template path assumption

oojacoboo:
As this is currently written, it's difficult to write behaviors and specify a template dir since it's using the __DIR__ constant, effectively making it relative.  The change keeps the same defaults, but expects a full absolute path, as it should be.

### DIFF
--- a/src/Propel/Generator/Builder/Om/AbstractOMBuilder.php
+++ b/src/Propel/Generator/Builder/Om/AbstractOMBuilder.php
@@ -992,12 +992,17 @@ abstract class AbstractOMBuilder extends DataModelBuilder
      *
      * @param  string $filename
      * @param  array  $vars
-     * @param  string $templateDir
+     * @param  string|null $templateDir
      * @return string
      */
-    public function renderTemplate($filename, $vars = [], $templateDir = '/templates/')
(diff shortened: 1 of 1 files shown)


dereuromark:
Is there a test case along with those changes maybe?
Please also rebase as current master is green.

src/Propel/Generator/Builder/Om/AbstractOMBuilder.php
```diff
@@ -992,12 +992,17 @@ public function getBehaviorContentBase($contentName, $modifier)
...
      * @return string
      */
-    public function renderTemplate($filename, $vars = [], $templateDir = '/templates/')
+    public function renderTemplate($filename, $vars = [], $templateDir = null)
     {
-        $filePath = __DIR__ . $templateDir . $filename;
+        if (!$templateDir) {
+            $filePath = __DIR__ . '/templates/' . $filename;
+        } else {
+            $filePath = $templateDir . $filename;
```
Incognito, line 1003: I really like the style used in https://github.com/propelorm/Propel2/pull/1491/files where the code uses `realpath` which would be less error-prone. Can we do something like that in this method?

PR #1519, opened 2019-09-18
Fix CS.

dereuromark:
Code sniffer is not happy about the additional unneeded newline in the docblocks.
This solves it.


dereuromark:
Can we merge this?

PR #1523, opened 2019-11-08
Added support for loading environment variables from .env file

GBeushausen:
Propel has the ability to create config values from environment variables. During development it's much easier to load these variables from the .env file.


Incognito:
Have you considered passing these from your project to the application without dotenv?

For example, on projects I work on where I usedotenv I'll load all runtime-envs with dotenv, and simply forward everything as configuration from one common place. This lets me decouple my code from depending this specific tool. I'd prefer this kind of approach instead. Right now this composer lock would mean every consumer of propel2 must also be compatible with dotenv `~2.3||~3.0||~4.0"`, which means someone who is ready to update to dotenv 5.0 must wait for a new propel2 patch or abandon propel (or fork it).

I like the idea and I see a lot of value in the feature, but I'd like to decouple it further.

composer.json
```diff
@@ -23,6 +23,7 @@
         "symfony/validator": "~2.3||~3.0.0||~3.1.0||^3.2.4||~4.0",
         "symfony/filesystem": "~2.3||~3.0||~4.0",
         "symfony/config": "~2.3||~3.0||~4.0",
+        "symfony/dotenv": "~2.3||~3.0||~4.0",
```
dereuromark, line 26: Instead of a hard dependency, this could also be a soft one, and we check if the class exists or not.
This way, most users not needing this wont require an additional dependency here.

.gitignore
```diff
@@ -6,6 +6,7 @@ autoload.php
 phpunit.xml
 pom.xml
 php-cs-fixer.phar
+.idea
```
dereuromark, line 9: Already in the list above.
So basically, the PR has now no more changes.

dereuromark:
Can you please fix up the PR to the changes you proposed initially?
Otherwise we need to close this one.

asaulenko (commented):
There is no actual diff with `master` in this PR anymore.

dereuromark:
Diff is in https://github.com/propelorm/Propel2/pull/1523/commits/d35188e84edc278c5856145d26a1dea49889d6ae etc, I guess the last merge/rebase might have removed the actual changes.

Incognito:
Previously I had asked:

> Have you considered passing these from your project to the application without dotenv?

It seems like coupling envvars directly into the library could lead to the need to do a lot of library overwrites or hacks, force very specific envvars in production, or limit applications to run against 1 db host.

Have these points been considered?

mringler:
@Incognito I am not sure if I understand your concern correctly. You are worried, that using symfony/dotenv as a required package in Propel might lead to incompatible version requirements between Propel and your project, when your project also requires symfony/dotenv, but with a later version. Is that correct?
If so, isn't that a potential problem with every requirement? What makes dotenv special or different from other packages?

Incognito:
Please consider this scenario:

1. Your have a website project using "symfony/dotenv": "^4.0"
2. You install this propel library into a symfony project
2. Your want to upgrade to "symfony/dotenv": "5.0"

You get a composer constraint conflict because your database driver demands a 4.0 version, so you are prevented from upgrading due to this dependency. The project must wait for a release of propel which supports the next version due to a hard constraint on how the internals of propel load environment variables. In general I assume projects fully control where configuration comes from and inject it into the libraries that need it to prevent this kind of superficial coupling and upgrade issues which can be quite challenging to fix when you need to do something slightly outside what the author imagined, or want to use specific names for envvars.

Please consider these issues:

1. The method is private and cannot be overridden by a library consumer who needs to fix it
2. The method will cause the try-catch to execute for every production web request served adding minor overhead
3. This method is an abstract base and will influence every file loaded (xml, json, ini, etc)
4. Propel overriding $_ENV here may break any project which has already assigned values here in some other means (now there are conflicting writes to this array).

I propose these alternatives:

1. Don't expose the project to dotenv in the first place, this will reduce the number of packages that propel is coupled to, rather just find a way to inject configuration.
2. Alternatively, create a dotenv loader and permit it to be run in dev environments but not production ones (https://github.com/propelorm/Propel2/blob/fee01cc8c230f732fc7a78e70f42da346fdeed79/src/Propel/Common/Config/Loader/LoaderResolver.php#L26 )

Presently I see the dependencies looking like this (arrows indicate composer requirements)

```
   ┌──────────┐
   │ Project  ├───────────┐
   └─────┬────┘           │
         │                │
         │                ▼
   ┌─────▼────┐      ┌──────────┐
   │ Propel   ├──────►  Dotenv  │
   └──────────┘      └──────────┘

```

I think it would make life for package consumers easier and more long-term sustainable if it looked like this:

```
   ┌──────────┐      ┌──────────┐
   │ Project  ├──────►  Dotenv  │
   └────┬─────┘      └──────────┘
        │
        │
   ┌────▼─────┐
   │ Propel   │
   └──────────┘
```

mringler:
Yes, I understand your concern regarding version conflicts in composer. My question was about what makes dotenv special. Because Propel already uses several symfony packages, and it seems like they create the exact same problem.

And I don't understand how overriding the method could fix a problem with composer dependencies - those are installed whether you use them or not, aren't they?

Incognito:
The vendor being symfony or not is unrelated to how I think about it, I'm simply looking at "should this library gain another dependency and is it worth the price of coupling? Is there another way to avoid that price?" I found each individual library was really something that had to be considered individually. The upgrades for the sf validator for instance was a total replatform of the concept, atleast one validator entirely changed purposes, and the auto-loading was it's own challenge. None of that had overlap with the problems in upgrading Console, yaml, config, etc. Each lib was their own thing, likely a result of sf by design not being monolithic.

The part with overriding the method: when a consumer of this project uses Propel they would not have the option of overriding and changing the behaviour, private kind of forbids projects from getting into the code and changing it if they need to. So either we provide it "omaske" like Ruby on Rails philosophy and double-down on "this is the way it is" or we give them the option to try and fight their way out of it with overriding and re-instantiating classes. That said, even making it protected here would still be quite painful for someone to override the behaviour to not use dotenv in a project.

mringler:
I agree that adding a dependency shouldn't be done willy-nilly. And I am also not convinced that this PR warrants it.

There already is a mechanism to separate global parameters and instance specific parameters (propel.json vs propel.json.dist). And I don't see how things get easier when allowing to put variables in a .env file instead of, say, a propel.ini file.
While it is important that Propel can retrieve configuration parameters through envvars, I see no need for Propel to be able to put them there (which is what this PR does). If someone insists on using env for configuration, they should have no problem loading the parameters themselves.
And if the conf vs conf.dist system is not working sufficiently, the focus should be on fixing that, instead of introducing workarounds.

So even if this PR weren't broken (changes removed in merge commit, wrong formatting, no tests), and while absolutely appreciating the participation, I don't think this is the way to go and would rather see this PR closed than merged.
Unless there is a really good argument for the opposite, that I am not seeing?

dereuromark:
Given the current state of this PR (no diff), lets close.
A new PR with proper changes including tests could be revisited.

PR #1529, opened 2020-02-20
alphabetical order in 'use' statements

DurandSacha:


dereuromark:
Refs https://github.com/propelorm/Propel2/issues/1592

dereuromark:
I will merge this even without the sniffer in place yet as this is very local and thus not conflicting with other topics.

PR #1530, opened 2020-02-20
add void return for application test

DurandSacha:


dereuromark:
The project is still PHP5 compatible, whats the reason to add those?

PR #1531, opened 2020-02-26
Void return for functions

DurandSacha:


ikeyan:
`: void` is supported on php 7.1 or later, so Propel2 cannot accept this PR, I think.

ikeyan:
@DurandSacha `: void` is supported on php 7.1 or later

dereuromark:
Duplicate of https://github.com/propelorm/Propel2/pull/1530

src/Propel/Common/Util/SetColumnConverter.php
```diff
@@ -58,7 +58,7 @@ public static function convertToInt($val, array $valueSet)
      *
      * @throws SetColumnConverterException
      */
-    public static function convertIntToArray($val, array $valueSet)
+    public static function convertIntToArray($val, array $valueSet) : string
```
PhilinTv, line 61: It returns array here.

src/Propel/Common/Util/SetColumnConverter.php
```diff
@@ -28,7 +28,7 @@ class SetColumnConverter
      *
      * @throws SetColumnConverterException
      */
-    public static function convertToInt($val, array $valueSet)
+    public static function convertToInt($val, array $valueSet) : void
```
PhilinTv, line 31: it returns 0 as well.

PhilinTv:
Hey @DurandSacha ,
Thank you for your contribution. There is a stable2.0.0 branch where we are going to drop the support of EOL PHP versions (incl php5 and php7.0). So you could start adjusting types there, but also you need to take care of the compatibility of the suggested return type and real return values. In some cases, such suggestions will require heavy refactoring and tests. 
Also, it's much better to do it all at once using tooling.

Cheers,

dereuromark:
We now have a new tag available you can build on top.
Closing this as the changes are not valid.

PR #1542, opened 2020-05-06
Only generate calls to parent hooks, if the class being generated has a base class

tspink:
The calls to parent hooks should only be generated if that class has a base class, otherwise the static analyser complains about the use of "parent", when there is no parent.

### DIFF
--- a/src/Propel/Generator/Builder/Om/templates/baseObjectMethodHook.php
+++ b/src/Propel/Generator/Builder/Om/templates/baseObjectMethodHook.php
@@ -7,9 +7,11 @@
      */
     public function preSave(ConnectionInterface $con = null)
     {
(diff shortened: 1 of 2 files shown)


src/Propel/Generator/Builder/Om/templates/baseObjectMethodHook.php
```diff
@@ -7,9 +7,11 @@
      */
     public function preSave(ConnectionInterface $con = null)
     {
+        <?php if ($hasBaseClass): ?>
```
stereomon, line 10: Shouldn't we use 

```
<?php if ($hasBaseClass): ?>
    return parent::preSave();
<?php else: ?>
   return true;
<?php endif ?>
```

PR #1549, opened 2020-05-12
Fix issue where propel reverse breaks with system-versioned tables

zwhedbee:
MariaDB 10.3.4 introduced system-versioned tables (https://mariadb.com/kb/en/temporal-data-tables/).

There is a known issue (possibly a bug) that can cause issues for ORM tools like Propel, described here: https://jira.mariadb.org/browse/MDEV-16857?focusedCommentId=117570&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-117570

"ORM tool will be confused as the primary key definition in the database doesn't match the one defined in application code after system versioning is enabled."

Basically, system versioning creates an invisible primary key that shows up with "SHOW KEYS FROM table" but not with "SHOW COLUMNS FROM table". This PR fixes that by skipping these invisible primary key columns when generating the schema.

### DIFF
--- a/src/Propel/Generator/Reverse/MysqlSchemaParser.php
+++ b/src/Propel/Generator/Reverse/MysqlSchemaParser.php
@@ -431,7 +431,9 @@ class MysqlSchemaParser extends AbstractSchemaParser
                 continue;
             }
             $name = $row['Column_name'];
-            $table->getColumn($name)->setPrimaryKey(true);
+            if ($column = $table->getColumn($name)) {
+                $column->setPrimaryKey(true);
+            }
         }
     }
 
--- a/src/Propel/Generator/Reverse/MysqlSchemaParser.php
+++ b/src/Propel/Generator/Reverse/MysqlSchemaParser.php
@@ -431,7 +431,8 @@ class MysqlSchemaParser extends AbstractSchemaParser
                 continue;
             }
             $name = $row['Column_name'];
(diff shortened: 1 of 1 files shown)


src/Propel/Generator/Reverse/MysqlSchemaParser.php
```diff
@@ -431,7 +431,9 @@ protected function addPrimaryKey(Table $table)
                 continue;
             }
             $name = $row['Column_name'];
-            $table->getColumn($name)->setPrimaryKey(true);
+            if ($column = $table->getColumn($name)) {
```
PhilinTv, line 434: Good catch!

Could you please avoid var assignment and condition on the same line?

MariaDB is not the only MySQL engine Propel supports.
Is there a test, that could confirm that other engines are not affected and for MariaDB it's solved?

dereuromark:
ping @zwhedbee 
Are you able to fix up the CS issue? Otherwise, we can also do that post-merge.

zwhedbee:
Sorry for the issues with the original PR and that I haven’t been able to resolve those. Post-merge would be fine.

src/Propel/Generator/Reverse/MysqlSchemaParser.php
PhilinTv, line 434:
```suggestion
            $column = $table->getColumn($name);
            if ($column) {
```

Issue #575, opened 2014-03-11
Using PSR-4 for model classes


gharlan:
It should be possible to use [PSR-4](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-4-autoloader.md) for generated model classes. Correct me if I'm wrong, but at the moment Propel2 supports PSR-0 only.

polybuildr:
Any updates on this issue?

marcj:
No

PR #1555, opened 2020-05-25
Use PSR-4 now.

dereuromark:
Refs https://github.com/propelorm/Propel2/issues/575

We should get all to psr-4 including tests (which currently are not yet in autoloader at all)

This would probably be for the 2.0 stable release.


Incognito:
Please rebase

dereuromark:
Done, looks as green as before.

PR #1559, opened 2020-06-05
allow to use non-primary columns in crossref tables

kasparsj:
Hi guys,

Not sure what was the reason for the change, but in Propel1 it was not mandatory to make the crossref table foreign keys also primary keys. It was handy when the crossref table relation had to be referenced further elsewhere.

I don't see any negative effects enabling this, but please correct me if I'm wrong.

Here is a pseudo-example:
```
<table name="user_group" isCrossRef="true">
  <column name="id" type="integer" primaryKey="true" required="true" />
  <column name="user_id" type="integer"/>
  <column name="group_id" type="integer"/>
  <foreign-key foreignTable="user">
    <reference local="user_id" foreign="id"/>
  </foreign-key>
  <foreign-key foreignTable="group">
    <reference local="group_id" foreign="id"/>
  </foreign-key>
</table>

<table name="user_group_logs">
    <column name="id"type="INTEGER" primaryKey="true" autoIncrement="true" required="true" />
    <column name="user_group_id" type="INTEGER" />
    <column name="log" type="VARCHAR" />
    <foreign-key foreignTable="user_group">
        <reference local="user_group_id" foreign="id" />
    </foreign-key>
</table>
```


dereuromark:
We now have master almost green ( https://travis-ci.org/github/propelorm/Propel2 )
Please try to rebase yours against that one, that will make it easier to verify your changes.

dereuromark:
@kasparsj ping - master is now fully green and if your change passes, we can continue here.

src/Propel/Generator/Model/CrossForeignKeys.php
```diff
@@ -130,9 +130,9 @@ public function getIncomingForeignKey()
      * @param  ForeignKey $fk
      * @return bool
      */
-    public function isAtLeastOneLocalPrimaryKeyNotCovered(ForeignKey $fk)
+    public function isAtLeastOneLocalColumnNotCovered(ForeignKey $fk)
```
Incognito, line 133: This change would invalidate the above comment as the logic is expanded beyond PKs.

dereuromark:
@kasparsj Are you interested in following up on this one? Then it can make it into upcoming release
Otherwise we need to de-scope it into next one.

dereuromark:
Looks good, we just need a test case as protection against regression here.

asaulenko (changes requested):
Needs some test to be developed, I guess.

PR #1563, opened 2020-06-12
Restore (most of) Travis

Incognito:
### Important changes:
- All platforms past EOL were removed, modern platforms replaced them (eg, dropped php5, added php7.4, dropped postgres 9.2, added postgres 11)
- PHPUnit upgraded to 7.x
- PgsqlSchemaParser needs to treat `CURRENT_TIMESTAMP` as an expression, the quotes break it in pg10+. There are likely more functions that need to be added here too: https://www.postgresql.org/docs/11/functions-datetime.html#FUNCTIONS-DATETIME-TABLE
- `tests/bin/setup*.sh` all run `./bin/propel test:prepare` now. I'm not sure how it was ever triggered before. It appears that it was added many years ago but removed (and now I've added it back).

### Outstanding issues:

I suggest we merge this branch now (pending that everything in the review is fine) and we fix these items as individual tickets.

- Some issues with PHP7.4 on every database (including agnostic) except Postgres.
- Travis-ci doesn't know how to install Maria 5.5 and 10.1, we have possible work-around options but I haven't attempted them yet.
- Travis-ci fails to install postgres 12, I haven't investigated the cause for this.
- Some code-drift between the `tests/bin/setup*.sh` and `tests/bin/setup*.bat` scripts, please take a closer look if you know how to execute the ./bin/propel commands on windows.
- Incompatible with `symfony/validator >= 4.2` due to some class movement.
- Symfony 5.0 incompatabilities
- Mysql function expressions like CURRENT_TIMESTAMP should no longer be injected as a string in mysql, needs broader test coverage to check for possible `defaultExpr` assignments. (Mysql changed the behaviour in the last 2 years, they no longer support the old syntax). We need specific tests and exclusions for postgres adapter, putting them into the generic schema will break databases that don't support those functions.  

### Future cleanups:
- Mysql is not versioned (it wasn't previously either). We should add specific builds like we did with Maria and Postgres.
- Incompatibilities with PHP8 (but that should get its own specific PR).
- Inherited dependency on symfony/translation through the symfony/validator should become a first-class dependency as it is used by the library (however there is some nuance to this).

Here's a snapshot of what tests look like right now (in the event you're trying to fix these tests in 2023 and Travis doesn't hold the record).

screencapture-travis-ci-org-github-spryker-Propel2-builds-697658166-2020-06-12-17_29_37


Incognito:
I think we should squash merge this.

.gitignore
```diff
@@ -1,3 +1,4 @@
+.idea/
```
PhilinTv, line 1: This should go with higher prio in a separate pr probably

composer.json
```diff
@@ -16,18 +16,18 @@
         }
     ],
     "require": {
-        "php": ">=5.5",
-        "symfony/yaml": "~2.3||~3.0||~4.0||~5.0",
```
PhilinTv, line 20: why do you drop 5.0 support? that was the main reason for the previous compatibility pr

Incognito: This PR is not built on that PR

PhilinTv: In the master we already support SF5. Why do you drop it?

src/Propel/Generator/Reverse/PgsqlSchemaParser.php
```diff
@@ -309,7 +309,9 @@ protected function addColumns(Table $table, $oid)
             }
 
             if (null !== $default) {
-                if ("'" !== substr($default, 0, 1) && strpos($default, '(')) {
+                $containsFunctionCall = "'" !== substr($default, 0, 1) && strpos($default, '(');
+                $isCurrentTimestamp = $default === 'CURRENT_TIMESTAMP';
```
PhilinTv, line 313: There are a few experessions were introduced
current_date
current_time
current_timestamp
localtime
localtimestamp

Are they handled elsewhere?

.travis.yml
```diff
@@ -1,158 +1,278 @@
...
+  depth: 1
+  quiet: true
+
+dist: bionic
+
 php:
-  - 5.6
+  - 7.1
+  - 7.2
+  - 7.3
```
dereuromark, line 13: Only min/max is needed, the other versions are already tested in matrix below quite extensively, from what we can see.

tests/bin/setup.mysql.bat
```diff
@@ -43,6 +43,7 @@ call %DIR%\base.bat :check
 if "%errorlevel%" == "1" exit /B 1
 
 "%mysql%" --host="%DB_HOSTNAME%" -u"%DB_USER%" %pw_option% -e ^"^
+SET GLOBAL sql_mode=(SELECT REPLACE(@@sql_mode,'ONLY_FULL_GROUP_BY',''));^
```
PhilinTv, line 46: It solves the BC-problem for tests, but not for the Propel users, right?

dereuromark: Users have to do that themselves for years.
We can go away from MySQL specific Hacks in follow up PRs maybe.

src/Propel/Generator/Reverse/PgsqlSchemaParser.php
PhilinTv: Are you sure you get the only uppercase?

Incognito: I think all of those should get added but I'd like to add tests to make sure they'll work over all supported platforms still.

tests/bin/setup.mysql.bat
Incognito: I spent some time debating this but came to this decision after looking at the matrix of options

<table><thead><tr><th>Does upgrading <br>cause a major?</th><th>Add In Tests</th><th>Add in Connection<br></th><th>Make code work without it</th></tr></thead><tbody><tr><td>Customer Has <br>ONLY_FULL_GROUP_BY<br>in project</td><td>No</td><td>No</td><td>Yes</td></tr><tr><td>Customer<br>Does not have<br>ONLY_FULL_GROUP_BY<br>in project</td><td>No</td><td>Yes</td><td>Maybe</td></tr></tbody></table>

I think this feature should get deprecated and replaced in the future, and I'm in favour of adding a new adapter to support modern MySQL (there was some discussion about it in old issue tickets a few years ago). For the time being, I think this is the best choice to stabilise the existing system and behaviours.

.travis.yml
Incognito: We have edgecases everywhere in this project that cannot be found with minmax.

The matrix expansion of travis doesn't work well for this project anyway, I'd like to remove it so we can correctly test mysql versions.

composer.json
Incognito: I tried to restore these ~5.0 versions and it looks like a deeper level of investigation will be needed.
Screenshot from 2020-06-15 10-17-13

Incognito: I attempted this again and we are not able to run as-is with 5.x, I managed to upgrade a few deps but a lot will be stuck in 4.x until we have a fix.

dereuromark:
@PhilinTv Please squash merge this.

PhilinTv:
e24294a contains psql expression dictionary.

src/Propel/Generator/Reverse/PgsqlSchemaParser.php
```diff
@@ -325,6 +325,34 @@ protected function addColumns(Table $table, $oid)
...
+        if ($containsFunctionCall) {
+            return true;
+        }
+
+        $defaultColumnValueExpressions = [
+            'CURRENT_DATE' => 'CURRENT_DATE',
+            'CURRENT_TIME' => 'CURRENT_TIME',
+            'CURRENT_TIMESTAMP' => 'CURRENT_TIMESTAMP',
+            'LOCALTIME' => 'LOCALTIME',
+            'LOCALTIMESTAMP' => 'LOCALTIMESTAMP'
```
Incognito, line 346:
```suggestion
            'LOCALTIME' => 'LOCALTIME',
            'LOCALTIMESTAMP' => 'LOCALTIMESTAMP'
```

The others require pgsql-specific tests as they don't work on other db engines.

PR #1565, opened 2020-06-16
Use term primary/replica instead, deprecate in BC way.

dereuromark:
Spryker would like to start using primary/replica functionality of Propel.
We would like to have this change first, though. It keeps naming consistent with modern tech like aws rds https://aws.amazon.com/rds/features/read-replicas/

### DIFF
--- a/tests/Propel/Tests/Runtime/Connection/ConnectionManagerPrimaryReplicaTest.php
+++ b/tests/Propel/Tests/Runtime/Connection/ConnectionManagerPrimaryReplicaTest.php
@@ -12,22 +12,22 @@ namespace Propel\Tests\Runtime\Connection;
 
 use Propel\Tests\Helpers\BaseTestCase;
 
-use Propel\Runtime\Connection\ConnectionManagerMasterSlave;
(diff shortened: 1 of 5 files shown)


tests/Propel/Tests/Runtime/Connection/ConnectionManagerPrimaryReplicaTest.php
```diff
@@ -12,22 +12,22 @@
 
 use Propel\Tests\Helpers\BaseTestCase;
 
-use Propel\Runtime\Connection\ConnectionManagerMasterSlave;
+use Propel\Runtime\Connection\ConnectionManagerPrimaryReplica;
 use Propel\Runtime\Adapter\Pdo\SqliteAdapter;
 
 use \PDO;
 
-class ConnectionManagerMasterSlaveTest extends BaseTestCase
```
PhilinTv, line 20: Please created another test for the ConnectionManagerPrimaryReplica.
Until ConnectionManagerMasterSlave is only deprecated this test should remain.

Issue #1571, opened 2020-06-23
PHP 7.4 incompatibilities [PHP]


Incognito:
Related to: https://github.com/propelorm/Propel2/issues/1541

Some issues with PHP7.4 on every database (including agnostic) except Postgres which has no issues.
Test errors are usually in one of these forms:
```
2) Propel\Tests\Generator\Platform\MysqlPlatformTest::testAddExtraIndicesForeignKeys
Trying to access array offset on value of type null
```

Or
 
```
1) Propel\Tests\Generator\Builder\Om\GeneratedQueryObjectColumnTypeTest::testWhere
object columns are searchable by serialized object using where()
Failed asserting that 0 matches expected 1.
```

PR #1581, opened 2020-06-24
Behaviour was changed in PHP7.4, this ternary preserves pre-existing …

Incognito:
Fixes https://github.com/propelorm/Propel2/issues/1571

### DIFF
--- a/src/Propel/Runtime/Map/DatabaseMap.php
+++ b/src/Propel/Runtime/Map/DatabaseMap.php
@@ -93,7 +93,7 @@ class DatabaseMap
         $table->setDatabaseMap($this);
         $this->tables[$table->getName()] = $table;
         $phpName = $table->getClassName();
-        if ('\\' !== $phpName[0]) {
+        if ('\\' !== $phpName ? $phpName[0] : null) {
             $phpName = '\\' . $phpName;
         }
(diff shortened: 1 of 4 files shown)


src/Propel/Runtime/Map/DatabaseMap.php
```diff
@@ -93,7 +93,7 @@ public function addTableObject(TableMap $table)
         $table->setDatabaseMap($this);
         $this->tables[$table->getName()] = $table;
         $phpName = $table->getClassName();
-        if ('\\' !== $phpName[0]) {
+        if ('\\' !== ($phpName ? $phpName[0] : null)) {
```
dereuromark, line 96: here it makes more sense to use

    if ($phpName && '\\' !== $phpName[0]) {

since `\` itself doesn't make sense.

dereuromark: But then also the next assignment is weird.
Maybe return early instead?

Incognito: Can you use the propose changes feature? It would be faster.

Incognito:
Remaining failure: https://github.com/propelorm/Propel2/issues/1583

PR #1582, opened 2020-06-24
Required root name for symfony/config 5.X versions

c15k0:
In the version 4.4 the vendor config show this warning message:

-- > https://github.com/symfony/config/blob/4.4/Definition/Builder/TreeBuilder.php (30)

in the version 5.X is mandatory to set  the root name in the constructor in order to user the Propel generator.

### DIFF
--- a/src/Propel/Common/Config/PropelConfiguration.php
+++ b/src/Propel/Common/Config/PropelConfiguration.php
@@ -29,8 +29,14 @@ class PropelConfiguration implements ConfigurationInterface
      */
     public function getConfigTreeBuilder()
     {
+        // Create TreeBuilder with default param, before version 4.4 is not required
         $treeBuilder = new TreeBuilder('propel');
-        $rootNode = $treeBuilder->getRootNode();
+        // Check if exists root method for versions before 4.4
+        if(method_exists($treeBuilder, 'root')) {
+            $rootNode = $treeBuilder->root('propel');
+        } else { // Versions greater than 4.4 and 5.x
+            $rootNode = $treeBuilder->getRootNode();
+        }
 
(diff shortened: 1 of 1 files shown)


dereuromark:
You will also have to adjust the constraints then: https://github.com/propelorm/Propel2/blob/master/composer.json#L25

c15k0:
It is not required to upgrade the JSON because it is only a warning notification for 4.X versions. When the other dependencies will upgrade to versions 5.X, this class is prepared to use the same branch version for the Symfony component.

src/Propel/Common/Config/PropelConfiguration.php
```diff
@@ -29,8 +29,14 @@ class PropelConfiguration implements ConfigurationInterface
      */
     public function getConfigTreeBuilder()
     {
-        $treeBuilder = new TreeBuilder();
-        $rootNode = $treeBuilder->root('propel');
+        // Create TreeBuilder with default param, before version 4.4 is not required
+        $treeBuilder = new TreeBuilder('propel');
+        // Check if exists root method for versions before 4.4
+        if(method_exists($treeBuilder, 'root')) {
```
dereuromark, line 35:
```suggestion
        if (method_exists($treeBuilder, 'root')) {
```

dereuromark:
I applied your patch on https://github.com/propelorm/Propel2/pull/1590 to outline in travis the issues we are still facing
- Translator issue in tests only (minor)
- small TypeError: Return value of "Propel\Generator\Command\ConfigConvertCommand::execute()" must be of the type int, "null" returned

Once those are resolved, we should be good for Symf5 compatibility.

dereuromark:
We merged the fix to master now.

Issue #1583, opened 2020-06-24
Mysterious inconsistency with comparison operators on BLOBs [Sqlite] [TravisCI]


Incognito:
There is one single error showing up in Travis for a build of `agnostic` + `PHP7.4`.
![Screenshot from 2020-06-24 16-06-24](https://user-images.githubusercontent.com/379322/85571763-b5085c00-b634-11ea-8f5f-831379959097.png)

As far as I can tell, all 4 agnostic tests should fail here.

Between the PHP7.3 and 7.4 builds, both jobs are running:
- The same linux image
- The same PHP extensions (I didn't check versions).
- The same composer modules and versions
- The same version of sqlite (I will explain why this is relevant below)

The failing test (`\Propel\Tests\Generator\Builder\Om\GeneratedQueryObjectColumnTypeTest::testWhere`) runs this line of code:

 ```
        $nb = \ComplexColumnTypeEntity10Query::create()
            ->where('ComplexColumnTypeEntity10.Bar LIKE ?', '%1234%')
            ->count();
```

The failure here is because of how agnostic tests work. If you look in the test setup of this class, you'll see the true database for "agnostic": `\Propel\Generator\Util\QuickBuilder::build` creates an in-memory sqlite3 database (no file created, data is lost at the end of execution). **This means that this "agnostic" test depends on sqlite3, and that's where queries are actually executed.**

So I created a local docker instance of PHP-7.4, 7.3, and 7.2 and tried three different linux distributions. This test fails on every possible combination in the exact same way. The only known place this test passes is Travis-CI 7.1, 7.2 and 7.3.

But a bit deeper into what's going on with the failure:
If we look at the actual data stored, we can see Propel is storing what Sqlite3 calls "blob literals", which means they were encoded as hexadecimal strings instead of literal strings. This behaviour entirely surprised me because I'm not really using blobs in sqlite3 very often:

```
sqlite> select X'50726F70656C32';
Propel2
```

You might be thinking: "well of course, that's just the string encoding for input". But actually, it's not _just_ one way to present  the data; the hex-encoding _is part of_ the data. This makes it a blob literal instead of a string literal.

```
sqlite> select X'50726F70656C32' = 'Propel2';
0
sqlite> select X'50726F70656C32' = hex('Propel2');
0
sqlite> select cast(X'50726F70656C32' as text) = 'Propel2';
1
```

It's even different in the hexdump of the database itself when you insert values this way (these examples below are artifacts from the testcase):

```
[...]

PR #1584, opened 2020-06-24
Split out LIKE condition on objects, which is buggy in php74

PhilinTv:
In order to unblock the PHP7.4, SF5 dependencies on projects and start gathering more collaborative feedbacks on their incompatibility I'm disabling LIKE validation on objects (cool, but not documented capability). We recover it later.

### DIFF
--- a/tests/Propel/Tests/Generator/Builder/Om/GeneratedQueryObjectColumnTypeTest.php
+++ b/tests/Propel/Tests/Generator/Builder/Om/GeneratedQueryObjectColumnTypeTest.php
@@ -76,10 +76,12 @@ EOF;
      */
     public function testWhereLike()
     {
+        $this->markTestSkipped('There are inconsistencies regarding the handling of this statement on different platforms.');
+
         $nb = \ComplexColumnTypeEntity10Query::create()
             ->where('ComplexColumnTypeEntity10.Bar LIKE ?', '%1234%')
             ->count();
-//        $this->assertEquals(1, $nb, 'object columns are searchable by serialized object using where()');
+        $this->assertEquals(1, $nb, 'object columns are searchable by serialized object using where()');
     }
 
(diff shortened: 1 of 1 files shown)


tests/Propel/Tests/Generator/Builder/Om/GeneratedQueryObjectColumnTypeTest.php
```diff
@@ -64,16 +64,24 @@ public function testColumnHydration()
...
 
+    /**
+     * Recover this undocumented functionality
+     */
+    public function testWhereLike()
+    {
+        $nb = \ComplexColumnTypeEntity10Query::create()
+            ->where('ComplexColumnTypeEntity10.Bar LIKE ?', '%1234%')
+            ->count();
+//        $this->assertEquals(1, $nb, 'object columns are searchable by serialized object using where()');
```
Incognito, line 82:
```suggestion
    {
        $this->markTestSkipped('There are inconsistencies regarding the handling of this statement on different platforms.');

        $nb = \ComplexColumnTypeEntity10Query::create()
            ->where('ComplexColumnTypeEntity10.Bar LIKE ?', '%1234%')
            ->count();
        $this->assertEquals(1, $nb, 'object columns are searchable by serialized object using where()');
```

It would be better to not let this be mark as as a passing test, but rather intentionally skipped due to platform issues.

Incognito: Relates to https://github.com/propelorm/Propel2/issues/1583

PhilinTv: Tests without assertion are marked as risky by phpunit.

PR #1586, opened 2020-06-25
Remove prefer lowest from allow failure

dereuromark:
prefer lowest should fail the built if they fail.

~~Also removed non-minimal prefer lowest as they have no additional value~~

we now still run 42 jobs, which we could further limit down
but this is probably a fair compromise for now

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,8 @@ dist: bionic
 
 php:
   - 7.1
+  - 7.2
+  - 7.3
   - 7.4
 
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,6 @@ dist: bionic
 
 php:
   - 7.1
-  - 7.2
-  - 7.3
   - 7.4
 
 env:
@@ -36,7 +34,7 @@ before_script:
   - if [[ $DB != 'agnostic' && $DB != 'sqlite' ]]; then ./tests/bin/setup.$DB.sh; fi
 
 script:
-  - ./vendor/bin/phpunit -v -c tests/$DB.phpunit.xml
+  - vendor/bin/phpunit -v -c tests/$DB.phpunit.xml
 
 jobs:
   include:
@@ -335,7 +333,7 @@ jobs:
       script:
         - vendor/bin/validate-prefer-lowest
         - composer show
-        - ./vendor/bin/phpunit -v -c tests/$DB.phpunit.xml
+        - vendor/bin/phpunit -v -c tests/$DB.phpunit.xml
 
     - name: Minimum versions on php7.3
       php: 7.3
@@ -346,7 +344,7 @@ jobs:
       script:
         - vendor/bin/validate-prefer-lowest
         - composer show
-        - ./vendor/bin/phpunit -v -c tests/$DB.phpunit.xml
+        - vendor/bin/phpunit -v -c tests/$DB.phpunit.xml
 
     - name: Minimum versions on php7.4
       php: 7.4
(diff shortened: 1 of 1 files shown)


dereuromark:
This is about what the title says.
https://travis-ci.org/github/propelorm/Propel2/builds/702351276
Still has all versions testes as we discussed in detail.

.travis.yml
```diff
@@ -9,8 +9,6 @@ dist: bionic
 
 php:
   - 7.1
-  - 7.2
-  - 7.3
```
Incognito, line 13: Please restore

Incognito:
Why does a failing test get added to acceptable failures in master?
Screenshot from 2020-06-30 10-14-50

dereuromark:
If there is a fail, we need to fix it.

Incognito:
I found problems with this tool, for example, when I manually specify the lowest versions of things I don't get this issue.

dereuromark:
What tool?
The issue seems to be the huge deprecation list of trying to run php74 on those outdated packages
And:

> There was 1 failure:
1) Propel\Tests\FieldnameRelatedTest::testFieldNameTypeConstants
Can not initialize fixtures.
/home/travis/build/propelorm/Propel2/tests/Propel/Tests/TestCaseFixtures.php:111

The truth is: It doesn't make sense to run a 2020 PHP env on a 2015 code base and below.
You either run that old one with php71 and below, or you try to at least upgrade those dependency packages to a reasonable version.
Thats why I was against keeping those artificial runs around that bear no real value to us, the project or anyone else.
Trying to fix those only limits other people in using it on their platform target.

.travis.yml
dereuromark: 7.2 and 7.3 is already extensively tested with real DBs here (half of the 44 jobs!).
Those two here have no meaning and relevance.
They are not even running on the real DBs, they only provide a smoke test on basic package compatibility. Thus only min/max needed.
Instead we should focus on getting my CHECKS job in, with phpstan and cs, as that one will provide way more useful insights on top of the matrix already existing.

We compromised on keeping your jobs for 7.2/7.3 in the matrix already, even with them also being already highly questionable in terms of overall usefulness. The likelyhood of any issue is way below << 0.001 % and the job running time/costs outweigh it already immensely.

dereuromark:
I rebased against master and sqashed the history.

dereuromark:
This PR is now ready and complete as it is, I explained above in detail.

.travis.yml
Incognito: I've explained this many times: I have bugs that can't be found when you do this.

>half of the 44 jobs

I don't care about the number of jobs that run in travis. Having computers do lots of meticulous-pain-staking-work for us is the entire point of having CI.

> Those two here have no meaning and relevance.

Here's the real stuff I'm working on:
versions

Again: If you do this, nobody can find and fix these bugs, and we can't keep propel contributors from breaking versions we support.

I don't care if PHP "tends to" do some specific behaviour or the maintainers of symfony "tend to" not do something. If you do the work and run the tests on something other than min+max, they fail.

> They are not even running on the real DBs,

Doch. It runs on a real database. All tests run against real databases.

- Agnostic is a suite that executes on sqlite without having a specific adapter in the way.
- Sqlite is literally running on sqlite
- mysql is literally running on sqlite

There's landmines between versions of symfony, php, and databases for everything. I've had to make a grid just to map this out in my other work as I try to make tests pass everywhere because there's no sane way to keep track of all these cases.

This is just my progress as of today. Half of those checkmarks were :x: on Monday.
Screenshot from 2020-07-01 09-55-40

The reality is we have lots of external dependencies that change over time, and lots of PHP versions that are blocking some of those changes, and lots of databases to support.

If you remove coverage or visibility to the thing I've been struggling with for weeks I will not support it.

dereuromark: Can you prove to me you cannot see them with the remaining ones still in place?
I only removed non relevant ones.

.travis.yml
```diff
@@ -9,8 +9,6 @@ dist: bionic
...
   - 7.4
```
Incognito, line 12:
```suggestion
  - 7.2
  - 7.3
  - 7.4
```

I can't push commits to this repo.

Issue #1570, opened 2020-06-23
Maria 10.1 [Travis-CI] [Easy]


Incognito:
Related to: https://github.com/propelorm/Propel2/issues/1541 ,  https://github.com/propelorm/Propel2/issues/1552

Travis-CI is unable to install MariaDB 10.1, it should be possible to manually install these via apt, but this is only a few months before it reaches EOL (17 Oct 2020).

Incognito:
Error from asking travis to install 10.1

```
mariadb
Installing MariaDB version 10.1
Executing: /tmp/apt-key-gpghome.e1s5VbXLRp/gpg.1.sh --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 0xf1656f24c74cd1d8
gpg: key F1656F24C74CD1D8: 7 signatures not checked due to missing keys
gpg: key F1656F24C74CD1D8: public key "MariaDB Signing Key <signing-key@mariadb.org>" imported
gpg: Total number processed: 1
gpg:               imported: 1
Hit:1 http://security.ubuntu.com/ubuntu bionic-security InRelease
Get:2 http://nyc2.mirrors.digitalocean.com/mariadb/repo/10.1/ubuntu bionic InRelease [6,247 B]
Get:3 http://archive.ubuntu.com/ubuntu bionic InRelease [242 kB]
Get:4 http://nyc2.mirrors.digitalocean.com/mariadb/repo/10.1/ubuntu bionic/main amd64 Packages [11.2 kB]
Hit:5 http://apt.postgresql.org/pub/repos/apt bionic-pgdg InRelease
Get:6 http://archive.ubuntu.com/ubuntu bionic-updates InRelease [88.7 kB]
Get:7 http://archive.ubuntu.com/ubuntu bionic-backports InRelease [74.6 kB]
Get:8 http://archive.ubuntu.com/ubuntu bionic/main Sources [1,063 kB]
Get:9 http://archive.ubuntu.com/ubuntu bionic/universe Sources [11.5 MB]
Get:10 http://archive.ubuntu.com/ubuntu bionic/restricted Sources [5,823 B]
Get:11 http://archive.ubuntu.com/ubuntu bionic/multiverse Sources [216 kB]
Get:12 http://archive.ubuntu.com/ubuntu bionic/main amd64 Packages [1,344 kB]
Get:13 http://archive.ubuntu.com/ubuntu bionic/main i386 Packages [1,328 kB]
[...]

PR #1589, opened 2020-06-26
Adds Maria 10.1

Incognito:
Fixes: https://github.com/propelorm/Propel2/issues/1570

Uses a deprecated unix distro because it's the only one from Travis that actually works with maria 10.1, everyone else using travis-ci fixed this with custom hacks or docker.

Screenshot from 2020-06-26 10-16-24

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,19 +41,27 @@ script:
 jobs:
   include:
     - php: 7.1
+      dist: trusty
       env: DB=mysql MARIADB=10.1
+      services:
       addons:
         mariadb: 10.1
     - php: 7.2
+      dist: trusty
       env: DB=mysql MARIADB=10.1
+      services:
       addons:
         mariadb: 10.1
     - php: 7.3
+      dist: trusty
       env: DB=mysql MARIADB=10.1
+      services:
       addons:
         mariadb: 10.1
     - php: 7.4
+      dist: trusty
       env: DB=mysql MARIADB=10.1
+      services:
       addons:
         mariadb: 10.1
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,24 +40,22 @@ script:
 
 jobs:
   include:
-# Travis-ci doesn't know how to install Maria 10.1
-#   - php: 7.1
-#     env: DB=mysql MARIADB=10.1
-#     addons:
-#       mariadb: 10.1
-#   - php: 7.2
-#     env: DB=mysql MARIADB=10.1
-#     addons:
-#       mariadb: 10.1
-#   - php: 7.3
-#     env: DB=mysql MARIADB=10.1
-#     addons:
-#       mariadb: 10.1
-#   - php: 7.4
-#     env: DB=mysql MARIADB=10.1
-#     addons:
-#       mariadb: 10.1
-#       TODO add maria 5.5, it is not EOL yet.
+    - php: 7.1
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -40,24 +40,30 @@ script:
...
-#     addons:
-#       mariadb: 10.1
-#       TODO add maria 5.5, it is not EOL yet.
+    - php: 7.1
+      dist: trusty
+      env: DB=mysql MARIADB=10.1
+      services:
+      addons:
+        mariadb: 10.1
+    - php: 7.2
```
dereuromark, line 49: Please only always add min, max. There is no point in adding php in between.

Incognito: -1. We have edge cases all the time.

dereuromark: Not in that language spectrum. PHP min+max solve all those issues, always.
We have years of experience proving this.

Incognito:
```$ diff 71 72
8,9c8,9
< phpdocumentor/reflection-docblock  4.3.4   
< phpdocumentor/type-resolver        1.0.1   
---
> phpdocumentor/reflection-docblock  5.1.0   
> phpdocumentor/type-resolver        1.2.0   
33,34c33,35
< symfony/filesystem                 v4.4.10 
< symfony/finder                     v4.4.10 
---
> symfony/deprecation-contracts      v2.1.2  
> symfony/filesystem                 v5.1.2  
> symfony/finder                     v5.1.2  
39c40
< symfony/service-contracts          v1.1.8  
---
> symfony/service-contracts          v2.1.2  
41c42
< symfony/translation-contracts      v1.1.7  
---
> symfony/translation-contracts      v2.1.2  
43c44
< symfony/yaml                       v4.4.10 
---
> symfony/yaml                       v5.1.2  
```

This will get worse as we start supporting libraries that need to support php8 and start dropping 7.1, 7.2, etc. The dependencies we heavily leverage here are changing all the time and result in actually different versions, which creates actually different bugs for us.

dereuromark: Once PHP8 becomes an issue, we can discuss that
for now, in 7.1-7.4, testing 7.2 and 7.3 is pointless as it provides no additional info here. none.

dereuromark: to be clear: 7.1 and 7.4 as min and max showcase the end of spectrums for PHP dependency, the language itself has a clear way of introducing and deprecating/removing functionality. so the ones in between never show anything that isn't shown by the outer ones.

dereuromark: I would like us to spend resources, and our time, on the other things still open, that matter.
Running 20 from 40+ tests in the suite for nothing at all is a waste of resources, and slows everyone down.

Incognito: The packages installed for 71 and 72 and 73 are not the same because of package constraints against php. We need to run everything.

In getting travis to pass I encountered non-stop silly edgecases that were hidden somewhere like one exact package range that was only installable under some other co-conditions of composer or travis.

This is not a showcase, it is a CI matrix to find exactly that kind of issue. I see 0 value in removing builds that test exact configurations.

dereuromark: Yes, of course. But the min and max range cover the extremes already.
The packages in between are not relevant for normal use cases.
Even if you find a glitch in the 0.0001% of cases, that doesnt warrant a 50%+ increase in job runs.
There is no benefit here.

PR #1591, opened 2020-06-26
Complexity reduction database model

Incognito:
As I work towards needing to change ORM and data mapping, reducing complexity to these "foundational" classes will help me both learn and maintain that code.

Previous:
Screenshot from 2020-06-26 16-14-17
https://codeclimate.com/github/propelorm/Propel2/src/Propel/Generator/Model/Database.php

After

Screenshot from 2020-06-26 16-39-52

### DIFF
--- a/src/Propel/Generator/Model/Database.php
+++ b/src/Propel/Generator/Model/Database.php
@@ -105,6 +105,11 @@ class Database extends ScopedMappingModel
     protected $tablePrefix;
 
     /**
+     * @var TablesStringifier
+     */
+    protected $tablesStringifier;
+
+    /**
      * Constructs a new Database object.
      *
      * @param string            $name     The database's name
@@ -135,6 +140,7 @@ class Database extends ScopedMappingModel
         $this->tablesByName              = [];
         $this->tablesByPhpName           = [];
         $this->tablesByLowercaseName     = [];
+        $this->tablesStringifier         = new TablesStringifier();
     }
 
     protected function setupObject()
@@ -528,6 +534,12 @@ class Database extends ScopedMappingModel
         }
     }
 
+    /**
+     * Removes table from this database.
+     *
+     * @param  Table|array $table
+     * @return void
+     */
     public function removeTable(Table $table)
     {
         if ($this->hasTable($table->getName(), true)) {
@@ -737,7 +749,7 @@ class Database extends ScopedMappingModel
      * @see \Propel\Common\Config\ConfigurationManager::getConfigProperty() method
      *
      * @param  string $name
-     * @return string
+     * @return string|void
      */
     public function getBuildProperty($name)
     {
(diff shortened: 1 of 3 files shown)


src/Propel/Generator/Model/Database.php
```diff
@@ -838,76 +865,12 @@ protected function setupTableReferrers()
...
-                $tableDef .= "\n    indices:\n" . implode("\n", $indices);
-            }
-
-            if ($unices) {
-                $tableDef .= "\n    unices:\n". implode("\n", $unices);
-            }
-
-            $tables[] = $tableDef;
-        }
+        $stringifiedTables = $this->tablesStringifier->stringify($this->getTables());
```
Incognito, line 868: I stole this naming convention from JSON's "stringify"

src/Propel/Generator/Model/Database.php
```diff
@@ -572,10 +589,12 @@ public function addSequence($sequence)
      */
     public function removeSequence($sequence)
     {
-        if ($this->sequences) {
-            if (false !== ($idx = array_search($sequence, $this->sequences))) {
-                unset($this->sequence[$idx]);
-            }
+        if (!$this->sequences) {
+
+            return;
```
Incognito, line 594: It made more sense to me if the "happy case" wasn't wrapped in something "conditional".

src/Propel/Generator/Model/Database.php
```diff
@@ -608,36 +627,44 @@ public function getSchemaDelimiter()
      */
     public function setSchema($schema)
     {
-        $oldSchema = $this->schema;
-        if ($this->schema !== $schema && $this->getPlatform()) {
```
Incognito, line 612: I unwrapped the contents of the IF.

src/Propel/Generator/Model/Database.php
```diff
@@ -608,36 +627,44 @@ public function getSchemaDelimiter()
...
-            $schemaDelimiter = $this->getPlatform()->getSchemaDelimiter();
-            $fixHash = function (&$array) use ($schema, $oldSchema, $schemaDelimiter) {
```
Incognito, line 614: I didn't see a solid reason for this to be a lambda so I moved it to private.

src/Propel/Generator/Model/Stringifier/TableStringifier.php
```diff
@@ -0,0 +1,117 @@
+<?php
+
+
+namespace Propel\Generator\Model\Stringifier;
+
+use Propel\Generator\Model\Table;
+
+class TableStringifier
```
Incognito, line 8: I didn't do any major rewriting of this logic, I only slightly restructured it.

src/Propel/Generator/Model/Stringifier/TableStringifier.php
```diff
@@ -0,0 +1,117 @@
+<?php
+
+
```
Incognito, line 3: I'm missing the preamble in the new files.

cristianoc72:
If it can be useful, I'm working for some time to a new object model here https://github.com/cristianoc72/propel-bls/tree/model

Basically is a backport or the new Propel 3 object model, plus a complete refactor of the schema reader (supports for yaml, json and php schemas too).
Unfortunately, the new object model breaks platforms which breaks the bilders which breaks almost everything in `Generator` namespace, so I'm fear it'll be an enormous PR 😄

Incognito:
Thanks @cristianoc72 I'll start looking into your work here, this is well aligned with something I want to do. :)

src/Propel/Generator/Model/Database.php
```diff
@@ -528,10 +539,16 @@ public function addTables(array $tables)
         }
     }
 
+    /**
+     * Removes table from this database.
+     *
+     * @param  Table|array $table
```
dereuromark, line 545: array?

Incognito:
```suggestion
     * @param  Table $table
```

It would be faster if you used this suggest changes tool.

src/Propel/Generator/Model/Database.php
```diff
@@ -608,36 +627,44 @@ public function getSchemaDelimiter()
...
+
+        $oldSchema = $this->schema;
+        $schemaDelimiter = $this->getPlatform()->getSchemaDelimiter();
+
+        $this->fixHash($this->tablesByName, $schema, $oldSchema, $schemaDelimiter);
+        $this->fixHash($this->tablesByLowercaseName, $schema, $oldSchema, $schemaDelimiter);
         parent::setSchema($schema);
     }
 
+    private function fixHash(&$array, $schema, $oldSchema, $schemaDelimiter)
```
PhilinTv, line 644: phpdoc?

src/Propel/Generator/Model/Database.php
```diff
@@ -608,36 +627,44 @@ public function getSchemaDelimiter()
...
-                        }
-                    } elseif ($oldSchema) {
-                        if (false !== strpos($k, $schemaDelimiter)) {
-                            $array[explode($schemaDelimiter, $k)[1]] = $v;
-                            unset($array[$k]);
-                        }
-                    }
-                }
-            };
+        if (!($this->schema !== $schema && $this->getPlatform())) {
```
PhilinTv, line 630: It became unnecessarily complicated for reading. 
Please open brackets or leave a comment.

src/Propel/Generator/Model/Stringifier/TablesStringifier.php
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * @license MIT License
+ */
+
+namespace Propel\Generator\Model\Stringifier;
+
+/**
+ * A class for representing a collection of Tables as a String
+ */
+class TablesStringifier
```
PhilinTv, line 16: I don't there is a big difference between TablesStringifier and TableStringifier, but spent 3mins understanding why "stringify" works with an array :)

If the difference is that small please squash them in a single class with 2 public methods.

src/Propel/Generator/Model/Database.php
dereuromark: We dont have a cs sniffer active yet. Once we do all those would be fixed by design.

src/Propel/Generator/Model/Stringifier/TablesStringifier.php
Incognito: One class has the job of representing many tables as a string, the other class has the job of representing one table as a string. They're similar but not the same. For example, the database __toString method is still messy as it is aware of how to stringify a schema instead of calling a schema stringifier, also the TableStringifer is messy because it mixes all tables concepts inside too.

The code quality score on this project will go back down when we make one class gain another reason to change.

src/Propel/Generator/Model/Database.php
Incognito: I don't understand what you mean by opening brackets. Do you want me to just simplify the boolean statement with demorgan's law like this?

```suggestion
        if ($this->schema === $schema && !$this->getPlatform()) {
```

src/Propel/Generator/Model/Stringifier/TablesStringifier.php
dereuromark: Can class docblocks be more clear in describing? Would that help?

dereuromark:
@PhilinTv ypu blocked this from moving forward 
Shall we continue here? If so, it seems this now requires your help.

Incognito:
This PR is mostly a result of research into the models, there's very little value added here.

PR #1593, opened 2020-06-30
FIX bug CLOB_EMU oracle

nizcnas:


composer.json
```diff
@@ -16,7 +16,7 @@
         }
     ],
     "require": {
-        "php": ">=7.1",
+        "php": ">=7.0.29",
```
dereuromark, line 19: We cannot Downgrade here

Please, can you instead add a test case for this?
Then we can merge.

nizcnas: set php version to 7.1

dereuromark:
Is there any way we can have a test case to back up the faulty issue and that it is now fixed?

nizcnas:
we will never enter this if
if ($column->getPDOType() == PropelTypes::CLOB_EMU) {
we have to use getType() so we can use 
bindParam instead of bindValue to pass the 4th parameter which is the length.
to reproduce the bug you must try to insert a very long text in clob column oracle database.

Issue #1192, opened 2016-05-24
MySQL Error with PDO::ATTR_EMULATE_PREPARES and Criteria::BINARY_ALL


wollanup:
I recently add attribute PDO::ATTR_EMULATE_PREPARES to my connection to get benefit of SQL types in query result (without this, all integer results are strings)

but now i'm facing this issue :

mysql client : `mysqlnd 5.0.12-dev`
php : `7.0.4-7ubuntu2`
propel version : last via composer (dev-master 9e4eb90 on 2016-05-24)

propel.yml : 

```
propel:
    database:
        connections:
            default:              
                adapter: mysql                   
                attributes:
                    ATTR_EMULATE_PREPARES: false
```

query : 
`EntityQuery->create->filterByBinaryField(3, Criteria::BINARY_ALL);`

result:

```
Propel\Runtime\Exception\PropelException
Unable to execute SELECT statement [SELECT * AS FROM table WHERE table.binary & :p1 = :p1]
File: /var/www/.../vendor/propel/propel/src/Propel/Runtime/ActiveQuery/Criteria.php
Line: 2650

Previous exception
Type: PDOException
Code:HY093
Message: SQLSTATE[HY093]: Invalid parameter number
File: /var/www/.../vendor/propel/propel/src/Propel/Runtime/Adapter/Pdo/PdoStatement.php
Line:57
Trace :
#0 /var/www/.../vendor/propel/propel/src/Propel/Runtime/Adapter/Pdo/PdoStatement.php(57): PDOStatement->execute(NULL)
#1 /var/www/.../vendor/propel/propel/src/Propel/Runtime/Connection/StatementWrapper.php(194): Propel\Runtime\Adapter\Pdo\PdoStatement->execute(NULL)
#2 /var/www/.../vendor/propel/propel/src/Propel/Runtime/ActiveQuery/Criteria.php(2644): Propel\Runtime\Connection\StatementWrapper->execute()
#3 /var/www/.../vendor/propel/propel/src/Propel/Runtime/ActiveQuery/ModelCriteria.php(1988): Propel\Runtime\ActiveQuery\Criteria->doSelect(Object(Propel\Runtime\Connection\DebugPDO))
#4 /var/www/.../vendor/propel/propel/src/Propel/Runtime/ActiveQuery/ModelCriteria.php(1024): Propel\Runtime\ActiveQuery\ModelCriteria->doSelect(Object(Propel\Runtime\Connection\DebugPDO))
```

[...]

PR #1598, opened 2020-07-03
fix(PDO): Can't prepare query with BINARY_ALL criteria when ATTR_EMUL…

wollanup:
fix(PDO): Can't prepare query with BINARY_ALL criteria when ATTR_EMULATE_PREPARE is false

see #1192


tests/Propel/Tests/Issues/Issue1598Test.php
```diff
@@ -0,0 +1,62 @@
...
+        $actualSql = $query->createSelectSql($params);
+
+        $this->assertEquals("SELECT  FROM issue_1192_item WHERE issue_1192_item.target & :p1 = :p2",
+            $actualSql,
+            'Generated SQL does not match expected SQL'
+        );
+
+        $this->assertEquals(2, count($params), 'Incorrect number of params');
+
+        $this->assertEquals($params[0], $params[1], 'Params are not identical');
```
dereuromark: Can we use assertSame() in those cases? As equals can have false positives with types.

PR #1604, opened 2020-07-07
Add PHPStan level 1 check with baseline until we can remove symf2.

dereuromark:
This is now removed from allow_failure list and part of normal travis run.

I also added docblocks to allow IDE discovery (auto complete / typehinting) as well as PHPStan to check those.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,3 +1,5 @@
+includes:
+    - phpstan-baseline.neon
 parameters:
     level: 1
     paths:
@@ -12,5 +14,9 @@ parameters:
         - '%rootDir%/../../../src/Propel/Generator/Behavior/Delegate/templates/*'
         - '%rootDir%/../../../src/Propel/Generator/Behavior/Sortable/templates/*'
         - '%rootDir%/../../../src/Propel/Generator/Behavior/NestedSet/templates/*'
+        - '%rootDir%/../../../src/Propel/Runtime/Connection/PdoConnection.php'
+        - '%rootDir%/../../../src/Propel/Generator/Command/Helper/ConsoleHelper.php'
+        - '%rootDir%/../../../src/Propel/Generator/Command/InitCommand.php'
     ignoreErrors:
         - '#Call to an undefined method .+Collection::.+Array\(\)#'
+        - '#Class .+TreeBuilder constructor invoked with 0 parameters, 1-3 required#'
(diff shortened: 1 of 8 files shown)


phpstan.neon
```diff
@@ -12,5 +14,9 @@ parameters:
         - '%rootDir%/../../../src/Propel/Generator/Behavior/Delegate/templates/*'
         - '%rootDir%/../../../src/Propel/Generator/Behavior/Sortable/templates/*'
         - '%rootDir%/../../../src/Propel/Generator/Behavior/NestedSet/templates/*'
+        - '%rootDir%/../../../src/Propel/Runtime/Connection/PdoConnection.php'
+        - '%rootDir%/../../../src/Propel/Generator/Command/Helper/ConsoleHelper.php'
+        - '%rootDir%/../../../src/Propel/Generator/Command/InitCommand.php'
     ignoreErrors:
         - '#Call to an undefined method .+Collection::.+Array\(\)#'
+        - '#Class .+TreeBuilder constructor invoked with 0 parameters, 1-3 required#'
```
Incognito, line 22: This one will be with us for a long time. Is it possible to inline some of these ignores into the file itself?

src/Propel/Generator/Behavior/Versionable/VersionableBehaviorObjectBuilderModifier.php
```diff
@@ -197,6 +278,11 @@ public function getVersion()
 ";
     }
 
+    /**
+     * @param string $script
```
Incognito, line 282: Might be worth increasing the visibility that this is by-reference. `&$script`, what do you think? Does the tool support it?

phpstan.neon
dereuromark: I dont think so. This is the best we can do for now. And doesnt hurt too much here.

src/Propel/Generator/Behavior/Versionable/VersionableBehaviorObjectBuilderModifier.php
dereuromark: This is not valid afaik. The variable is still only this. You can check by trying it, afaik phpstan or psalm will call this out as not valid docblock.

Incognito: :+1:

PR #1606, opened 2020-07-08
Remove nightly except one

dereuromark:
There is no point having 3 tests display the same fail
Once this is solved in a few months we can and will enable full coverage here again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -297,17 +297,6 @@ jobs:
         - COMPOSER=./tests/composer/composer-php8.json composer install
         - composer show
     - php: nightly
-      env: DB=mysql
-      install:
-        - rm -f ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini
-        - COMPOSER=./tests/composer/composer-php8.json composer install
-        - composer show
-    - php: nightly
-      env: DB=sqlite
-      install:
-        - rm -f ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini
-        - COMPOSER=./tests/composer/composer-php8.json composer install
-        - composer show
 
     - name: "PHP 7.4, Agnostic, Symfony ~2.7"
       php: 7.4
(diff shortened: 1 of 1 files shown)


Incognito:
The tests run entirely different parts of the code.

dereuromark:
No, they dont right now
They all fail with basic composer, exactly the same
Completely pointless to have 2 more of those in there for months to come.

Incognito:
If you add `--ignore-platform-reqs` you can see a lot of issues and get ahead of them.

.travis.yml
```diff
@@ -297,17 +297,6 @@ jobs:
         - COMPOSER=./tests/composer/composer-php8.json composer install
         - composer show
     - php: nightly
```
Incognito, line 299:
```suggestion
```

dereuromark:
Yes, you can add those then for this one build.
PHP8 shouldn't be in here in the first place as we are far ahead from getting this done.
But if one build helps to get it started, that's already fair enough. We don't need 3 right now, especially since you already added 100+ builds and that slows every commit check down by ~1h.

.travis.yml
dereuromark: There was no error - I guess it just silently skipped this one?

dereuromark:
Let us see what the issues are and we can probably switch to MySQL here, sure thing.

.travis.yml
Incognito: It's easier to understand quantum mechanics than it is to predict how travis will react to any given config.

Incognito:
I have a stale branch with some fixes but actually I think your work in PHPstan will fix more than half of them.

dereuromark:
Phpstan only fixes "static analyzer issues", not real bugs usually or?

Incognito:
A big issue that pops up in PHP8 is needing return types everywhere. The second one is required method params after an optional param. I thought 'stan was taking care of some of that for us (because we have no issues like that in Spryker)

dereuromark:
opt param is also a code sniffer issue, so we really should tackle that after phpstan level 3 is merged.

Issue #1601, opened 2020-07-06
PHP8


Incognito:
PHP8 will be here November 26, 2020, and Propel doesn't work with it right now.

dereuromark:
Closing in favor of https://github.com/propelorm/Propel2/issues/1568 PHP8 ticket

PR #1611, opened 2020-07-09
Fix php8 compatibility

dereuromark:
Fix php8 compatibility by adjusting public API of extended PDO func
https://github.com/propelorm/Propel2/issues/1601


.travis.yml
```diff
@@ -9,7 +9,6 @@ php:
   - 7.2
   - 7.3
   - 7.4
-  - nightly
```
dereuromark: Why removing nightly (php8)?

dereuromark: I would instead remove 7.3 and 7.4 to save some workers until both 7.2 and 8.0 work.

bezpiatovs: It's temporary) I commented out almost all combinations to faster tests passing. Since agnostic tests became green I will return all things back.

composer.json
```diff
@@ -31,7 +31,7 @@
         "ext-json": "*",
         "monolog/monolog": "^1.3",
         "phpstan/phpstan": "^0.12.4",
-        "phpunit/phpunit": "^7.5.15",
+        "phpunit/phpunit": "9.4.2",
```
dereuromark, line 34: Best to say 8||9

src/Propel/Runtime/Connection/PdoConnection.php
```diff
@@ -50,28 +68,29 @@ public function getName()
...
         if (is_array($options)) {
             foreach ($options as $key => $option) {
-                $index = (is_numeric($key)) ? $key : constant('self::' . $key);
+                $index = (is_numeric($key)) ? $key : constant('\PDO::' . $key);
                 $pdoOptions[$index] = $option;
             }
         }
 
-        parent::__construct($dsn, $user, $password, $pdoOptions);
+        //var_dump($dsn, $user, $password, $pdoOptions);
```
gechetspr, line 82: please remove this

src/Propel/Runtime/Connection/PdoConnection.php
```diff
@@ -168,8 +191,23 @@ public function prepare($statement, $driver_options = null)
      *
      * @return string
      */
-    public function quote($string, $parameterType = PDO::PARAM_STR)
+    public function quote($string, $parameter_type = \PDO::PARAM_STR)
+    {
+        return $this->pdo->quote($string, $parameter_type);
+    }
+
+    public function beginTransaction()
```
gechetspr, line 199: would be nice to have docblock here

src/Propel/Runtime/Connection/StatementWrapper.php
```diff
@@ -86,11 +86,10 @@ public function prepare($options)
      */
     public function query()
     {
-        /** @var \PDOStatement $statement */
-        $statement = $this->connection->getWrappedConnection()->query($this->sql);
-        $this->statement = $statement;
+        $wrapped_connection = $this->connection->getWrappedConnection();
```
gechetspr, line 89: $wrappedConnection

tests/Propel/Tests/Common/Config/ConfigurationManagerTest.php
```diff
@@ -153,14 +153,13 @@ public function testUnsupportedExtensionsAreIgnored()
...
     /**
-     * @expectedException \Propel\Common\Config\Exception\InvalidArgumentException
-     *
-     * @exceptionMessage Propel expects only one configuration file
-     *
      * @return void
      */
     public function testMoreThanOneConfigurationFileInSameDirectoryThrowsException()
     {
+        $this->expectException(\Propel\Common\Config\Exception\InvalidArgumentException::class);
```
gechetspr, line 160: Please move this to the use statement

tests/Propel/Tests/Common/Config/ConfigurationManagerTest.php
```diff
@@ -176,14 +175,13 @@ public function testMoreThanOneConfigurationFileInSameDirectoryThrowsException()
...
     /**
-     * @expectedException \Propel\Common\Config\Exception\InvalidArgumentException
-     *
-     * @exceptionMessage Propel expects only one configuration file
-     *
      * @return void
      */
     public function testMoreThanOneConfigurationFileInDifferentDirectoriesThrowsException()
     {
+        $this->expectException(\Propel\Common\Config\Exception\InvalidArgumentException::class);
```
gechetspr, line 182: Please check all places in this file

gechetspr (changes requested):
A few minor thing to fix

src/Propel/Runtime/Connection/StatementWrapper.php
bezpiatovs: done

src/Propel/Runtime/Connection/PdoConnection.php
bezpiatovs: done

src/Propel/Runtime/Connection/PdoConnection.php
bezpiatovs: done

.travis.yml
```diff
@@ -6,9 +6,9 @@ git:
 dist: bionic
 
 php:
-  - 7.2
```
dereuromark, line 9: Why dropping PHP 72?
I dont think we should do that at this point if we dont have to.
Rather remove 7.3, so the amount of jobs stays the same.

bezpiatovs: Done. Your right, despite on official 7.2 EOL, users are still coupled on this version. Also, PHPUnit 9 is strict >=7.3, so added 8||9 as you mentioned before too.

dereuromark:
You added nightly as global one, so far it runs only
- DB=agnostic
- DB=mysql
- DB=sqlite

without specific version

Can we also add nightly with at least min per DB version?
e.g.
MARIADB=10.1
MARIADB=10.4 DB_USER=travis
etc?

Just to make sure PHP8 also works with each of those specific versions?

We could also remove some jobs from here, e.g. 7.3 runs (since 7.2 + 7.4 cover enough for us) to make up for it.

dereuromark:
Can we also run psalm over the code base? It usually finds a few more issues than phpstan.
E.g. "ParamNameMismatch" as this is an important PHP8 check for method param name API.

stereomon (changes requested):

tests/Propel/Tests/Common/Config/Loader/JsonFileLoaderTest.php
```diff
@@ -95,15 +98,15 @@ public function testJsonFileIsEmpty()
...
-     * @expectedException \Propel\Common\Config\Exception\InputOutputException
-     * @expectedExceptionMessage You don't have permissions to access configuration file notreadable.json.
-     *
      * @requires OS ^(?!Win.*)
      *
      * @return void
      */
     public function testJsonFileNotReadableThrowsException()
     {
+        $this->expectException(InputOutputException::class);
```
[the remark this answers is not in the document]
bezpiatovs: Created followup ticket.

src/Propel/Runtime/Collection/CollectionIterator.php
```diff
@@ -229,20 +229,26 @@ public function append($value)
     }
 
     /**
+     * @param int $sort_flags
+     *
      * @return void
      */
-    public function asort()
+    public function asort($sort_flags = SORT_REGULAR)
```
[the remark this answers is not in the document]
bezpiatovs: in this case we probably can't get rid of this.
because this function overrides native `\ArrayIterator` and in PHP8 there is strict signature compatibility needed.

src/Propel/Runtime/Connection/ConnectionWrapper.php
```diff
@@ -231,7 +231,7 @@ public function commit()
         $return = true;
         $opcount = $this->nestedTransactionCount;
 
-        if ($opcount > 0) {
+        if ($opcount > 0 && $this->inTransaction()) {
```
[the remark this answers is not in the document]
bezpiatovs: It is. But in our case it's a bit about BC:
- in PHP 7.x there was a bug, that in case if you are inside transaction and execute any DDL query, you will still be inside transaction (that is not OK and controversial with docs), but most probably it just relates to a PDO internal "flag", while transaction is already auto-commited on DB level;
- in PHP 8 this bug was fixed, and since you call any DDL query, you will be auto-commited without any notices.
- in case we try to commit or rollback while not in transaction, it will lead to an exception.

Since we discovered such behavior only on PHP8, in this case, I suggest, we need to check whether we are `inTransaction` after every single query executed (ideally), and react somehow on it's change. In our case, it means we need to probably re-think all `$opcount` mechanism since we can't really rely on it anymore I assume. This approach is just to be BC with older versions of PHP.

@dereuromark is aware of this situation. From my perspective, we need a followup ticket here.

src/Propel/Runtime/Connection/StatementInterface.php
```diff
@@ -0,0 +1,197 @@
...
+     * @return bool TRUE on success or FALSE on failure.
+     */
+    public function setAttribute($attribute, $value);
+
+    /**
+     * Retrieve a statement attribute
+     *
+     * @param int $attribute
+     *
+     * @return mixed the attribute value.
```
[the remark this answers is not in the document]
bezpiatovs: in this case I just took signature from docs:
https://www.php.net/manual/ru/class.pdo.php

src/Propel/Runtime/Connection/StatementInterface.php
```diff
@@ -0,0 +1,197 @@
...
+     *
+     * @param string $className Name of the created class.
+     * @param array $ctorArgs Elements of this array are passed to the constructor.
+     *
+     * @return mixed
+     */
+    public function fetchObject($className = 'stdClass', array $ctorArgs = []);
+
+    /**
+     * Fetch the SQLSTATE associated with the last operation on the statement handle
```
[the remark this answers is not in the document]
dereuromark: We can do more such minor cleanup in follow up PRs, but we need to focus on shipping this into master asap.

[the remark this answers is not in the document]
bezpiatovs: Anyway I fixed it already. Will push soon

src/Propel/Runtime/Connection/ConnectionWrapper.php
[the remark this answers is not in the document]
dereuromark: Yeah, I just hope we don't introduce other issues with it.

src/Propel/Generator/Util/QuickBuilder.php
```diff
@@ -265,7 +265,7 @@ public function buildSQL(ConnectionInterface $con)
             }
             try {
                 $stmt = $con->prepare($statement);
-                if ($stmt instanceof PDOStatement) {
+                if ($stmt instanceof StatementWrapper) {
```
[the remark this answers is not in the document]
bezpiatovs: fixed

src/Propel/Runtime/Connection/ConnectionInterface.php
```diff
@@ -148,8 +148,7 @@ public function transaction(callable $callable);
      * @param string $statement The SQL statement to prepare and execute.
      *                          Data inside the query should be properly escaped.
      *
-     * @return int The number of rows that were modified or deleted by the SQL
-     *             statement you issued. If no rows were affected, returns 0.
+     * @return mixed The number of rows that were modified or deleted.
```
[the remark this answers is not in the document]
bezpiatovs: fixed

src/Propel/Runtime/Connection/PdoConnection.php
```diff
@@ -24,6 +24,24 @@ class PdoConnection extends PDO implements ConnectionInterface
      */
     protected $name;
 
+    /**
+     * @var \PDO
+     */
+    protected $pdo;
+
+    /**
+     * Forward any call to a method not found to the proxied connection.
```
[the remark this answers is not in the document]
bezpiatovs: fixed

src/Propel/Runtime/Connection/ConnectionWrapper.php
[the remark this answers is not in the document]
bezpiatovs: Created followup ticket.

stereomon (approved):

demkos (approved):

PR #1595, opened 2020-07-01
Fix symfony deps [sf/* ^5, sf/validator ^4]

Incognito:
Shims, Polly-fills, and in general reconciles issues with all major versions of symfony, PHP, and DB Adapter.

There's one outstanding issue with symfony5 validators as they change behaviours in a non-BC way. 

Addresses:
- https://github.com/propelorm/Propel2/issues/1578
- https://github.com/propelorm/Propel2/issues/1575
- https://github.com/propelorm/Propel2/issues/1574 (Drops <2.7) 
- https://github.com/propelorm/Propel2/issues/1573
- https://github.com/propelorm/Propel2/issues/1534

Some new tickets will need to be generated if this PR is accepted.

- [ ] Fix sf5 validators
- [ ] Refactor `ValidateBehavior::addLoadValidatorMetadataMethod` to cleanup class selection precedence. Would require new tests.
- [ ] Add a non-interactive version of InitCommandTest
- [ ] PHP8 fixes
- [ ] Console exit codes should use constants to improve readability

# Test results

<table>
<thead>
  <tr>
    <th></th>
    <th colspan="5">PHP 7.1</th>
    <th colspan="5">PHP 7.2</th>
    <th colspan="5">PHP 7.3</th>
    <th colspan="5">PHP 7.4</th>
    <th>PHP 8</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td></td>
    <td>Maria<br></td>
    <td>MySql</td>
    <td>Postgres</td>
    <td>Sqlite</td>
    <td>Agnostic</td>
    <td>Maria</td>
    <td>MySql</td>
    <td>Postgres</td>
    <td>Sqlite</td>
    <td>Agnostic</td>
    <td>Maria</td>
    <td>MySql</td>
    <td>Postgres</td>
    <td>Sqlite</td>
    <td>Agnostic</td>
    <td>Maria</td>
    <td>MySql</td>
    <td>Postgres</td>
    <td>Sqlite</td>
    <td>Agnostic</td>
    <td>Agnostic</td>
  </tr>
  <tr>
    <td>Symfony ~2.7</td>
    <td>✓</td>
    <td>✓</td>
    <td>✓</td>
    <td>✓</td>
    <td>✓</td>
    <td>✓</td>
    <td>✓</td>
    <td>✓</td>
    <td>✓</td>
    <td>✓</td>
    <td>✓</td>
    <td>✓</td>
    <td>✓</td>
    <td>✓</td>
    <td>✓</td>
[...]

PR #1617, opened 2020-07-16
Fix return value for "no migration needed" case

dereuromark:
https://github.com/propelorm/Propel2/pull/1595/files#diff-1ccdce29404b8984d1b2101f0069a564R84
it looks like this change was "correct in the sense of what was before and the transition towards int",
as false means 1 (error)

But if you look more deeply the already existing false should have been true (and thus 0) as return value.
This code, if no migration has to be done, should actually say "all good" and not report an error here.

//cc @Incognito


dereuromark:
> Does line 116 also need this change?

=> $output->writeln('<error>preUp() returned false. Aborting migration.</error>');

So no, I think that one line so far is the only wrong return.

Incognito:
:shipit:

PR #1625, opened 2020-07-21
always create unique indices by constraint (for pgsql platform)

daniel-rose:
Unique indices are currently generated in two different ways:

- "getAddTableDDL" creates it over an unique constraint

- "getAddIndexDDL" creates it without unique constraint

The Problem is, that the method "getDropIndexDDL" will remove unique indices always by constraint.

### DIFF
--- a/tests/Propel/Tests/Generator/Platform/PgsqlPlatformTest.php
+++ b/tests/Propel/Tests/Generator/Platform/PgsqlPlatformTest.php
@@ -601,6 +601,17 @@ CREATE INDEX "babar" ON "foo" ("bar1","bar2");
     }
 
     /**
+     * @dataProvider providerForTestGetUniqueIndexDDL
+     */
+    public function testAddUniqueIndexDDL($index)
+    {
+        $expected = '
+ALTER TABLE "foo" ADD CONSTRAINT "babar" UNIQUE ("bar1");
+';
+        $this->assertEquals($expected, $this->getPlatform()->getAddIndexDDL($index));
+    }
+
+    /**
      * @dataProvider providerForTestGetIndicesDDL
      */
     public function testAddIndicesDDL($table)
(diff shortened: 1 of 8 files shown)


dereuromark:
Hi!
Is there any way we can introduce a unit test for this to back up the functional change?

daniel-rose:
@dereuromark this is not an improvement. It is a bugfix. If you want to delete an unique key which is created by "getAddIndexDDL", the migration will be failed.

dereuromark:
You are right, I adjusted it.

In the long run we should find an easier way to diff rows, excluding whitespace changes..
The 4 fails are at least partially around that topic.
Maybe we can find a way for now to make those tests green.

tests/Propel/Tests/Generator/Platform/PlatformMigrationTestProvider.php
```diff
@@ -198,6 +198,7 @@ public function providerForTestGetModifyTableIndicesDDL()
     <table name="foo">
         <column name="id" primaryKey="true" type="INTEGER" autoIncrement="true" />
         <column name="bar" type="INTEGER" />
+        <column name="bax" type="VARCHAR" size="12" required="true" />
```
stereomon, line 201: Usually, these names should be more speaking but that would be too much this PR. Can you please follow foo, bar, baz, baz naming here, and move the new column after the `baz` column?

tests/Propel/Tests/Generator/Platform/PlatformTestProvider.php
```diff
@@ -247,6 +247,22 @@ public function providerForTestGetIndexDDL()
         ];
     }
 
+    public function providerForTestGetUniqueIndexDDL()
```
stereomon, line 250: Please add a DocBlock with return annotation.

tests/Propel/Tests/Generator/Platform/PgsqlPlatformTest.php
```diff
@@ -600,6 +600,17 @@ public function testAddIndexDDL($index)
         $this->assertEquals($expected, $this->getPlatform()->getAddIndexDDL($index));
     }
 
+    /**
+     * @dataProvider providerForTestGetUniqueIndexDDL
+     */
+    public function testAddUniqueIndexDDL($index)
```
stereomon, line 606: Please add a type hint (Index) and update the DocBlock accordingly.

PR #1631, opened 2020-07-23
add json to native pgsql type

### LINKED ISSUE #1630: [Bug][PostGres][json] diff detects change of column type varchar to json
creating a diff of a database with column type json detects a change from varchar to json.

this is because the `src/Propel/Generator/Reverse/PgsqlSchemaParser.php:288` does not know `json` as a type native Pgsql type and therefore uses `varchar` as a fallback

tienbuide:
fixes #1630


dereuromark:
Do we have to also check other DBs here?

Issue #959, opened 2015-07-01
propel sql:insert ignore ['settings']['queries'] database config parameter


Uvadzucumi:
Example:
I have "set names" query in my mysql connection config, and add initial sql data queries file in generated sqldb.map file. additional sql file contained data in utf-8 codepage and incorrectly added after run propel sql:inisert command. For fix this problem I add dirty hack in src/Propel/Generator/Manager/SqlManager.php file. method insertSql() after lile 170
part of my config & fix: https://gist.github.com/Uvadzucumi/961ad3cde0be3815fc36

marcj:
I see, that should actually happen in the construction of the PDO instance.

djavorek:
This is still valid. Is there any workaround or do I have to use this hack?
I am using 2.0@dev

dereuromark:
@djavorek Do you have a fix you can PR to this repo? We are happy to review and merge in any contributions here.

djavorek:
Will do it later this day.

PR #1641, opened 2020-08-03
Run config queries on sql:insert

djavorek:
Included solution, which was proposed in [Issue 959 - propel sql:insert ignore ['settings']['queries'] database config parameter](https://github.com/propelorm/Propel2/issues/959).

It needs to be checked and tested. If anything should be changed, please write a comment.
These changes would be important for me, as I can't insert UTF-8 content via sql:insert.

### DIFF
--- a/src/Propel/Generator/Manager/SqlManager.php
+++ b/src/Propel/Generator/Manager/SqlManager.php
@@ -187,6 +187,7 @@ class SqlManager extends AbstractManager
             }
 
             $con = $this->getConnectionInstance($database);
+            
             if (isset($this->connections[$database]['settings']['queries']) &&
             count($this->connections[$database]['settings']['queries'])) {
                 $initQueries = $this->connections[$database]['settings']['queries'];
@@ -200,7 +201,7 @@ class SqlManager extends AbstractManager
                             throw new Exception($message, 0, $e);
                         }
                     }
-                }
+                });
             }
             
             $con->transaction(function () use ($con, $sqls) {
--- a/src/Propel/Generator/Manager/SqlManager.php
+++ b/src/Propel/Generator/Manager/SqlManager.php
@@ -187,6 +187,22 @@ class SqlManager extends AbstractManager
             }
 
             $con = $this->getConnectionInstance($database);
+            if (isset($this->connections[$database]['settings']['queries']) &&
+            count($this->connections[$database]['settings']['queries'])) {
+                $initQueries = $this->connections[$database]['settings']['queries'];
+                $con->transaction(function () use ($con, $initQueries) {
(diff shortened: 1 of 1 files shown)


djavorek:
One MariaDB CI check fails, others pass. Can someone check this?

dereuromark:
We might also need a test case here for this specific case.

djavorek:
Well, it will take some time for me, to create a correct one. I'll check on it when I can.

mringler:
Hmm, usually the queries in settings should be run automatically when the connection is initialized. There should be no need to add the initialization code somewhere else a second time.
What database are you using? Which adapter is set in propel.json?

djavorek:
I was using MySQL 5.6, however I am not sure about what adapter I was using.

mringler:
I looked at the MySQL adapter, but it seems to work as expected:

`PdoAdapter::initConnection()` is where the additional queries are run:
```php
public function initConnection(ConnectionInterface $con, array $settings)
{
    if (isset($settings['charset'])) {
        $this->setCharset($con, $settings['charset']);
    }

    if (isset($settings['queries']) && is_array($settings['queries'])) {   // this part is duplicated by this PR
        foreach ($settings['queries'] as $query) {
            $con->exec($query);
        }
    }
}
```
Adding a `debug_print_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);` into `initConnection()` shows how a call to `sql:insert` leads to a call to that function:

```
#0  Propel\Runtime\Adapter\Pdo\PdoAdapter->initConnection() called at [src/Propel/Runtime/Adapter/Pdo/PdoAdapter.php:70]
#1  Propel\Runtime\Adapter\Pdo\PdoAdapter->getConnection() called at [src/Propel/Runtime/Connection/ConnectionFactory.php:41]
#2  Propel\Runtime\Connection\ConnectionFactory::create() called at [src/Propel/Generator/Manager/SqlManager.php:225]
#3  Propel\Generator\Manager\SqlManager->getConnectionInstance() called at [src/Propel/Generator/Manager/SqlManager.php:188]
#4  Propel\Generator\Manager\SqlManager->insertSql() called at [src/Propel/Generator/Command/SqlInsertCommand.php:69]
#5  Propel\Generator\Command\SqlInsertCommand->execute() called at [vendor/symfony/console/Command/Command.php:256]
...

So calling `sql:insert` leads to a call to `initConnection()`, which does, what this PR does a second time. Not sure why it was not working for you. If you have your project still available, can you do a similar check as above, to see if `initConnection()` is called?

dereuromark:
@djavorek ping

djavorek:
Based on the copied code, this should be ok. I can't validate this easily, but I beleive it is fixed.

Issue #1568, opened 2020-06-23
PHP8 Incompatibilities [PHP]


Incognito:
Related to: https://github.com/propelorm/Propel2/issues/1541

Visible in travis “nightly php release” builds. The first obvious issues look like this:

```
Deprecated: Method ReflectionParameter::getClass() is deprecated in /home/travis/build/propelorm/Propel2/vendor/phpunit/phpunit/src/Framework/MockObject/MockMethod.php on line 308
PHP Deprecated:  Method ReflectionParameter::getClass() is deprecated in /home/travis/build/propelorm/Propel2/vendor/phpunit/phpunit/src/Framework/MockObject/MockMethod.php on line 308
```

And

```
Deprecated: Required parameter $modifiedClause follows optional parameter $columnPhpName in /home/travis/build/propelorm/Propel2/tests/Propel/Tests/Runtime/ActiveQuery/ModelCriteriaWithSchemaTest.php on line 58
```

Incognito:
This will likely require an update to Phpunit, which means supported versions of PHP would become incompatible with tests.

dereuromark:
We should try to upgrade phpunit here to 8+
As this is require-dev this shouldn't have a direct impact on projects using this library.

Incognito:
If you can drop php 7.1 you unlock phpunit 8.

dereuromark:
Done in https://github.com/propelorm/Propel2/pull/1618
So that unlocks a follow up PR then here.

dereuromark:
Is anyone interested in further getting this repo PHP 8 ready?
you can either built on top of https://github.com/propelorm/Propel2/pull/1611 - or use a separate branch here.

dereuromark:
@propelorm/propel2 
Pease review https://github.com/propelorm/Propel2/pull/1611 as PR for PHP 8 compatibility now.

We can then release soon a new version.

dereuromark:
We are going to merge the PHP 8 support this week.
Everyone: Please do a check on it if possible.

PR #1645, opened 2020-08-06
Remove is_resource() as this will always be false in PHP 8

dereuromark:
See PHP8 on resource to object changes

This requires verification if this still works fine for both PHP 7 and 8.

Refs https://github.com/spryker/code-sniffer/pull/228

### DIFF
--- a/src/Propel/Generator/Builder/Om/ObjectBuilder.php
+++ b/src/Propel/Generator/Builder/Om/ObjectBuilder.php
@@ -1046,7 +1046,7 @@ abstract class " . $this->getUnqualifiedClassName() . $parentClass . ' implement
         }
 
         $script .= "
-        if (null == \$this->$cloUnserialized && \$this->$clo !== false) {
+        if (null == \$this->$cloUnserialized && \$this->$clo !== null && \$this->$clo !== false) {
             if (\$serialisedString = stream_get_contents(\$this->$clo)) {
                 \$this->$cloUnserialized = unserialize(\$serialisedString);
             }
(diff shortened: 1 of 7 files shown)


dereuromark:
Does anyone of the community here have ideas on how to tackle this?
We  should start preparing some of the next releases here already in a way that PHP8 won't be a big BC break again end of this year.

dereuromark:
@gechetspr This doesn't work, we need a fix here that works for PHP7 and 8 somehow.
Maybe this requires a larger refactor so that we don't have to check the type but can rely on it being a resource?

dereuromark:
@propelorm/propel2 Does anyone have an idea on how to solve this PHP 8 blocker?

staabm:
@dereuromark not sure what the actual problem is.. what is different in PHP8? whats the actual change in php8 which triggers a problem here?

dereuromark:
In PHP8 resources are objects, so that check will never work (always returns false).

staabm:
Most of the mentioned code see to work on streams which IMO are still resources.

Do you have a concrete use-case which no longer works?

dereuromark:
It is possible that with the latest PHP releases they shimmed it back in a way, that we can keep the current method.
If that's the case, great.

Then we just need to solve the other more problematic PHP 8 incompatibilities (mainly around typehints).

dereuromark:
I will close this then for now in favor of linking to https://github.com/propelorm/Propel2/issues/1568
And we will see how we can progress from there.

PR #1647, opened 2020-08-07
Update Profiler.php

acabutto:
When getProfileBetween() is invoked before start() an undefined index error raises when trying to access $startSnapshot['microtime'] or $endSnapshot['microtime'].

### DIFF
--- a/src/Propel/Runtime/Util/Profiler.php
+++ b/src/Propel/Runtime/Util/Profiler.php
@@ -235,7 +235,12 @@ class Profiler
     public function getProfileBetween($startSnapshot, $endSnapshot)
     {
         $profile = '';
-
+        
+        // it may happen that $start/$endSnapshot might be undefined when getProfileBetween() is called 
+        // leading to an undefined index error two lines below, this is just a workaround. More investigation is needed
+        if (!isset($endSnapshot['microtime'], $startSnapshot['microtime']))
+            return $profile;
+        
         if ($this->slowTreshold) {
             if ($endSnapshot['microtime'] - $startSnapshot['microtime'] >= $this->slowTreshold) {
                 $profile .= 'SLOW ';


dereuromark:
Can you please check the test failures? https://travis-ci.org/github/propelorm/Propel2/jobs/715807964#L539

SpikedCola:
We are also affected by these notices. 

The tests appear to fail for two reasons: 
- start() isn't called, so $startSnapshot['microtime'] is not set.
- the test runs `getProfileBetween([], ['memoryUsage' => 343245])` , without specifying microtime in the stubbed $endSnapshot.

So, if microtime isn't specified by the test, it seems like the failure is actually correct? However, this doesn't account for the lack of ->start()

dereuromark:
@SpikedCola Can you create an alternative PR please?
This once seems stale.

SpikedCola:
Should I open an issue instead? This appears to go beyond 'microtime', it also happens with the 'memoryUsage' key (again, because $startSnapshot is empty, all checks will fail). Submitting a PR that just adds isset() checks (so the test cases pass) doesn't seem like the right solution to me.

dereuromark:
Either way works.

Technowl:
I think this error is related to PHP 7.4.
https://www.php.net/manual/en/migration74.incompatible.php "Array-style access of non-arrays"

I was in the process of upgrading our project from using php 7.0 to 7.4 while at the same time upgrading from a very old (2016) version of Propel2 to the latest version, and both the 2016 and the latest versions of Propel gave this or similar error messages.

That means that the Profiler has had this flaw where it checks for slow operations without a start snapshot for quite a while.

If, like me, that never bothered you in the past and all you care about is PHP 7.4 compatibility, then the change in this PR does the trick.
I'd recommend merging this or a similar PR as soon as possible, and then dealing with the underlying flaw with there being no start snapshot without the need for haste, so that at least it's not broken in PHP 7.4.

dereuromark:
Please rebase or make a new PR.

mringler:
**tl;dr**: The solution in this PR is the correct solution for the problem (though the condition is wrong, `$endSnapshot` is never empty).

From what I understand, `Profiler.php` has to be initialized with `start()`, which stores start time ('microtime'), initial memory usage and memory peak usage. For every log message, the difference between the initial values and the current status is printed as a header:
```
 [2021-03-17 18:57:39]      Time:  0.236ms | Memory:    900kB | Memory Delta:  +16.9kB | Memory Peak:    901kB | SELECT ...
```
The `ProfilerConnectionWrapper` class wraps the database connection object (`ConnectionInterface`), and starts the Profiler when `prepare()`, `query()` or `exec()` is called, in order to log query statistics with execution time and memory usage after execution.

However, during construction of the `ConnectionWrapper` (parent class of `ProfilerConnectionWrapper`), it tries to write a log message itself:
```php

    /**
     * Creates a Connection instance.
     *
     * @param \Propel\Runtime\Connection\ConnectionInterface $connection
     */
    public function __construct(ConnectionInterface $connection)
    {
        $this->connection = $connection;
        if ($this->useDebug) {
            $this->log('Opening connection');
        }
    }
```
At that point, `Profiler::start()` has not been called, so when accessing the start time and memory values, they are not there, causing the warning.

There are four ways this can be fixed:
- initialize the Profiler with start values
- return empty header if no start values are set (this PR does that, by checking if `$startSnapshot['microtime']` is set)
- remove the 'Opening connection' message
- use `$this->getLogger()->info('Opening connection);` in the constructor of `ConnectionWrapper`

Some observations:
- The `ConnectionWrapper` is set to swallow all log messages not triggered by `exec`, `query` or `execute` (see `ConnectionWrapper::log()`. So its own  'Opening connection' message does not make it to the log
- Once set, the stored start values are never removed, only reset. If they are not reset, profiler gives difference since last `start()`, i.e. during log of `ProfilerConnectionWrapper::__destuct()` (which is also swallowed).
- The connection is given to the wrapper as an argument. Even if we wanted to measure time and memory usage during connection initialization, we couldn't, since it is already opened when the wrapper sees it.

Personally, I would remove the 'Opening connection' message, as it doesn't do anything anyway (unless a child class overrides the `log` function). Initializing the Profiler with start values or returning an empty header is a matter of taste, both have up- and downsides.

SpikedCola:
@mringler thanks for detailing the issue so clearly! Removing the 'Opening connection' message makes sense to me.

mringler:
Hey @SpikedCola, thank you for the response! This place feels like a graveyard sometimes - glad someone else out there survived the zombie apocalypse.
I have created a new PR #1716 for this. If you have a minute to review and request changes or approve, it would help getting this finished. No worries if you don't, of course.
Best, Moritz

dereuromark:
Replaced by https://github.com/propelorm/Propel2/pull/1716
Please approve the other PR.

PR #1652, opened 2020-09-02
Updates Generator Model Table composite primary check to check null

SourceCode:
I have noticed an issue in propel/propel/src/Propel/Generator/Model/Table.php - involving hasCompositeKey and when a table is missing it, it never makes it to its exception (foreignTable is null).

if ($foreignTable->hasCompositePrimaryKey()) is the check that is the issue in the scenario. I am going to try and send a MR in for it, but its pretty edge - but debugging it required diving as the exception / error was useless to indicate where it happens.

With the notNull assertion check this ensures this does not fail, and allows propel to proceed to the needed exception which results in: 

`[Propel\Generator\Exception\BuildException]
Table "my_table" contains a foreign key to nonexistent table "non_existant_table"`

To reproduce create a table with a foreign key to a table that does not exist. The resulting Exception is the failure in the code, not the failure of processing the schema or what the cause may be.

### DIFF
--- a/src/Propel/Generator/Model/Table.php
+++ b/src/Propel/Generator/Model/Table.php
@@ -921,7 +921,7 @@ class Table extends ScopedMappingModel implements IdMethod
 
             // check for incomplete foreign key references when foreign table
             // has a composite primary key
-            if ($foreignTable->hasCompositePrimaryKey()) {
+            if ($foreignTable !== null && $foreignTable->hasCompositePrimaryKey()) {
                 // get composite foreign key's keys
                 $foreignPrimaryKeys = $foreignTable->getPrimaryKey();
                 // check all keys are referenced in foreign key


SourceCode:
Reviewing deeper I am closing this PR - however a major problem still exists, what happens is that in a large system, with many tables, if a table is removed but its foreign key is not, what you will result with is a debug output that begins at where the problem as occurued.

What this means is when I have the following:
A - Foreign key to C
B 
C - foreign key to B

Then:
A - Foreign key to C
B  - {Deleted}
C - foreign key to B

In the debug output, just table C will show - having 'reset the list of tables to process' starting with the table after the issue.

How this is fixed I have not figured out yet, but once you are in a very large database setup it can be difficult to isolate what happened. 

(I am moving a large schema over to a new iteration of an application and removed what I thought was an orphaned table - I was wrong, and it caused 2 hours of debugging or more lol)

SourceCode:
`Fatal error: Uncaught Error: Call to a member function hasCompositePrimaryKey() on null in /var/www/html/vendor/propel/propel/src/Propel/Generator/Model/Table.php on line 791`

Is there error that you will receive and it will be ambiguous as to the cause and my solution does not resolve it.

SourceCode:
This issue is very strange as it seems to only be affected in WSL2 + Docker on Windows 10.

What I did to verify was build the same project by deleting, pulling, rebuilding.

- WSL2 + Ubuntu 18 via /mnt/ - Builds Pass
- WSL2 + Docker on Alpine - Build Fail
- MacOS + Docker on Alpine - Build Pass

Always complains about hasCompositePrimaryKey and unable to locate the cause 😢

SourceCode:
Well this one is insane, but here it is:
https://github.com/docker/for-win/issues/4824

Essentially when crossing the WSL2 bridge you must be mindful of directory permissions as you will receive an exit code 255 as windows will expose the filesystem as 777 permissions, but once WSL2 actually accesses the files on the windows file system the permissions are different and will reject and not provide any reason to you the console.

dereuromark:
Thank you for the findings.

Your code change might still be valid in terms of pure PHP.
Since the value can be null we shouldnt blindly check on it causing a fatal error.
Better to throw a meaningful exception for example in that case.

PR #1654, opened 2020-09-09
Fixes transaction handling when \Throwable is thrown

### LINKED ISSUE #1653: Improper transaction handling when \Throwable is thrown
```php
$con = Propel::getConnection('some_connection');

try
{
	$con->transaction(function() {
		throw new \Error('boom');
	});
}
// my code is handling both errors and exceptions and can continue doing its work ...
catch (\Throwable $throwable)
{
	// .. yet we're now in a never-ending transaction
	assert(false === $con->inTransaction()); // this assertion fails
}
```

prgTW:
fixes #1653

Situation that is being fixed:
```php
$con = Propel::getConnection('some_connection');

try
{
	$con->transaction(function() {
		throw new \Error('boom');
	});
}
// my code is handling both errors and exceptions and can continue doing its work ...
catch (\Throwable $throwable)
{
	// .. yet we're now in a never-ending transaction
	assert(false === $con->inTransaction()); // this assertion fails
}
```

### DIFF
--- a/src/Propel/Common/Util/SetColumnConverter.php
+++ b/src/Propel/Common/Util/SetColumnConverter.php
@@ -43,7 +43,7 @@ class SetColumnConverter
             $bitValueArr[array_search($value, $valueSet)] = '1';
         }
 
-        return base_convert(implode(array_reverse($bitValueArr)), 2, 10);
+        return base_convert(implode('', array_reverse($bitValueArr)), 2, 10);
     }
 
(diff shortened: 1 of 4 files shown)


dereuromark (approved):

dereuromark:
Can someone else please also review and approve so we can merge this?

demkos (approved):

PR #1657, opened 2020-10-09
Add phpstan template annotations to collection/criteria lib objects

nederdirk:
To Do: annotations in generated classes


dereuromark:
Nice. this looks quite useful :)

nederdirk:
> Nice. this looks quite useful :)

Thank you for your encouragement. I noticed that some of the types in our application code didn't resolve well, so I set out to solve the upstream instead of adding a lot of rules to our baseline

dereuromark:
Do you plan on adding more? As your current PR is still a draft.

nederdirk:
> Do you plan on adding more? As your current PR is still a draft.

I do, and I plan on continuing this work on Thursday 15 October

nederdirk:
Dear @dereuromark,

while I was busy with this, I ran into this problem: `ModelCriteria` objects have a property `$formatter` (inherited from `BaseModelCriteria`) which determines the output type of all functions querying the database backend. One of the possible formatters is `ArrayFormatter`, which returns the records as an `ArrayCollection<array<string,mixed>>`. This poses a problem for `phpstan`, as it (correctly) can't unify `ArrayCollection<array<string,mixed>>` with `Collection<T of ActiveRecordInterface>`.

What would you think is the best way for me to go forward with this PR? I don't want to add type signatures that lie about the parameter or return types, but having `find()` return `Collection<T>|Collection<array<string,mixed>>` defeats the purpose of this PR; it would still result in a lot of assertions in calling code, thus not effectively increasing the value of static analysis.

dereuromark:
I would try to keep the existing param/return as is, and instead focus on adding those as `phpstan-*` additions on top.
This way they are optional and also helpful for our CI validation runs.

src/Propel/Runtime/Collection/OnDemandCollection.php
```diff
@@ -247,7 +256,7 @@ public function exchangeArray($input)
      *
      * @return array
      */
-    public function getArrayCopy()
+    public function getArrayCopy($keyColumn = null, $usePrefix = false)
```
nederdirk, line 259: This makes `OnDemandCollection::getArrayCopy` compatible with its parent's `ObjectCollection::getArrayCopy`

nederdirk (commented):
@dereuromark Thank you for your quick reply! I don't fully understand what you mean, and I have tried to be meticulous with not touching existing `@var`-annotations or syntax type hints unless absolutely necessary, or where the existing `@var`-annotations were incorrect. Can you perhaps point out specific instances that keep you from considering this PR for merging?

src/Propel/Runtime/ActiveQuery/Criteria.php
```diff
@@ -186,7 +186,7 @@ class Criteria
     protected $joins = [];
 
     /**
-     * @var \Propel\Runtime\ActiveQuery\Criteria[]
+     * @var array<string,\Propel\Runtime\ActiveQuery\Criteria>
```
dereuromark, line 189: Do all tools and IDEs understand this newer syntax?
Or should those also go into phpstan-var here?

nederdirk: Thank you! I'm already so familiar with documenting `array<keytype,valuetype>` that I completely forgot that it is non-standard phpdoc. No-one can ever attest that all tools understand some nonstandard syntax, although I know from my own and colleagues' experience that some of the big ones (JetBrains PHPstorm, many plugins for Visual Studio Code, Sublime) all support it well.
I'll fix this and re-submit the PR!

nederdirk: ... and done!

To keep this PR manageable, I'll leave the template annotations for generated classes for a later time. With the changes as presented at least the results of many queries will provide more insight for static analysis tools.

dereuromark:
Are you able to rebase and get ci green? Then we can continue.

szepeviktor:
IF you write a small PHPStan extension you can reduce the number of ignores to zero.

PR #1658, opened 2020-10-12
Locking reads

prgTW:
This PR adds ability for locking reads, either shared or exclusive

Compatibility with:
- [MySQL 5.6+](https://dev.mysql.com/doc/refman/5.6/en/innodb-locking-reads.html)
- [PostgreSQL 8.1+](https://www.postgresql.org/docs/8.1/sql-select.html#SQL-FOR-UPDATE-SHARE)

Example:
```php
$booksReadInShareMode = (new BookQuery)
    ->filterById(123)
    ->forShare()
    ->find();
// SELECT ... FROM book LOCK IN SHARE MODE -- mysql
// SELECT ... FROM book FOR SHARE -- pgsql

$booksToBeUpdated = (new BookQuery)
  ->filterById(123)
  ->forUpdate()
  ->find();
// SELECT ... FROM book FOR UPDATE
```

### DIFF
--- a/src/Propel/Runtime/ActiveQuery/Criteria.php
+++ b/src/Propel/Runtime/ActiveQuery/Criteria.php
@@ -144,6 +144,13 @@ class Criteria
     protected $selectModifiers = [];
 
     /**
+     * Lock to be used to retrieve rows (if any).
+     *
+     * @var \Propel\Runtime\ActiveQuery\Lock|null
+     */
+    protected $lock;
+
+    /**
      * Storage of conditions data. Collection of Criterion objects.
      *
      * @var \Propel\Runtime\ActiveQuery\Criterion\AbstractCriterion[]
@@ -306,6 +313,7 @@ class Criteria
         $this->ignoreCase = false;
         $this->singleRecord = false;
         $this->selectModifiers = [];
+        $this->lock = null;
         $this->selectColumns = [];
         $this->orderByColumns = [];
         $this->groupByColumns = [];
@@ -1248,6 +1256,75 @@ class Criteria
     }
 
     /**
+     * @return \Propel\Runtime\ActiveQuery\Lock|null Get read lock value.
+     */
+    public function getLock(): ?Lock
+    {
+        return $this->lock;
+    }
+
+    /**
+     * Apply a shared read lock to be used to retrieve rows.
+     *
+     * @param string[] $tableNames
+     * @param bool $noWait
+     *
+     * @return $this Modified Criteria object (for fluent API)
+     */
+    public function lockForShare(array $tableNames = [], bool $noWait = false)
+    {
+        $this->withLock(Lock::SHARED, $tableNames, $noWait);
+
+        return $this;
+    }
+
+    /**
+     * Apply an exclusive read lock to be used to retrieve rows.
+     *
+     * @param string[] $tableNames
+     * @param bool $noWait
+     *
+     * @return $this Modified Criteria object (for fluent API)
+     */
+    public function lockForUpdate(array $tableNames = [], bool $noWait = false)
+    {
+        $this->withLock(Lock::EXCLUSIVE, $tableNames, $noWait);
+
+        return $this;
+    }
+
+    /**
+     * Apply a read lock to be used to retrieve rows.
+     *
+     * @see Lock::SHARED
+     * @see Lock::EXCLUSIVE
+     *
+     * @param string $lockType
+     * @param string[] $tableNames
+     * @param bool $noWait
+     *
+     * @return $this Modified Criteria object (for fluent API)
+     */
+    protected function withLock(string $lockType, array $tableNames = [], bool $noWait = false)
+    {
+        $this->lock = new Lock($lockType, $tableNames, $noWait);
+
+        return $this;
+    }
+
+    /**
+     * Retrieve rows without any read locking.
+     *
+     * @return $this Modified Criteria object (for fluent API)
+     */
+    public function withoutLock()
+    {
+        $this->lock = null;
+
+        return $this;
+    }
+
+    /**
      * Sets ignore case.
      *
      * @param bool $b True if case should be ignored.
@@ -1646,6 +1723,15 @@ class Criteria
                     }
                 }
 
+                $aLock = $this->lock;
+                $bLock = $criteria->getLock();
+                if ($aLock instanceof Lock && !$aLock->equals($bLock)) {
+                    return false;
+                }
+                if ($bLock instanceof Lock && !$bLock->equals($aLock)) {
+                    return false;
+                }
+
                 return true;
             } else {
                 return false;
@@ -1686,6 +1772,12 @@ class Criteria
         $selectModifiers = $criteria->getSelectModifiers();
         if ($selectModifiers && !$this->selectModifiers) {
             $this->selectModifiers = $selectModifiers;
+        }
+
+        // merge lock
+        $lock = $criteria->getLock();
+        if ($lock && !$this->lock) {
+            $this->lock = $lock;
         }
 
         // merge select columns
@@ -2112,6 +2204,10 @@ class Criteria
 
         if ($this->getLimit() >= 0 || $this->getOffset()) {
             $adapter->applyLimit($sql, $this->getOffset(), $this->getLimit(), $this);
+        }
+
+        if (null !== $this->lock) {
+            $adapter->applyLock($sql, $this->lock);
         }
 
         return $sql;
--- a/src/Propel/Runtime/ActiveQuery/Lock.php
+++ b/src/Propel/Runtime/ActiveQuery/Lock.php
@@ -1,0 +1,114 @@
+<?php
+
+/**
+ * MIT License. This file is part of the Propel package.
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types = 1);
+
+namespace Propel\Runtime\ActiveQuery;
+
+/**
+ * Class represents a query lock
+ *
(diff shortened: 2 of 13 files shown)


src/Propel/Runtime/ActiveQuery/Criteria.php
```diff
@@ -1247,6 +1255,41 @@ public function hasSelectModifier($modifier)
         return in_array($modifier, $this->selectModifiers);
     }
 
+    /**
+     * @return null|Lock Get lock value.
```
dereuromark, line 1259: Can we use FQCN in docblocks here, please?

prgTW: sure

src/Propel/Runtime/ActiveQuery/Criteria.php
```diff
@@ -1247,6 +1255,41 @@ public function hasSelectModifier($modifier)
...
+
+    /**
+     * Apply a read lock to be used to retrieve rows.
+     *
+     * @see self::SHARED
+     * @see self::EXCLUSIVE
+     *
+     * @return $this Modified Criteria object (for fluent API)
+     */
+    public function withLock(string $lockType, array $tableNames = [], $noWait = false)
```
dereuromark, line 1274: The param docblocks seem missing for these.
Also, could $noWait be bool typehinted?

prgTW: Sure, I could convert all my code to be properly typehinted according to minimum php reqs which is 7.2

prgTW: done

tests/Propel/Tests/Runtime/ActiveQuery/CriteriaTest.php
```diff
@@ -1122,10 +1122,10 @@ public function testWithSimpleLock()
...
         $c = new Criteria();
         $c->withLock(Lock::SHARED);
-        $this->assertInstanceOf(Lock::class, $c->getLock(), 'withLock() adds a shared read lock to the Criteria');
-        $this->assertEquals(Lock::SHARED, $c->getLock()->getType());
-        $this->assertEmpty($c->getLock()->getTableNames());
-        $this->assertFalse($c->getLock()->isNoWait());
+        self::assertInstanceOf(Lock::class, $c->getLock(), 'withLock() adds a shared read lock to the Criteria');
+        self::assertEquals(Lock::SHARED, $c->getLock()->getType());
+        self::assertEmpty($c->getLock()->getTableNames());
+        self::assertFalse($c->getLock()->isNoWait());
```
dereuromark, line 1128: this was actually correct, see also their docs.
In general, you can use assertSame though for scalar value comparison better.

prgTW: According to the [docs](https://phpunit.readthedocs.io/en/7.5/assertions.html) it's only a matter of personal preference but I can revert this. I've changed it because I had hints from inspection system in PHPStorm.

Regarding assertEquals vs assertSame - yes, I really don'y know whether to conform to how tests are currently written (not to introduce anything more besides the needs for this PR) or to write them how I personally feel (to pass inspections, properly typehinted etc.) 😄

src/Propel/Runtime/ActiveQuery/Criteria.php
```diff
@@ -1247,6 +1255,45 @@ public function hasSelectModifier($modifier)
...
+
+        return $this;
+    }
+
+    /**
+     * Retrieve rows without any read locking.
+     *
+     * @return $this Modified Criteria object (for fluent API)
+     */
+    public function withoutLock(): self
```
dereuromark, line 1290: Please avoid chainable typehints. $this must stay without
Not sure if the sniffer already reports this.

tests/Propel/Tests/Runtime/ActiveQuery/CriteriaTest.php
prgTW: reverted `self::` to `$this->`

src/Propel/Runtime/ActiveQuery/Criteria.php
prgTW: You mean this?
```suggestion
    public function withoutLock()
```
sniffer didn't report this

dereuromark:
https://travis-ci.org/github/propelorm/Propel2/jobs/735176770
You can run the sniffer to auto fix those issues.

prgTW:
@dereuromark yes, I forgot there's `cs-check`/`cs-fix` composer scripts. Already fixed

dereuromark:
Do you intend to add more? As your PR is still in draft mode.

prgTW:
Yes, I can flag it as "ready for review" because, for now, I'm unaware of anything that this feature breaks or to be tested more thoroughly. Of course I'm open for discussion, hints etc..

prgTW:
I also thought about adding `forShare()`/`forUpdate()` instead of `withLock` to skip passing `Lock::SHARED` or `Lock::EXCLUSIVE` explicitly. 
```diff
Pros
+ being closer to how we write native SQL query (`SELECT .. FOR SHARE` and `SELECT .. FOR UPDATE`)
+ safety, as one could pass a custom string e.g. `non-existing` as the lock type (or validation should be introduced)

Cons:
- introducing new type of lock (future proof) would require new method
```

src/Propel/Runtime/ActiveQuery/Lock.php
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * Lock type, either shared or exclusive
+     *
+     * @see self::SHARED
+     * @see self::EXCLUSIVE
+     *
+     * @var string
+     */
+    private $type;
```
dereuromark, line 32: Not sure what the rest of the system is using
But can we also use protected visibility here?
If anyone needs to extend this, this would make it much harder in that rare case - for no real benefit at this point.

dereuromark:
The current implementation sounds like a good compromise. I would probably keep it as is for now.

src/Propel/Runtime/ActiveQuery/Lock.php
prgTW: sure, I've changed to `protected` as requested, makes more sense if there will be children classes in the future

dereuromark (approved):

src/Propel/Runtime/ActiveQuery/Criteria.php
```diff
@@ -1247,6 +1255,75 @@ public function hasSelectModifier($modifier)
...
+
+    /**
+     * Apply an exclusive read lock to be used to retrieve rows.
+     *
+     * @param string[] $tableNames
+     * @param bool $noWait
+     *
+     * @return $this Modified Criteria object (for fluent API)
+     */
+    public function forUpdate(array $tableNames = [], bool $noWait = false)
```
ehsanmx, line 1289: As far as I know, the method name `forUpdate` is a very standard name, and some other projects already implemented this as it was necessary, this can cause BC breaking:
e.g. 
https://github.com/spryker/propel-orm/blob/master/src/Spryker/Zed/PropelOrm/Business/Builder/QueryBuilder.php#L447

 so maybe a more unique name can reduce the chance of the breaking or do you have a better idea. this is the same for `forShare`

ehsanmx (changes requested):

src/Propel/Runtime/ActiveQuery/Criteria.php
prgTW: Hi @ehsanmx. Yes, these names are common-looking because I wanted to make the naming similar to how Propel is built in general. Originally the API was `withLock(Lock::SHARED)` but that just didn't feel right IMHO, not with other methods named so simple and similar-looking to what SQL offers in standard - that's the "why" part of this PR.

I know that such change could introduce a BC break in other projects but should we look at who and how implemented the library? Choosing other names for these methods like `selectForUpdate` may subject to the same problem - somebody somewhere may have implemented such method already - it's rather a matter of probability of such situation in other projects. Personally I don't think that it's a BC break because the API isn't changed or removed but instead a feature is added so it's a minor change (according to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)). I don't know if Propel follows semantic versioning (doubt it really) although the idea of following it would be nice.

But, not to be so negative about it 😉, I would propose also `lockForShare`/`lockForUpdate` or `withSharedLock`/`withExclusiveLock`, however these don't feel as nice and close to native SQL parts as the originally proposed names. Maybe that's the same reason why currently proposed naming was also implemented in [QueryBuilder.php#L447](https://github.com/spryker/propel-orm/blob/master/src/Spryker/Zed/PropelOrm/Business/Builder/QueryBuilder.php#L447)?

ehsanmx: Hey @prgTW,
Thanks for your reply, I think having more keywords in the method name like `lockForShare/lockForUpdate` reduce the chance of breaking and I'm ok with that

dereuromark:
The PHP 8 compatibility seems to affect this PR now. Requires a resolve now.

src/Propel/Runtime/ActiveQuery/Criteria.php
prgTW: Naming changed as requested - `lockForShare`,`lockForUpdate`,`withoutLock` :)

prgTW:
Rebased on newest master and squashed

ehsanmx (approved):

prgTW:
Tests are passing but one `phpstan` error, however I would prefer this error to be fixed in a separate PR no to mix responsibilities

dereuromark:
Great work!

Issue #1659, opened 2020-10-21
Can not build with non-primary foreign key


chuongnh3atgmaildotcom:
Seems #1259 caused an issue when building model with non-primary foreign key.
That case is mentioned in this discussion https://github.com/propelorm/Propel2/pull/1259#discussion-diff-76629831. The discussion hasn't finished and now I'm finding that exact case.

**Steps to Reproduce**
1. foreign key refers to an unique key of $foreignTable instead of its primary key, and foreign table has composite primary key.
    ```
    <?xml version="1.0" encoding="ISO-8859-1" standalone="no"?>
    <database name="bookstore" defaultIdMethod="native" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"    namespace="Propel\Tests\Bookstore">
      <table name="event">
        <column name="id" type="integer" required="true" primaryKey="true" autoIncrement="true"/>
        <column name="name" type="varchar" size="50" required="true"/>
        <column name="Event" type="varchar" size="50" required="true"/>
        <column name="organiser_id" type="integer" required="true"/>
        <!-- This FK is incomplete -->
        <foreign-key foreignTable="organiser" name="test_event_fk_Event_organiser_name">
            <reference local="Event" foreign="name"/>
        </foreign-key>
      </table>
      <table name="organiser">
        <!-- Has composite PK -->
        <column name="id" type="integer" required="true" primaryKey="true" autoIncrement="true"/>
        <column name="secondary" type="integer" required="true" primaryKey="true"/>
        <column name="name" type="varchar" size="50" required="true"/>
        <unique>
            <unique-column name="name"/>
        </unique>
      </table>
    </database>
    ```
2.  run `bin/propel model:build`

**Expected Behavior**
Build successfully

**Current Behavior**
error thrown: \Propel\Generator\Exception\BuildException Table "event" contains a foreign key to table "organiser" but does not have a reference to foreign primary key "id"

**Possible Solution**
Do not check for incomplete foreign key if it does not refer to any foreign key

dereuromark:
Are you able to make a PR with suggested changes to fix this?

chuongnh3atgmaildotcom:
@dereuromark 
https://github.com/propelorm/Propel2/pull/1661

PR #1661, opened 2020-10-22
#1659 skip incomplete check for non-primary foreign key

chuongnh3atgmaildotcom:
Related Issue: #1659

### DIFF
--- a/src/Propel/Generator/Model/ForeignKey.php
+++ b/src/Propel/Generator/Model/ForeignKey.php
@@ -841,7 +841,7 @@ class ForeignKey extends MappingModel
      *
      * @return bool Returns true if all columns inside this foreign key are not primary keys of the foreign table
      */
-    public function isForeignNonPrimaryKey()
+    public function isForeignNonPrimaryKey(): bool
     {
         $foreignTable = $this->getForeignTable();
 
--- a/src/Propel/Generator/Model/ForeignKey.php
+++ b/src/Propel/Generator/Model/ForeignKey.php
@@ -836,6 +836,30 @@ class ForeignKey extends MappingModel
     }
 
     /**
+     * Returns whether this foreign key is not the primary key of the foreign
+     * table.
+     *
+     * @return bool Returns true if all columns inside this foreign key are not primary keys of the foreign table
+     */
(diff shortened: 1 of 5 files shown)


src/Propel/Generator/Model/ForeignKey.php
```diff
@@ -835,6 +835,31 @@ public function isForeignPrimaryKey()
             && !array_diff($foreignPKCols, $foreignCols));
     }
 
+    /**
+     * Returns whether this foreign key is not the primary key of the foreign
+     * table.
+     *
+     * @return bool Returns true if all columns inside this foreign key are not primary keys of the foreign table
+     */
+    public function isForeignNonPrimaryKey()
```
dereuromark, line 844: Do we want to add `: bool` typehint here?

chuongnh3atgmaildotcom: Could you please explain why it is BC? This is similar to above methods `isForeignPrimaryKey`

dereuromark: This method does not exist. So anyone that extended this class will not have a problem with the typehint here
Adding it to other existing methods will cause a fatal error on extended classes.

dereuromark (approved):

a-sabaa (approved):

tests/Propel/Tests/Issues/Issue1659Test.php
```diff
@@ -0,0 +1,106 @@
...
+use Propel\Tests\TestCaseFixtures;
+
+/**
+ * Test : Propel should not allow incomplete foreign key references when foreign table has a composite primary key
+ * Issue described in https://github.com/propelorm/Propel2/issues/675.
+ * Originally described in http://stackoverflow.com/questions/7947085/are-incomplete-key-references-in-propel-useful
+ *
+ * @group model
+ */
+class Issue1659Test extends TestCaseFixtures
```
matweew, line 21: We need to find better name for this class

dereuromark: Sure, that can be done as follow up. For now I guess people find the details when needed.

PR #1664, opened 2020-10-25
Compatibility with Symfony 4.0+ instead of 4.3+

prgTW:
Lowering the requirements for `symfony/*` libs from  4.3+ to 4.0+.

According to current dependencies definitions in `composer-symfony4-min.json` tests of such are passing:

https://github.com/propelorm/Propel2/blob/3087361bd27f3598f3ecd70e9fd272a5c2575b88/tests/composer/composer-symfony4-min.json#L20-L26

https://travis-ci.org/github/propelorm/Propel2/jobs/726959500
image

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,13 +17,13 @@
     ],
     "require": {
         "php": ">=7.2",
-        "symfony/yaml": "^3.4.0||^4.3.0||^5.0.0",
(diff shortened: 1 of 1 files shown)


dereuromark:
< 4.3 is kind of EOL, but if it is supported/green I don't see the harm :+1:

PR #1665, opened 2020-10-29
fix identifierQuoting for Versionable behavior

kasparsj:


dereuromark:
Looks good and BC to me.
Are you able to provide a small test case along with your chance. To make sure we won't run into this again in the future? That would be awesome.

gechetspr (approved):

PR #1672, opened 2020-11-05
Added support for keytype in the magic import/export methods

### LINKED ISSUE #1671: toJSON but then lowercase
Hi,

I prefer to keep everything in one standard within my codebase. Now is JSON case-sensitive (unfortunately), and I'm having trouble with Propel not giving me the option to change how it outputs JSON. I want to use the `toJSON()` function, but I want all the field names to be lowercase (and not `TYPE_PHPNAME`). I don't want to add post-processing everywhere; and rewriting existing functionality is also not desired. Would it be possible to pass through an optional parameter down to the `exportTo($parser, $includeLazyLoadColumns = true)` function, changing the hardcoded `TableMap::TYPE_PHPNAME`?

I think the change would involve the following:

1. Prepare the enum value here
https://github.com/propelorm/Propel2/blob/ae14a1fb891d1e45be54f8d9c8e016d7a9a4c0e1/src/Propel/Generator/Builder/Om/templates/baseObjectMethodMagicCall.php#L53-L58

2. Change `exportTo()` to accept an extra variable
https://github.com/propelorm/Propel2/blob/0839979f5721f4810b9b02ac8e96e5e8f4358182/src/Propel/Generator/Builder/Om/templates/baseObjectMethods.php#L180-L199

3. Pass it down to `$this->toArray()`
https://github.com/propelorm/Propel2/blob/0839979f5721f4810b9b02ac8e96e5e8f4358182/src/Propel/Generator/Builder/Om/templates/baseObjectMethods.php#L198

----

Possibly it also should be supported here?
https://github.com/propelorm/Propel2/blob/225a124256ccc5b18cf3325988ad1f49ba756811/src/Propel/Runtime/Collection/Collection.php#L541-L571

----

Am I missing anything in this flow? Would this fit in the Propel style? I would be willing to spend some time on building this if there is viability in supplying this setting. Otherwise Ill just have to write something custom for myself to process the keys to lowercase.

MagicLegend:
Resolves #1671


dereuromark:
Can you add a small test case for this please? :)
Then it looks good!

MagicLegend:
Thanks for the feedback, will do tomorrow!

MagicLegend:
@dereuromark One of the tests is failing because of the following:

```
There was 1 failure:

1) Propel\Tests\Runtime\ActiveRecord\ActiveRecordConvertTest::testToJSON with data set #1 ('{"id":9012,"title":"Don Juan"...N*"]}}', 'camelName')

--- Expected

+++ Actual

@@ @@

-'{"id":9012,"title":"Don Juan","isbn":"0140422161","price":12.99,"publisherId":1234,"authorId":5678,"publisher":{"id":1234,"name":"Penguin","books":["*RECURSION*"]},"author":{"id":5678,"firstName":"George","lastName":"Byron","email":null,"age":null,"books":["*RECURSION*"]}}'

+'{"id":9012,"title":"Don Juan","iSBN":"0140422161","price":12.99,"publisherId":1234,"authorId":5678,"publisher":{"id":1234,"name":"Penguin","books":["*RECURSION*"]},"author":{"id":5678,"firstName":"George","lastName":"Byron","email":null,"age":null,"books":["*RECURSION*"]}}'
```

What seems to have happened is that the `camelName` only took care of the first capital letter. What would be the desired behaviour in this case? I would expect it to be fully lower case (like [here](https://gchq.github.io/CyberChef/#recipe=To_Camel_case(false)&input=SVNCTg)). Where can I find the associated functionality?

----

Second issue:

```
3) Propel\Tests\Runtime\ActiveRecord\ActiveRecordConvertTest::testfromJSON with data set #0 ('{"Id":9012,"Title":"Don Juan"...N*"]}}', 'phpName')

Propel\Runtime\Exception\PropelException: Method getFieldNames() expects the parameter \$type to be one of the class constants TableMap::TYPE_PHPNAME, TableMap::TYPE_CAMELNAME, TableMap::TYPE_COLNAME, TableMap::TYPE_FIELDNAME, TableMap::TYPE_NUM. {"Id":9012,"Title":"Don Juan","ISBN":"0140422161","Price":12.99,"PublisherId":1234,"AuthorId":5678,"Publisher":{"Id":1234,"Name":"Penguin","Books":["*RECURSION*"]},"Author":{"Id":5678,"FirstName":"George","LastName":"Byron","Email":null,"Age":null,"Books":["*RECURSION*"]}} was given.

/home/travis/build/MagicLegend/Propel2/src/Propel/Runtime/Map/TableMapTrait.php:29

/home/travis/build/MagicLegend/Propel2/tests/Fixtures/bookstore/build/classes/Propel/Tests/Bookstore/Base/Book.php:1644

/home/travis/build/MagicLegend/Propel2/tests/Fixtures/bookstore/build/classes/Propel/Tests/Bookstore/Base/Book.php:1691

/home/travis/build/MagicLegend/Propel2/tests/Fixtures/bookstore/build/classes/Propel/Tests/Bookstore/Base/Book.php:5080

/home/travis/build/MagicLegend/Propel2/tests/Propel/Tests/Runtime/ActiveRecordConvertTest.php:218
````

I have to be honest; I'm not sure what goes wrong here. As far as I can see the failing code comes from the generated `toArray()` function, which is created here:

https://github.com/propelorm/Propel2/blob/1dd4968b0feb71b9af65eb1c0b2ffed051535ac6/src/Propel/Generator/Builder/Om/ObjectBuilder.php#L2924-L3028

However, I don't see what change I made that could've broken this behaviour. Do you have any idea?

**Comment by MagicLegend** [comment]
_(I started writing the previous comment when only the first test had failed)_

I just noticed that only the aganostic tests fail... All other tests are green. Why is there something different?

**Comment by dereuromark** [comment]
Yeah only agnostic ones - https://travis-ci.org/github/propelorm/Propel2/jobs/741856981
I also wonder how.

But we do need to fix them up if we want to get this merged.

**Comment by mringler** [comment]
I had a minute to look into why the test failed:

First one is pretty straight forward: if you set format to be `camelName`, you get what is set as `camelName` in BookTableMap:
```php
    protected static $fieldNames = array (
        self::TYPE_PHPNAME       => array('Id', 'Title', 'ISBN', 'Price', 'PublisherId', 'AuthorId', ),
        self::TYPE_CAMELNAME     => array('id', 'title', 'iSBN', 'price', 'publisherId', 'authorId', ),
        self::TYPE_COLNAME       => array(BookTableMap::COL_ID, BookTableMap::COL_TITLE, BookTableMap::COL_ISBN, BookTableMap::COL_PRICE, BookTableMap::COL_PUBLISHER_ID, BookTableMap::COL_AUTHOR_ID, ),
        self::TYPE_FIELDNAME     => array('id', 'title', 'isbn', 'price', 'publisher_id', 'author_id', ),
        self::TYPE_NUM           => array(0, 1, 2, 3, 4, 5, )
    );
```
And there, the column is named 'iSBN'. One could argue that it should be 'isbn', but that is a problem with the model generator, not with the `toJson` function. Changing the test case to expect 'iSBN' (or better yet, retrieving the name from the table map) correctly fixes the test case.

The second one is a off-by-one error, I'll mark the line in the code.

@MagicLegend, still interested in fixing this? It's a nice feature, and you already put quite some work into it. Would be nice to see it done.

MagicLegend:
@mringler Thank you for taking the time to look into it!

1. Would it be okay to change the test to accept `iSBN`, even though technically it's an incorrect `camelCase` interpretation?
2. You haven't marked anything in the PR; is it related to something else?

And yes; I'd still like to see this merged if possible :)

mringler:
Hey @MagicLegend, happy to see your still around for this!

1, Yes, changing to 'iSBN' is correct. It is not a proper camelCase name, but it is what Propel gives you. It is not a problem in this PR.
Good question why only agnostic tests fail, they all should. Irritating. Maybe a problem with the setup? I am not involved in that, so your guess is as good as mine.

2, You should see the line with the error in the code review between my initial comment and yours. You can also find it by clicking on "Files changed" at the top.
Basically, in the `__call()` function, you store the JSON string in `$keyType`, instead of the actual type (line 49 in baseObjectMagicMethodCall.php), and pass that to `->importFrom()` as key type. Again, this should fail for all test types, not just agnostic.
Changing the index from 0 to 1 fixes this issue, I think.

Changing these two things should make it work. Let me know how it goes.

Oh, if you want to have a look at the camelCase issue, it happens in Propel\Generator\Model\Column.php, line 616ff:
```php
    /**
     * Returns the camelCase version of the PHP name.
     *
     * The studly name is the PHP name with the first character lowercase.
     *
     * @return string
     */
    public function getCamelCaseName()
    {
        return lcfirst($this->getPhpName());
    }
```
Obviously, this does not work well with allcap names.

mringler (commented):
Gee, I newver submitted my review. No wonder you couldn't find it. Sorry.

mringler:
Hey @MagicLegend, didn't hear back from you, so I added the fixes in a new PR #1723

MagicLegend:
Hey! I'm sorry, havoc at work around a golive. Glad to hear you've fixed it!

PR #1676, opened 2020-11-25
Wrong behavior when inserting rows into crossRef table with auto increment primary key column.

WebsourceCz:
When inserting rows into crossRef table with auto increment primary key column defined at other than first position in table schema, then auto increment value is set to first defined primary key column instead of the correct one.

### DIFF
--- a/src/Propel/Generator/Builder/Om/ObjectBuilder.php
+++ b/src/Propel/Generator/Builder/Om/ObjectBuilder.php
@@ -6609,16 +6609,18 @@ abstract class " . $this->getUnqualifiedClassName() . $parentClass . ' implement
         } catch (Exception \$e) {
             throw new PropelException('Unable to get autoincrement id.', 0, \$e);
         }";
-            $column = $table->getFirstPrimaryKeyColumn();
-            if ($column) {
+            foreach ($table->getPrimaryKey() as $col) {
+                if (!$col->isAutoIncrement()) {
+                    continue;
+                }
                 if ($table->isAllowPkInsert()) {
                     $script .= "
         if (\$pk !== null) {
-            \$this->set" . $column->getPhpName() . "(\$pk);
+            \$this->set" . $col->getPhpName() . "(\$pk);
         }";
                 } else {
                     $script .= "
-        \$this->set" . $column->getPhpName() . '($pk);';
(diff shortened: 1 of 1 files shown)


dereuromark:
Can you add a test case for this please?

dereuromark:
ping @WebsourceCz

src/Propel/Generator/Builder/Om/ObjectBuilder.php
```diff
@@ -6609,16 +6609,18 @@ protected function addDoInsertBodyRaw()
...
-            if ($column) {
+            foreach ($table->getPrimaryKey() as $col) {
+                if (!$col->isAutoIncrement()) {
+                    continue;
+                }
                 if ($table->isAllowPkInsert()) {
                     $script .= "
         if (\$pk !== null) {
-            \$this->set" . $column->getPhpName() . "(\$pk);
+            \$this->set" . $col->getPhpName() . "(\$pk);
```
dereuromark, line 6619: Can we keep the $column variable, or is there a conflict?

dereuromark:
Maybe you can provide as an example or alike we can add into the tests manually
Without that we cannot merge this I am afraid.

WebsourceCz:
Imagine you have table for M:N relation of another two tables and you need to work with relation rows using autoincrement id. So xml schema of this table is for example:

<table name="page_page_group" isCrossRef="true">
    <column name="page_id" type="integer" primaryKey="true" required="true" />
    <column name="page_group_id" type="integer" primaryKey="true" required="true" />
    <column name="id" type="integer" primaryKey="true" autoIncrement="true" required="true" />
    <column name="priority" type="integer" required="true" defaultValue="0" />
    <foreign-key foreignTable="page" name="page_page_group_FK_1" onDelete="cascade" onUpdate="cascade">
      <reference local="page_id" foreign="id" />
    </foreign-key>
    <foreign-key foreignTable="page_group" name="page_page_group_FK_2" onDelete="cascade" onUpdate="cascade">
      <reference local="page_group_id" foreign="id" />
    </foreign-key>
  </table>

When new row is inserted by Propel, old code found page_id as first primary key and set new auto increment id value returned from database to page_id column instead of id column. I know that I can set id as first column in the table, but I think, that when I tried it, there was some problem with auto creating of generated object and query methods for cross relation.

PR #1682, opened 2020-12-08
[WIP] Move continuous integration to Github Actions

### LINKED ISSUE #1668: PRIO: Move to GitHub actions
Since Travis is not supporting FOSS anymore really, we would need to use
now GitHub Actions for example.

Example of PHP with sqlite, mysql, postgres and prefer lowest:
https://github.com/FriendsOfCake/search/pull/299

Anyone wants to try doing this?
That would be awesome.

We can also do only parts of the current test matrix and incrementally improve it.

cristianoc72:
I marked this PR as _WIP_ because I need some help:

-  ~~I can't connect to Postgres server. I tryed in many ways (setting up passwords or not, define a database name or not, installing it via an action or calling directly Docker). Since I have never worked with Postgres in depth, I'm sure I'm mistaking some obvious details that I don't know.~~
- For Symfony 3 and 4-min, `Propel\Tests\Generator\Command\InitCommandTest::testExecute()` fails. Unfortunately, I can't reproduce that error on my laptop: locally everything works fine, with all Symfony versions. Any ideas?

After solved ~~these points~~ this point, we can join `tests.yml` with `ci.yml`.

This PR closes #1668

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,83 @@
+name: Propel Test Suite
+
+on: [push, pull_request]
+
+jobs:
+    testsuite:
+        runs-on: ubuntu-18.04
+        strategy:
+            fail-fast: false
+            matrix:
+                php-version: ['7.2', '7.4']
+                db-type: [sqlite, mysql, pgsql, agnostic]
+                symfony-version: [ '3-min', '3-max', '4-min', '4-max', '5-min', '5-max' ]
+        steps:
+            - name: Install PostgreSQL latest
+              if: matrix.db-type == 'pgsql' && matrix.php-version == '7.4'
+              uses: CasperWA/postgresql-action@v1.2
+              with:
+                  postgresql db: 'propel-tests'
+                  postgresql user: 'postgres'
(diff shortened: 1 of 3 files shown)


dereuromark:
I also cannot reproduce locally.
Maybe we just merge it for now, since Travis is dead completely, so right now we are blind.

I prepared https://github.com/propelorm/Propel2/commit/dbe30d877b9c542666c44cb51c5c803a5bddba7b as it seems we need to move away from whitelist to include in XMLs.

And we should check that phpstan/psalm are also updated.

cristianoc72:
@stereomon you're right! 
I found several issues (like this https://github.com/actions/runner/issues/241) about Github Actions terminal and Symfony documentation tells that it manipulates tty to hide inputs. Maybe Travis implement a more advancend terminal.

I'm thinking about how to fix it. The only choiche seems to be to disable hidden inputs in test environment.

stereomon:
I have a fix for the issue, thanks for pointing to the actions/runner#241! I will finalize my changes and push them.

cristianoc72:
See #1686

PR #1688, opened 2020-12-21
Introduce PHP 8

cristianoc72:
Make Propel2 compatible with PHP 8 (well, it's already compatible: just few adjustments).

~~Test suite is green but the new version of PhpStan raises some esotheric (for me) errors: could you suggest me, please, how to fix them?~~

I added some composer scripts, if you don't like I can remove them, of course.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,12 +14,15 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php-version: [ '7.2', '7.4' ]
+                php-version: [ '7.2', '7.4', '8.0' ]
                 db-type: [ sqlite, mysql, pgsql, agnostic ]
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -16,7 +16,7 @@
         }
     ],
     "require": {
-        "php": ">=7.2",
+        "php": "^7.2|^8.0",
```
dereuromark, line 19: This must be reverted.

.github/workflows/ci.yml
```diff
@@ -14,12 +14,15 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php-version: [ '7.2', '7.4' ]
+                php-version: [ '7.2', '7.4', '8.0' ]
```
dereuromark, line 17: @stereomon 
We might have overlooked that when porting from Travis

composer.json
cristianoc72: Fixed.

PR #1689, opened 2020-12-27
Speed up some tests

cristianoc72:
Speed up some tests by mocking filesystem, with
`mikey176/vfsstream` library.


dereuromark:
Removing i/o ops here is also a big win regarding permission issues. 👍

cristianoc72:
> Removing i/o ops here is also a big win regarding permission issues. +1

Yep! In fact my original idea was to make the `QuickBuilder` completely based on virtual filesystem. The biggest obstacle is that Propel makes extensive use of `chdir` and `cwd` functions, that aren't stream compatible.
A possible solution could be define something like Symfony `kernel_dir` variable and use it as base directory for file manipulations. 
Work in progress :smile:

PR #1691, opened 2020-12-29
QuickBuilder refactor

cristianoc72:
Refactor`QuickBuilder` to build classes on virtual filesystem.
This commit introduces the following differences:

-  The built classes are grouped in a single file and written on virtual filesystem, by default. Optionally, it's possible to write them on physical filesystem, for debugging purpose. When we use the physical filesystem, the classes are written in separated files, to make the debug easier. E.g.:
```php
<?php
use Propel\Generator\Util\QuickBuilder;

$schema = '<database>.............</database>';

//---The following builds on virtual filesystem
$builder = new QuickBuilder();
$builder->build($schema);
     // or alternatively
QuickBuilder::buildSchema($schema);


//---The following builds on physical filesystem
$builder = new QuickBuilder();
$builder->setVfs(false);
$builder->build($schema);
     // or alternatively
QuickBuilder::buildSchema($schema, null, null, null, null, false);
```
-  I moved `VfsTrait` from __Tests__  to __Propel\Generator\Util__ namespace, because otherwise PhpStan can't autoload it, rising an error (the more I use PhpStan the more I hate it :smile: ).
-   The refactored code and relative tests are strict typed.

### DIFF
--- a/src/Propel/Generator/Util/QuickBuilder.php
+++ b/src/Propel/Generator/Util/QuickBuilder.php
@@ -1,4 +1,4 @@
-<?php
+<?php declare(strict_types = 1);
 
 /**
  * MIT License. This file is part of the Propel package.
@@ -17,7 +17,10 @@ use Propel\Generator\Exception\BuildException;
 use Propel\Generator\Model\Database;
 use Propel\Generator\Model\Diff\DatabaseComparator;
 use Propel\Generator\Model\Table;
+use Propel\Generator\Platform\PlatformInterface;
 use Propel\Generator\Platform\SqlitePlatform;
+use Propel\Generator\Reverse\SchemaParserInterface;
+use Propel\Runtime\Adapter\AdapterInterface;
 use Propel\Runtime\Adapter\Pdo\SqliteAdapter;
 use Propel\Runtime\Connection\ConnectionInterface;
 use Propel\Runtime\Connection\ConnectionWrapper;
@@ -27,37 +30,39 @@ use Propel\Runtime\Propel;
 
 class QuickBuilder
 {
+    use VfsTrait;
+
     /**
      * The Xml.
      *
      * @var string
      */
-    protected $schema;
+    protected $schema = '';
 
     /**
      * The Database Schema.
      *
      * @var string
      */
-    protected $schemaName;
+    protected $schemaName = '';
 
     /**
-     * @var \Propel\Generator\Platform\PlatformInterface
+     * @var \Propel\Generator\Platform\PlatformInterface|null
      */
     protected $platform;
 
     /**
-     * @var \Propel\Generator\Config\GeneratorConfigInterface
+     * @var \Propel\Generator\Config\GeneratorConfigInterface|null
      */
     protected $config;
 
     /**
-     * @var \Propel\Generator\Model\Database
+     * @var \Propel\Generator\Model\Database|null
      */
     protected $database;
 
     /**
-     * @var \Propel\Generator\Reverse\SchemaParserInterface
+     * @var \Propel\Generator\Reverse\SchemaParserInterface|null
      */
     protected $parser;
 
@@ -74,11 +79,19 @@ class QuickBuilder
     protected $identifierQuoting = false;
 
     /**
+     * If use the virtual or physical filesystem.
+     * Default to virtual.
+     *
+     * @var bool
+     */
+    protected $vfs = true;
+
+    /**
      * @param string $schema
      *
      * @return void
      */
-    public function setSchema($schema)
+    public function setSchema(string $schema): void
     {
         $this->schema = $schema;
     }
@@ -86,7 +99,7 @@ class QuickBuilder
     /**
      * @return string
      */
-    public function getSchema()
+    public function getSchema(): string
     {
         return $this->schema;
     }
@@ -96,7 +109,7 @@ class QuickBuilder
      *
      * @return void
      */
-    public function setSchemaName($schemaName)
+    public function setSchemaName(string $schemaName): void
     {
         $this->schemaName = $schemaName;
     }
@@ -104,7 +117,7 @@ class QuickBuilder
     /**
      * @return string
      */
-    public function getSchemaName()
+    public function getSchemaName(): string
     {
         return $this->schemaName;
     }
@@ -114,15 +127,15 @@ class QuickBuilder
      *
      * @return void
      */
-    public function setParser($parser)
+    public function setParser(SchemaParserInterface $parser): void
     {
         $this->parser = $parser;
     }
 
     /**
-     * @return \Propel\Generator\Reverse\SchemaParserInterface
+     * @return \Propel\Generator\Reverse\SchemaParserInterface|null
      */
-    public function getParser()
+    public function getParser(): ?SchemaParserInterface
     {
         return $this->parser;
     }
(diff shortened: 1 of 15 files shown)


src/Propel/Generator/Util/QuickBuilder.php
```diff
@@ -235,9 +274,9 @@ public function build($dsn = null, $user = null, $pass = null, $adapter = null,
     /**
      * @return \Propel\Generator\Model\Database|null
      */
-    public function getDatabase()
+    public function getDatabase(): ?Database
     {
-        if ($this->database === null) {
+        if (!isset($this->database)) {
```
stereomon, line 279: What's the reason behind this change? !isset is slower than === null

src/Propel/Generator/Util/QuickBuilder.php
```diff
@@ -174,26 +187,52 @@ public function setConfig(GeneratorConfigInterface $config)
      */
     public function getConfig()
     {
-        if ($this->config === null) {
+        if (!isset($this->config)) {
```
stereomon, line 190: What's the reason behind this change? `!isset` is slower than `=== null`

src/Propel/Generator/Util/QuickBuilder.php
```diff
@@ -144,9 +157,9 @@ public function setPlatform($platform)
      *
      * @return \Propel\Generator\Platform\PlatformInterface
      */
-    public function getPlatform()
+    public function getPlatform(): PlatformInterface
     {
-        if ($this->platform === null) {
+        if (!isset($this->platform)) {
```
stereomon, line 162: What's the reason behind this change? `!isset` is slower than `=== null`

src/Propel/Generator/Util/QuickBuilder.php
```diff
@@ -114,17 +127,17 @@ public function getSchemaName()
...
 
     /**
-     * @return \Propel\Generator\Reverse\SchemaParserInterface
+     * @return \Propel\Generator\Reverse\SchemaParserInterface|null
      */
-    public function getParser()
+    public function getParser(): ?SchemaParserInterface
     {
-        return $this->parser;
+        return $this->parser ?? null;
```
stereomon, line 140: Isn't `$this->parser` `null` by default?

dereuromark: The docblock in property needs to be fixed then to append `|null`, as well

src/Propel/Generator/Util/QuickBuilder.php
dereuromark: I would also like to avoid seeing any unrelated change here that is not necessary as it only blocks moving forward.

src/Propel/Generator/Util/QuickBuilder.php
```diff
@@ -343,11 +382,11 @@ public function getSQL()
     }
 
     /**
-     * @param string[]|null $classTargets
+     * @param array $classTargets
      *
      * @return string
      */
-    public function getBuildName($classTargets = null)
+    public function getBuildName(array $classTargets = []): string
```
dereuromark, line 389: This accepted null so far.
I am not against making this a bit more strict - but we need to be careful about either the BC approach or to clearly state this in the release notes for the next release coming out.

src/Propel/Generator/Util/QuickBuilder.php
```diff
@@ -357,7 +396,7 @@ public function getBuildName($classTargets = null)
             $tables[] = $table->getName();
         }
         $name = implode('_', $tables);
-        if (!$classTargets || count($classTargets) === 5) {
+        if ($classTargets === [] || count($classTargets) === 5) {
```
dereuromark, line 399: This can stay the way it was.

src/Propel/Generator/Util/QuickBuilder.php
```diff
@@ -606,8 +597,64 @@ public function isIdentifierQuotingEnabled()
...
-    public function setIdentifierQuoting($identifierQuoting)
+    public function setIdentifierQuoting(bool $identifierQuoting): void
     {
         $this->identifierQuoting = $identifierQuoting;
     }
+
+    /**
+     * Create separate classes to write to physical filesystem.
+     *
+     * @param array $classes
```
dereuromark, line 608: You can be more concrete here, e.g. string[]
Same for other array usually

src/Propel/Generator/Util/QuickBuilder.php
```diff
@@ -606,8 +597,64 @@ public function isIdentifierQuotingEnabled()
...
+            }
+        }
+
+        return $includes;
+    }
+
+    /**
+     * Create an all-classes file to write to virtual filesystem.
+     *
+     * @param array $classes
```
dereuromark, line 636: Same

src/Propel/Generator/Util/QuickBuilder.php
cristianoc72: Nope. Starting from PHP 7.4, when you use strict types, the properties aren't initialize to null anymore. If `$this->parser` is not set, it's `uninitialized` and can't be accessed.

src/Propel/Generator/Util/QuickBuilder.php
cristianoc72: Starting from PHP 7.4, when you use strict types, the properties aren't initialized to null anymore, so  testing `=== null` rises a `Fatal error: Uncaught Error: Typed property QuickBuilder::$platform must not be accessed before initialization`.

Anyway, I can explicitly initialize the variables to `null`, if you think it's better.

src/Propel/Generator/Util/QuickBuilder.php
cristianoc72: idem

src/Propel/Generator/Util/QuickBuilder.php
cristianoc72: Fixed.

src/Propel/Generator/Util/QuickBuilder.php
cristianoc72: Fixed.

src/Propel/Generator/Util/QuickBuilder.php
cristianoc72: I adjusted the code to change it as little as possible.

src/Propel/Generator/Util/QuickBuilder.php
cristianoc72: Fixed.

src/Propel/Generator/Util/QuickBuilder.php
cristianoc72: Fixed.

src/Propel/Generator/Util/QuickBuilder.php
cristianoc72: Fixed.

src/Propel/Generator/Util/QuickBuilder.php
cristianoc72: I adjusted the code to change it as little as possible.

src/Propel/Generator/Util/QuickBuilder.php
cristianoc72: Fixed.

stereomon (approved):

PR #1694, opened 2021-01-11
Updated TableMap generator to add column name map for normalization, …

stereomon:
Updated TableMap generator to add column name map for normalization, updated TableMap class to make use of the new column name normalization.

This will increase performance of `TableMap::addColumn()`, `TableMap::getColumn()` and `TableMap::hasColumn()`.

### DIFF
--- a/src/Propel/Generator/Builder/Om/TableMapBuilder.php
+++ b/src/Propel/Generator/Builder/Om/TableMapBuilder.php
@@ -454,7 +454,7 @@ class " . $this->getUnqualifiedClassName() . " extends TableMap
     {
         $script .= '
     /**
-     * Holds a list of column names to be normalized
+     * Holds a list of column names and their normalized version.
      *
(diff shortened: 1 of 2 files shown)


src/Propel/Generator/Builder/Om/TableMapBuilder.php
```diff
@@ -444,6 +445,64 @@ protected function addFieldsAttributes()
...
+    /**
+     * @param string $script
+     *
+     * @return void
+     */
+    protected function addNormalizedColumnNameMap(&$script): void
+    {
+        $script .= '
+    /**
+     * Holds a list of column names to be normalized
```
dereuromark, line 457: to be normalized?
How about
```suggestion
     * Holds a list of column names and their normalized version.
```
maybe?

dereuromark (approved):

PR #1698, opened 2021-01-27
Avoid "Undefined index: column" in Criteria.php

peter17:
I'm getting thousands of "PHP Notice:  Undefined index: column" in server logs.

There is no guarantee that createSelectSql($params) will add those keys in $params, so, protecting access to those keys.


dereuromark (approved):

dereuromark:
One could use short nullsafe ternary here now.

demkos (approved):

PR #1699, opened 2021-02-09
Explicitly state supported PHP versions

szepeviktor:
...in composer.json


dereuromark:
This is not a recommended approach and will most likely not make it into master.

szepeviktor:
It may prevent unwanted PHP 9 compatibility.

dereuromark:
That is not an issue usually, speaking from experience since PHP5. Or in other words: The side effects of the PHP restricting is usually worse.
For details please see also Symfony or alike on their changes regarding this.

szepeviktor:
https://github.com/symfony/http-client/blob/c6f5bbe8e3e48f68c56285527f605b95c6f7dd9c/composer.json#L24

I see.

PR #1700, opened 2021-02-25
Add parameter-dependent psalm return type hint to temporal accessors

nederdirk:
Please refer to the Psalm documentation for conditional types[^0] for
more information. Although the docs specify that template parameters are
required, the reference-a-variable shorthand for templates also work,
see the [example].

Usually this project seems to prefer `@phpstan-`-prefixes, but this
construct is currently not available to PHPstan[^1].

[^0]: https://psalm.dev/docs/annotating_code/type_syntax/conditional_types/
[^1]: https://github.com/phpstan/phpstan/issues/3853
[example]: https://psalm.dev/r/11cdedd7b3

### DIFF
--- a/src/Propel/Generator/Builder/Om/ObjectBuilder.php
+++ b/src/Propel/Generator/Builder/Om/ObjectBuilder.php
@@ -894,6 +894,8 @@ abstract class " . $this->getUnqualifiedClassName() . $parentClass . ' implement
      * @return string|{$dateTimeClass}{$orNull} Formatted date/time value as string or $dateTimeClass object (if format is NULL), NULL if column is NULL" . ($handleMysqlDate ? ', and 0 if column value is ' . $mysqlInvalidDateString : '') . "
      *
      * @throws PropelException - if unable to parse/validate the date/time value.
+     *
+     * @psalm-return (\$format is null ? {$dateTimeClass}{$orNull} : string{$orNull})
      */";
     }
 


dereuromark (approved):

src/Propel/Generator/Builder/Om/ObjectBuilder.php
```diff
@@ -894,6 +894,8 @@ public function addTemporalAccessorComment(&$script, Column $column)
      * @return string|{$dateTimeClass}{$orNull} Formatted date/time value as string or $dateTimeClass object (if format is NULL), NULL if column is NULL" . ($handleMysqlDate ? ', and 0 if column value is ' . $mysqlInvalidDateString : '') . "
      *
      * @throws PropelException - if unable to parse/validate the date/time value.
+     *
+     * @psalm-return (\$format is null ? {$dateTimeClass}{$orNull} : string{$orNull})
```
mringler, line 898: The linked docs says this only works with template parameters, but [here](https://psalm.dev/articles/conditional-love) it says this is a short hand syntax that works, too, so I guess it is ok

mringler (approved):

src/Propel/Generator/Builder/Om/ObjectBuilder.php
nederdirk: Hey @mringler, somewhere else you wrote:

>When updating my project, I realized that in your other psalm PR, where you added `@psalm-return ($format is null ? DateTime : string)` to DateTime getters, I did not realize that these function also return `null`. You might want to fix that.
Correct type is probably
`@psalm-return null|($format is null ? DateTime : string)`

As you see, the returned type is `{$dateTimeClass}{$orNull}` or `string{$orNull}`, so if the field you're referencing is nullable, the `null` typehint should already be there!

mringler: Oh, you are right. The columns I saw were all required. Should have checked again. Sorry!

PR #1703, opened 2021-02-26
BehaviorLocator looks in dev packages too

mringler:
I installed my own behavior as dev-dependency, but it just didn't work. Turns out, when scanning composer.lock, the BehaviorLocator only looks in packages, not in packages-dev. Womp womp.


stereomon:
Can you elaborate on why you need to have a dev-only behavior?

mringler:
> Can you elaborate on why you need to have a dev-only behavior?

Well, from my understanding, it should be transparent for an application if a package comes from packages or dev-packages. So no matter why someone wanted to use dev-packages, not supporting it breaks the contract.
But there are several good reasons why someone wanted to use dev-packages, for example when model builds are only done during development. Another scenario would be behavior development. Or just plain old "because they want to" - they are not doing anything wrong, let's just assume they have their reasons.

As for me, I tried to install my package as dev requirement during behavior deveopment, using a semi-outdated manual, and it just wasn't working, to my increasing despair. Took me about an hour of painful code exploration to find that only half of the available places were inspected. 

I would argue, if you specifically do not want behaviors in dev-packages, you should throw an exception, telling the user their behavior is in the wrong place. Just ignoring it looks sloppy. Particularly when it is so easy to fix.

Does that make sense?

src/Propel/Generator/Util/BehaviorLocator.php
```diff
@@ -204,14 +204,17 @@ private function getCoreBehavior($name)
      *
      * @return array
      */
-    private function loadBehaviors($composerLock)
+    private function loadBehaviorsFromLockFile($composerLock)
```
dereuromark, line 207: Do we need to switch the method name here?
The argument is already clear to be the lock file.

dereuromark (approved):

src/Propel/Generator/Util/BehaviorLocator.php
mringler: When you use the method, you get `$this->loadBehaviors($anyNameYouDareToUse)`. So that's a problem. Not with `$this->loadBehaviorsFromLockFile($doesAnyoneElseSmellToast)`.
 Also, the method does not load behaviors generally, so why mislead?

dereuromark: Can you please add the docblock type for the argument?

mringler: What do you mean? Docblock has parameters with types and even a description (not to my credit, that was already there)

dereuromark: And if you modify a private method, you can also add types for argument and return where applicable.
As there is no BC issue in that case.

mringler: sure, no problem, I have added type hints

PR #1704, opened 2021-03-01
Bugfix: table alias and regular column name leads to cross join

mringler:
Ok, so this drove me nuts for quite a while. When using aliases and regular column names (i.e. a.name instead of a.Name), queries broke horribly, as Propel inserted these columns with a cross join.

Example:

```php
$c = new ModelCriteria('bookstore', 'Propel\Tests\Bookstore\Book');
$c->setModelAlias('b', true);
$c->where('b.title = ?', 'foo');    // note: using title, not Title
$c->join('b.Author a');
$c->where('a.first_name = ?', 'john'); // note: using first_name, not FirstName
```
led to query

```sql
SELECT  
FROM book b 
CROSS JOIN book     -- nooooooo
CROSS JOIN author   -- noooooooooooooo
INNER JOIN author a ON (b.author_id=a.id) 
WHERE b.title = :p1 AND a.first_name = :p2'
```

Same happened with regular query objects. 

Turns out, around line 2100 in ModelCriteria.php, where column literals are resolved, checking if the aliased table name should be used was only done with phpNames, not normal names. This lead to the table being added twice to the from clause, once aliased and once normally, the latter without a join condition, causing the cross join.

Test cases only checked for phpNames, so it wasn't detected.

I kept running into this for quite a while, so I am happy to squash that bugger. Please, please make it go away for good.

### DIFF
--- a/src/Propel/Runtime/ActiveQuery/Criteria.php
+++ b/src/Propel/Runtime/ActiveQuery/Criteria.php
@@ -255,7 +255,7 @@ class Criteria
     /**
      * Default operator for combination of criterions
      *
-     * @see addUsingOperator
+     * @see addUsingOperator()
      * @var string Criteria::LOGICAL_AND or Criteria::LOGICAL_OR
      */
     protected $defaultCombineOperator = Criteria::LOGICAL_AND;
--- a/src/Propel/Runtime/ActiveQuery/ModelCriteria.php
+++ b/src/Propel/Runtime/ActiveQuery/ModelCriteria.php
@@ -2014,23 +2014,24 @@ class ModelCriteria extends BaseModelCriteria
         $key = $matches[0];
         [$column, $realFullColumnName] = $this->getColumnFromName($key);
 
-        if ($column instanceof ColumnMap) {
-            $this->replacedColumns[] = $column;
-            $this->foundMatch = true;
-
-            if (strpos($key, '.') !== false) {
-                [$tableName, $columnName] = explode('.', $key);
-                $realColumnName = substr($realFullColumnName, strrpos($realFullColumnName, '.') + 1);
-                if (isset($this->aliases[$tableName])) {
-                    //don't replace a alias with their real table name
-                    return $this->quoteIdentifier($tableName . '.' . $realColumnName);
-                }
-            }
-
-            return $this->quoteIdentifier($realFullColumnName);
+        if (!$column instanceof ColumnMap) {
+            return $this->quoteIdentifier($key);
         }
 
-        return $this->quoteIdentifier($key);
+        $this->replacedColumns[] = $column;
+        $this->foundMatch = true;
+
+        if (strpos($key, '.') !== false) {
+            [$tableName, $columnName] = explode('.', $key);
+            if (isset($this->aliases[$tableName])) {
+                //don't replace a alias with their real table name
+                $realColumnName = substr($realFullColumnName, strrpos($realFullColumnName, '.') + 1);
+
+                return $this->quoteIdentifier($tableName . '.' . $realColumnName);
+            }
+        }
+
+        return $this->quoteIdentifier($realFullColumnName);
     }
 
     /**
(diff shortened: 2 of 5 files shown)


src/Propel/Runtime/ActiveQuery/Criteria.php
```diff
@@ -255,7 +255,7 @@ class Criteria
     /**
      * Default operator for combination of criterions
      *
-     * @see addUsingOperator
+     * @see addUsingOperator()
```
mringler, line 258: cleanup on the way, not really part of the bugfix

src/Propel/Runtime/ActiveQuery/ModelCriteria.php
```diff
@@ -2013,23 +2013,24 @@ protected function doReplaceNameInExpression($matches)
         $key = $matches[0];
         [$column, $realFullColumnName] = $this->getColumnFromName($key);
 
-        if ($column instanceof ColumnMap) {
-            $this->replacedColumns[] = $column;
-            $this->foundMatch = true;
+        if (!$column instanceof ColumnMap) {
```
mringler, line 2016: Couldn't read the method, so I untangled it. Again, not really part of the bugfix. Left it in because it is less convoluted.

src/Propel/Runtime/ActiveQuery/ModelCriteria.php
```diff
@@ -2088,20 +2089,17 @@ protected function getColumnFromName($phpName, $failSilently = true)
             throw new UnknownModelException(sprintf('Unknown model, alias or table "%s"', $prefix));
         }
 
-        if ($tableMap->hasColumnByPhpName($phpName)) {
```
mringler, line 2091: this is the actual fix. Note old lines 2102f, where $realColumnName was always set to the name from the table, while $realColumnName was set to the aliased table in line 2094f (now 2097f), but only if column name was a phpName

mringler (commented):
short run through the changes, if it helps

src/Propel/Runtime/ActiveQuery/ModelCriteria.php
```diff
@@ -2088,20 +2089,17 @@ protected function getColumnFromName($phpName, $failSilently = true)
...
-            $column = $tableMap->getColumnByPhpName($phpName);
+        $column = ($tableMap->hasColumnByPhpName($phpName)) ? $tableMap->getColumnByPhpName($phpName)
```
stereomon, line 2092: Could you extract that in a more readable method?

mringler: Added a method `findColumnByName()` on TableMap. I also changed the variable name from `$phpName` to `$columnName`, apparently it does not have to be a phpName.

tests/Propel/Tests/Runtime/Map/TableMapTest.php
```diff
@@ -58,7 +58,7 @@ public function testConstructor()
         try {
             $tmap = new TableMap();
             $this->assertTrue(true, 'A table map can be instantiated with no parameters');
-        } catch (Exception $e) {
+        } catch (\Exception $e) {
```
dereuromark, line 61: Can we revert this?
It should pick it up from the use statement above, no?

mringler: Removed the backslash and added the missing `use`. Better?

dereuromark (approved):

stereomon (approved):

mringler:
@stereomon Thanks for the review/approval!

@dereuromark Thank you for merging!

What's up with Codecov recently? According to the page, this PR reduces coverage by 0.03% in one file, because I removed more lines than I added. Fair enough. But looking at the report, I apparently screwed up coverage across the whole project, mostly in unrelated files, like SqliteSchemaParser. Saw the same behavior in another PR.
Looking at the [list of commits in the project page](https://app.codecov.io/gh/propelorm/Propel2/commits?page=1), there are lots of commits marked with `unable to find commit on github`, including this  PR's base commit (see [here](https://app.codecov.io/gh/propelorm/Propel2/compare/1704/commits)). Maybe related?

dereuromark:
Maybe some codecov config is set wrongly?

mringler:
Maybe. Though googling "codecov unable to find commit on github waiting for ci" finds Codecov support tickets, that don't indicate user errors.

PR #1705, opened 2021-03-01
Feature: Add exists operator

mringler:
Did you realize, there is no possibility to do EXISTS in Propel, except through `->where('EXISTS (...))`, which has its own caveats (buggy, no query logic). So I had a look at how this probably should work, and it actually went ahead pretty quickly.
These changes allow to create exist queries in two ways:

First one is through regular useQuery logic, same as with the other high-level Propel statements. The relation condition is added to the inner query (in the example author.Id = book.author_id):  
```php
BookQuery::create()
->useAuthorExistsQuery()    // or ->useExistsQuery('Author')
  ->filterBy...
->endUse()
```

Second one is more basic, where a nested query is set as exists query. Here, any table can be used, but the filter in the exists-query must be added manually:
```php
$existsQuery = BookQuery::create()
->filterBy...
->where('Book.AuthorId = Author.Id'); // own join condition in where

AuthorQuery::create()
->whereExists($existsQuery)
```
As exists is such a basic part of SQL, I feel like there are lots of ways to use it which will not work with this implementation (yet). So it might need more refinement over time. Tests run through fine, but if you find anything, please let me know. Same goes for any feedback, of course.

### DIFF
--- a/src/Propel/Generator/Builder/Om/QueryBuilder.php
+++ b/src/Propel/Generator/Builder/Om/QueryBuilder.php
@@ -13,6 +13,7 @@ use Propel\Generator\Model\CrossForeignKeys;
 use Propel\Generator\Model\ForeignKey;
 use Propel\Generator\Model\PropelTypes;
 use Propel\Generator\Model\Table;
+use Propel\Runtime\ActiveQuery\Criterion\ExistsCriterion;
 
 /**
  * Generates a base Query class for user object model (OM).
@@ -1550,6 +1551,7 @@ abstract class " . $this->getUnqualifiedClassName() . ' extends ' . $parentClass
 
         $this->addUseRelatedQuery($script, $fkTable, $queryClass, $relationName, $joinType);
         $this->addWithRelatedQuery($script, $fkTable, $queryClass, $relationName, $joinType);
+        $this->addUseRelatedExistsQuery($script, $fkTable, $queryClass, $relationName);
     }
 
     /**
@@ -1570,6 +1572,7 @@ abstract class " . $this->getUnqualifiedClassName() . ' extends ' . $parentClass
 
         $this->addUseRelatedQuery($script, $fkTable, $queryClass, $relationName, $joinType);
         $this->addWithRelatedQuery($script, $fkTable, $queryClass, $relationName, $joinType);
+        $this->addUseRelatedExistsQuery($script, $fkTable, $queryClass, $relationName);
     }
 
     /**
@@ -1604,6 +1607,59 @@ abstract class " . $this->getUnqualifiedClassName() . ' extends ' . $parentClass
             ->useQuery(\$relationAlias ? \$relationAlias : '$relationName', '$queryClass');
     }
 ";
+    }
+
+    /**
+     * Adds a useExistsQuery and useNotExistsQuery to the object script.
+     *
+     * @param string $script The script will be modified in this method.
+     * @param \Propel\Generator\Model\Table $fkTable The target of the relation
+     * @param string $queryClass Query object class name that will be returned by the exists statement.
+     * @param string $relationName Name of the relation
+     *
+     * @return void
+     */
+    protected function addUseRelatedExistsQuery(&$script, Table $fkTable, $queryClass, $relationName)
+    {
+        $relationDescription = ($relationName === $fkTable->getPhpName()) ?
+            "relation to $relationName table" :
+            "$relationName relation to the {$fkTable->getPhpName()} table";
+
+        $notExistsType = ExistsCriterion::TYPE_NOT_EXISTS;
+        $existsType = ExistsCriterion::TYPE_EXISTS;
+
+        $script .= <<< EOT
+    /**
+     * Use the $relationDescription for an EXISTS query.
+     *
+     * @see \Propel\Runtime\ActiveQuery\ModelCriteria::useExistsQuery()
+     *
+     * @param string|null \$queryClass Allows to use a custom query class for the exists query, like ExtendedBookQuery::class
+     * @param string|null \$modelAlias sets an alias for the nested query
+     * @param string \$typeOfExists Either ExistsCriterion::TYPE_EXISTS or ExistsCriterion::TYPE_NOT_EXISTS
+     *
+     * @return $queryClass The inner query object of the EXISTS statement
+     */
+    public function use{$relationName}ExistsQuery(\$modelAlias = null, \$queryClass = null, \$typeOfExists = '$existsType')
+    {
+        return \$this->useExistsQuery('$relationName', \$modelAlias, \$queryClass, \$typeOfExists);
+    }
+
+    /**
+     * Use the $relationDescription for a NOT EXISTS query.
+     *
+     * @see use{$relationName}ExistsQuery()
+     *
+     * @param string|null \$modelAlias sets an alias for the nested query
+     * @param string|null \$queryClass Allows to use a custom query class for the exists query, like ExtendedBookQuery::class
+     *
+     * @return $queryClass The inner query object of the NOT EXISTS statement
+     */
+    public function use{$relationName}NotExistsQuery(\$modelAlias = null, \$queryClass = null)
+    {
+        return \$this->useExistsQuery('$relationName', \$modelAlias, \$queryClass, '$notExistsType');
+    }
+EOT;
     }
 
     /**
--- a/src/Propel/Runtime/ActiveQuery/BaseModelCriteria.php
+++ b/src/Propel/Runtime/ActiveQuery/BaseModelCriteria.php
@@ -29,6 +29,11 @@ class BaseModelCriteria extends Criteria implements IteratorAggregate
     protected $modelTableMapName;
 
     /**
+     * @var bool
+     */
+    protected $useAliasInSQL = false;
+
+    /**
      * @var string|null
      */
     protected $modelAlias;
@@ -258,6 +263,22 @@ class BaseModelCriteria extends Criteria implements IteratorAggregate
     public function getTableMap()
     {
         return $this->tableMap;
+    }
+
+    /**
+     * Returns the name of the table as used in the query.
+     *
+     * Either the SQL name or an alias.
+     *
+     * @return string
+     */
+    public function getTableNameInQuery()
+    {
+        if ($this->useAliasInSQL && $this->modelAlias) {
+            return $this->modelAlias;
+        }
+
+        return $this->getTableMap()->getName();
     }
 
     /**
--- a/src/Propel/Runtime/ActiveQuery/Criterion/ExistsCriterion.php
+++ b/src/Propel/Runtime/ActiveQuery/Criterion/ExistsCriterion.php
@@ -1,0 +1,95 @@
+<?php
+
+/**
+ * MIT License. This file is part of the Propel package.
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Propel\Runtime\ActiveQuery\Criterion;
+
+use Propel\Runtime\ActiveQuery\ModelJoin;
+use Propel\Runtime\Map\RelationMap;
+
+/**
+ * Specialized Criterion used for EXISTS
+ */
+class ExistsCriterion extends AbstractCriterion
+{
+    public const TYPE_EXISTS = 'EXISTS';
+
+    public const TYPE_NOT_EXISTS = 'NOT EXISTS';
+
+    /**
+     * The inner query of the exists
+     *
+     * @var \Propel\Runtime\ActiveQuery\ModelCriteria
+     */
+    private $existsQuery;
+
+    /**
+     * Build NOT EXISTS instead of EXISTS
+     *
+     * @var string $keyword Either ExistsCriterion::TYPE_EXISTS or ExistsCriterion::TYPE_NOT_EXISTS
+     *
+     * @phpstan-var ExistsCriterion::TYPE_*
+     */
+    private $typeOfExists = self::TYPE_EXISTS;
+
+    /**
+     * @phpstan-param ExistsCriterion::TYPE_*|null $typeOfExists
+     *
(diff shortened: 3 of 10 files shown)


mringler:
Hmm, test  don't run through because import of BookQuery and Author query fails. Womp womp. Any suggestions?

tests/Propel/Tests/Runtime/ActiveQuery/ExistsTest.php
```diff
@@ -0,0 +1,277 @@
...
+
+    private function createAuthor(string $lastName, string $firstname = 'LeDefaultFirstname')
+    {
+        $author = new Author();
+        $author->setLastName($lastName)->setFirstName($firstname)->save($this->con);
+
+        return $author;
+    }
+
+    private function createBook(Author $author, $isGood = false)
```
nederdirk, line 269: A very minor remark: the test would tell its 'story' better, when the `$isGood` boolean parameter would be replaced with a `string $title`

nederdirk (commented):
Hey @mringler, I wish this feature was available a few weeks ago (when I needed it in a project)! I have a few remarks, mostly regarding code clarity and readability. I'll leave it up to the maintainers to choose whether to require changes.

Thanks again!

mringler (commented):
Hey @nederdirk, thank you for your review and your kind words! I agree with all your points, and make adjustments accordingly. Please let me know if there are other issues.

src/Propel/Runtime/ActiveQuery/Criterion/ExistsCriterion.php
```diff
@@ -0,0 +1,90 @@
...
+     * @var \Propel\Runtime\ActiveQuery\ModelCriteria
+     */
+    private $existsQuery;
+
+    /**
+     * Build NOT EXISTS instead of EXISTS
+     *
+     * @var bool
+     */
+    private $isNotExists = false;
```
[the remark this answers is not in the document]
mringler: Yes, I agree. I thought about it, and then was admittedly too lazy to do it. Users shouldn't have to fumble with that anyway, but still. Interesting tip about the `@phpstan-var` parameter. Didn't know about that. Hope I am doing it right

src/Propel/Runtime/ActiveQuery/ModelCriteria.php
```diff
@@ -2013,23 +2100,24 @@ protected function doReplaceNameInExpression($matches)
         $key = $matches[0];
         [$column, $realFullColumnName] = $this->getColumnFromName($key);
 
-        if ($column instanceof ColumnMap) {
-            $this->replacedColumns[] = $column;
-            $this->foundMatch = true;
+        if (!$column instanceof ColumnMap) {
+            return $this->quoteIdentifier($key);
```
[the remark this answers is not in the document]
mringler: Yes, this is part of a necessary bugfix in another branch (https://github.com/propelorm/Propel2/pull/1704), and actually wasn't even part of the fix, but just cleanup on the way. I still think it was good to clean it up, but the way it ripples through to here is annoying, I agree.

src/Propel/Runtime/ActiveQuery/ModelCriteria.php
```diff
@@ -2088,20 +2176,17 @@ protected function getColumnFromName($phpName, $failSilently = true)
             throw new UnknownModelException(sprintf('Unknown model, alias or table "%s"', $prefix));
         }
 
-        if ($tableMap->hasColumnByPhpName($phpName)) {
-            $column = $tableMap->getColumnByPhpName($phpName);
+        $column = ($tableMap->hasColumnByPhpName($phpName)) ? $tableMap->getColumnByPhpName($phpName)
```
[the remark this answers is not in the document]
mringler: I agree once more. This is again part of the other bugfix, and was fixed there. I merged it in, hope this doesn't complicate further merges. Again, let me say just for relieve: it sucks to depend on an unmerged branch.

tests/Propel/Tests/Runtime/ActiveQuery/Criterion/ExistsCriterionTest.php
```diff
@@ -0,0 +1,94 @@
...
+    public function testBuildsRelationConditionDuringInit()
+    {
+        $authorQuery = AuthorQuery::create();
+        $bookQuery = BookQuery::create();
+        $bookRelationMap = $authorQuery->getTableMap()->getRelation('Book');
+        new ExistsCriterion($authorQuery, $bookQuery, false, $bookRelationMap);
+        $params = [];
+        $bookSql = $bookQuery->createSelectSql($params);
+
+        $this->assertEquals('SELECT  FROM book WHERE author.id=book.author_id', $bookSql);
```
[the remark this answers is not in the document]
mringler: Yes, it comes from accessing the inner statement directly, and not through the ExistsFilter. When the ExistsFilter generates SQL, it sets a `SELECT 1 AS existsFlag`, but I wanted to test more isolated.
I would argue that the Criteria object should set the table columns by default, and never return incorrect SQL, but that is a different problem.

tests/Propel/Tests/Runtime/ActiveQuery/ExistsTest.php
mringler: That would be more readable by itself, yes. In this particular case, I honestly like it better this way, as it makes the `createTestData()` method less cluttered, and allows you to see the whole data setup in just a glance, which was helpful when writing the tests. When you know what to look for, you can identify the number of "good" books in about a second.

Now that I have written the tests and don't rely on the short glace reminder anymore, I don't mind changing it to a more verbose signature, if you insist

src/Propel/Runtime/ActiveQuery/Criterion/ExistsCriterion.php
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * @phpstan-param ExistsCriterion::TYPE_*|null $type
+     *
+     * @param \Propel\Runtime\ActiveQuery\ModelCriteria $outerQuery
+     * @param \Propel\Runtime\ActiveQuery\ModelCriteria $existsQuery
+     * @param string|null $type Either ExistsCriterion::TYPE_EXISTS or ExistsCriterion::TYPE_NOT_EXISTS
+     * @param \Propel\Runtime\Map\RelationMap|null $relationMap where outer query is on the left side
+     */
+    public function __construct($outerQuery, $existsQuery, ?string $type = null, ?RelationMap $relationMap = null)
```
nederdirk, line 47: Am I right that the `$type` parameter could be improved by having the default value be `self::TYPE_EXISTS` ? (looking at line 51 below)

src/Propel/Runtime/ActiveQuery/Criterion/ExistsCriterion.php
```diff
@@ -0,0 +1,98 @@
...
+    protected function appendPsForUniqueClauseTo(&$sb, array &$params)
+    {
+        $existsQuery = $this->existsQuery
+            ->clearSelectColumns()
+            ->addAsColumn('existsFlag', '1')
+            ->createSelectSql($params);
+        if ($this->type === self::TYPE_NOT_EXISTS) {
+            $sb .= 'NOT ';
+        }
+        $sb .= 'EXISTS (' . $existsQuery . ')';
```
nederdirk, line 76: This string-building operation can probably be simplified as such: `$sb .= $this->type . ' (' . $existsQuery . ')';`

tests/Propel/Tests/Runtime/ActiveQuery/ExistsTest.php
nederdirk: The PHP interpreter doesn't mind either way, but it will save an interested developer/maintainer in a few months some clicks and thinking when scanning `createTestData`. In general I would say it's worth the effort to strive for maximum clarity

nederdirk (commented):
Hey @mringler, I'm sorry for letting this wait so long. With your latest changes this PR is getting even better. With a few more things to think about or change we can probably land this plane!

src/Propel/Runtime/ActiveQuery/Criterion/ExistsCriterion.php
mringler: Since the input parameter can be any string, I have to check if it is one of the expected ones anyway. Think of `new ExistsCriterion($outer, $inner, 'EXERSTST')`. So line 51 has to stay there anyway, and this comes down to if you think null is a better or worse default value. I think better, as it has less overhead.

src/Propel/Runtime/ActiveQuery/Criterion/ExistsCriterion.php
mringler: Right, I'll change the TYPE_NOT_EXISTS so that this works

src/Propel/Generator/Builder/Om/QueryBuilder.php
```diff
@@ -1606,6 +1608,55 @@ public function use" . $relationName . 'Query($relationAlias = null, $joinType =
...
+     *
+     * @see \Propel\Runtime\ActiveQuery\ModelCriteria::useExistsQuery()
+     *
+     * @param string|null \$queryClass Allows to use a custom query class for the exists query
+     * @param string|null \$modelAlias sets an alias for the nested query
+     * @param bool \$isNotExists do NOT EXISTS instead of EXISTS
+     *
+     * @return $queryClass The inner query object of the EXISTS statement
+     */
+    public function use{$relationName}ExistsQuery(\$modelAlias = null, \$queryClass = null, \$isNotExists = false)
```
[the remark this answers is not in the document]
mringler: Actually, it doesn't even work with a bool anymore, so I have to change it to the string input. Same for the generated useNotExists.

src/Propel/Runtime/ActiveQuery/Criterion/ExistsCriterion.php
mringler: Thinking about it, having this depend on the type token being the correct SQL operator is not optimal, but now I have changed it already and it probably won't ever matter

tests/Propel/Tests/Runtime/ActiveQuery/Criterion/ExistsCriterionTest.php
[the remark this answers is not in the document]
mringler: Well, if said poor developer doesn't know how unfinished propel statements look like, this is probably as good a place as any to get familiar with it. I am more worried about the poor developer who repairs the ill behavior with the empty select columns, and has to update all the tests (try`grep -Hrn 'SELECT  FROM' tests/ | wc -l`). Though this probably won't happen.
On that note, I am curious, can you tell me how much Propel is still used? Looking at the state of the code, the state of the test cases, and how it seems to be stuck in 2.0@dev indefinitely, it seems like the project is on end-of-life support. Is that accurate? Would be a shame, I like a lot of it.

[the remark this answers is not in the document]
mringler: Interesting, thank you

nederdirk (approved):

src/Propel/Runtime/ActiveQuery/BaseModelCriteria.php
```diff
@@ -28,6 +28,11 @@ class BaseModelCriteria extends Criteria implements IteratorAggregate
      */
     protected $modelTableMapName;
 
+    /**
+     * @var bool
+     */
+    protected $useAliasInSQL = false;
```
dereuromark, line 34: We try to use same style for acronyms as for the rest of naming: `...InSql`
Is that possible here too?

src/Propel/Runtime/ActiveQuery/BaseModelCriteria.php
```diff
@@ -261,6 +266,22 @@ public function getTableMap()
...
+    /**
+     * Returns the name of the table as used in the query.
+     *
+     * Either the SQL name or an alias.
+     *
+     * @return string
+     */
+    public function getTableNameInQuery()
+    {
+        if ($this->useAliasInSQL && !empty($this->modelAlias)) {
```
dereuromark, line 278: Can we use non cloaking

    && $this->modelAlias

instead?

src/Propel/Runtime/ActiveQuery/Criterion/ExistsCriterion.php
```diff
@@ -0,0 +1,95 @@
...
+        $sb .= $this->typeOfExists . ' (' . $existsQuery . ')';
+    }
+
+    /**
+     * @param \Propel\Runtime\ActiveQuery\ModelCriteria $outerQuery
+     * @param \Propel\Runtime\Map\RelationMap $relationMap where outer query is on the left side
+     *
+     * @return \Propel\Runtime\ActiveQuery\Criterion\AbstractCriterion
+     */
+    private function buildJoinCondition($outerQuery, RelationMap $relationMap)
```
dereuromark, line 82: Is there a reason not to have this protected?

src/Propel/Runtime/ActiveQuery/BaseModelCriteria.php
dereuromark: Oh, I see you moved this from somewhere else. So we better keep the same style then for now.

mringler: Yes, I remember, that was a property defined in child class BaseModelCriteria, but then accessed in the parent class through `static`, and then stan complained about it (rightly so). And even worse, my editor couldn't find it, so I cleaned it up

src/Propel/Runtime/ActiveQuery/BaseModelCriteria.php
mringler: Yes, good point. Changed it

src/Propel/Runtime/ActiveQuery/Criterion/ExistsCriterion.php
mringler: Agreed, protected is better

PR #1706, opened 2021-03-03
Bugfix: reset partial flag when populating a relation

mringler:
Another day, another merge request. Hope this isn't unwelcome.

This is a small one: Calling `ObjectCollection->populateRelation()` does not reset the partial flag, which leads to consecutive queries on an already initialized relation. So if a data object with a partially loaded relation is retrieved from instance pooling, `populateRelation()` loads the relation data, but when accessing the data, another reload is triggered.

Simple example:
```php
$book = BookQuery::create()
->leftJoinWithAuthor() // will create a partially loaded book relation on autor object
->findOne();

$authors = AuthorQuery::create()->find(); // already loaded author object will be retrieved from cache 
$authors->populateRelation('Book'); // loads all books

$book->getAuthor()->getBooks(); // loads books again when relation is still marked as partial
```

**Why this matters:** Apart from being inconsistent, this significantly reduces performance in some scenarios, where the bug can easily cause 100+ unexpected queries.
For example, in my project, I have patients and their relatives. You can search in both groups at the same time. The generated view lists for each patient (among other data) the name of the patient's relatives, and relatives do the same with patient names. So I do something like this:
```php
$patients = PatientQuery::create()->findBy($searchTerm);
$patients->populateRelation('Relatives');     // Prefetch associated relatives, they are initialized with a partial "Patient" relation 

$relatives = RelativesQuery::create()->findBy($searchTerm); // Some of them are already cached and won't be re-initialized
$relatives->populateRelation('Patients');     // All patients are loaded, but "partial" flag is not updated
...
foreach($relatives as $relative) {
    $relative->getPatients(); // Causes an additional query for each relative where "Patient" relation is marked partial, despite populateRelation
}
```
In my case, this causes literally hundreds of unnecessary queries, even though I specifically use `populateRelation` to avoid a n+1 scenario.

### DIFF
--- a/src/Propel/Runtime/Collection/ObjectCollection.php
+++ b/src/Propel/Runtime/Collection/ObjectCollection.php
@@ -379,8 +379,10 @@ class ObjectCollection extends Collection
         if ($relationMap->getType() === RelationMap::ONE_TO_MANY) {
             // initialize the embedded collections of the main objects
             $relationName = $relationMap->getName();
+            $resetPartialStatusMethod = 'resetPartial' . $relationMap->getPluralName();
             foreach ($this as $mainObj) {
                 $mainObj->initRelation($relationName);
+                $mainObj->$resetPartialStatusMethod(false);
             }
             // associate the related objects to the main objects
             $getMethod = 'get' . $symRelationMap->getName();
--- a/tests/Propel/Tests/Runtime/Collection/ObjectCollectionTest.php
+++ b/tests/Propel/Tests/Runtime/Collection/ObjectCollectionTest.php
@@ -15,6 +15,7 @@ use Propel\Runtime\Exception\RuntimeException;
 use Propel\Runtime\Formatter\ObjectFormatter;
 use Propel\Runtime\Propel;
 use Propel\Tests\Bookstore\Author;
+use Propel\Tests\Bookstore\AuthorQuery;
 use Propel\Tests\Bookstore\Book;
 use Propel\Tests\Bookstore\Country;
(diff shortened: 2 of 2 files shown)


dereuromark (approved):

dereuromark:
Since this can potentially be somewhat bc breaking (as behavior might change based on the so far incorrect but expected result), this is good to have outlined in the changelog explicitly then.

mringler:
That certainly can't hurt. It's possible that someone did not use `populateRelation()` when they thought it was not working.

Otherwise, I am pretty sure the results are the same as before, except in one very specific case, when  a partially loaded relation is populated with a filtered query:
```php
$book = BookQuery::create()->leftJoinWithAuthor()->findOne();
$author = $book->getAuthor(); // this author has a partially loaded book relation
authors = AuthorQuery::create()->find(); // above author is loaded from cache with partial relation
$filterQuery = BookQuery::create()->filterByPrice(['min' => 100]); // filter for only some books
$authors->populateRelation('Book', $filterQuery); // populate book relation with above filter 

$author->getBooks(); // this gives a different result now
```
Currently, the last line returns all books by the author, since the partial status triggers a full reload. For all other authors, `getBooks()` returns only the filtered books as expected. It looks like the filter argument from `populateRelation()` does not work consistently.
With the fix, there is no reload anymore, and all authors return their filtered books.

So I think this can only cause trouble if someone had loaded all elements with a partial relation, and then populated with a filter query, but relied on that filter not working. That someone would have egg on their face now, when the filter query suddenly works.

stereomon (approved):

PR #1708, opened 2021-03-05
Tests cleanup

mringler:
Some of the tests were "risky" when they did not perform any assertions, and phpunit cluttered the output with warning. Most of them check there is no exception, so I marked them with `@doesNotPerformAssertions`. Two tests were not finished, so I added `$this->markTestIncomplete();`.

One test failed with MySQL because it did not account for MySQL 8, which removed display size for int (as in INT(4), see https://stackoverflow.com/questions/60892749/mysql-8-ignoring-integer-lengths). Simple fix is to use VARCHAR instead.

Two tests failed with MySQL (and probably with Oracle, too) because of `Expression of SELECT list is not in GROUP BY clause`. I think this is a bug in Propel, which should not force users to change server mode, but since these tests are about something else, I only change the server mode locally in those tests. 

I did not change the test themselves, except for `PropelPDOTest.php`, where connection initialization unnecessarily deviated from how it is done in parent class.

One test is still printed as risky, as the test is skipped as part of an unresolved problem (according to the comment there).

Wish I (or someone else) had done this earlier, working with dirty tests is not good.

### DIFF
--- a/tests/Propel/Tests/Generator/Builder/Om/GeneratedQueryDoDeleteTest.php
+++ b/tests/Propel/Tests/Generator/Builder/Om/GeneratedQueryDoDeleteTest.php
@@ -8,6 +8,7 @@
 
 namespace Propel\Tests\Generator\Builder\Om;
 
+use Exception;
 use Propel\Runtime\ActiveQuery\Criteria;
 use Propel\Runtime\Propel;
 use Propel\Tests\Bookstore\AuthorQuery;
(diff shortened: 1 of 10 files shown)


tests/Propel/Tests/Generator/Builder/Om/GeneratedQueryDoDeleteTest.php
```diff
@@ -8,6 +8,7 @@
 
 namespace Propel\Tests\Generator\Builder\Om;
 
+use Exception;
```
nederdirk, line 11: And this `use`-statement also seems to be unused

nederdirk (commented):
Hey @mringler, this looks good! I found a few unused `use`-statements, but those are **very** minor issues :smile:

PR #1710, opened 2021-03-06
make primary table always available in query

mringler:
So, I am trying to get the `ModelCriteria::addSelectQuery()` to work, but there are several issues. Guess my sabbatical starts to bore me. Time to do this for money again soon.

So far, the issues I have found with `ModelCriteria::addSelectQuery()`:

- primary table is not set, leading to queries without `FROM`
- columns from `ModelCriteria::select()` are not resolved`, so selection does not work, causing "nonaggregated columns" error in nested group by
- using `ModelCriteria::select()` to join with a subquery is quirky (a statement like `A.id = B.id` will resolve all columns by table A, so if table A is the subquery, the nested table is used even though it is not available, and if table B is the suquery, its columns are resolved by the outer table, causing naming collisions)

This addresses the first issue, when queries are generated without a `FROM` clause:
Looking at the code, the same problems occurs at several positions, when no `WHERE` clause was given, and was fixed locally, instead of going to the root. Quick fix would have been to add the same workaround to nested queries, but the simple solution is to set the primary table during initialization of `BaseModelCriteria`, which has the benefit of making the other workarounds obsolete.
Only a few lines of code need to be changed, thought a lot of tests required adjustment, when suddenly (more) correct queries were returned.

### DIFF
--- a/src/Propel/Runtime/ActiveQuery/BaseModelCriteria.php
+++ b/src/Propel/Runtime/ActiveQuery/BaseModelCriteria.php
@@ -158,23 +158,28 @@ class BaseModelCriteria extends Criteria implements IteratorAggregate
      * Sets the model name.
      * This also sets `this->modelTableMapName` and `this->tableMap`.
      *
-     * @param string $modelName
+     * @param string|null $modelName
      *
      * @return $this The current object, for fluid interface
      */
     public function setModelName($modelName)
     {
-        if (strpos($modelName, '\\') === 0) {
-            $this->modelName = substr($modelName, 1);
-        } else {
-            $this->modelName = $modelName;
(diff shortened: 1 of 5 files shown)


nederdirk (approved):

mringler (commented):

dereuromark (approved):

gechetspr:
This changes looks like the same as in https://github.com/propelorm/Propel2/pull/1711
Can we close it in this case?

mringler:
That's right, I had to include the commit in the other PR. Back then, I thought you were going to process them in order, so this would go through first, and then I'd rebase the other PR, to keep the two separated. Didn't work out that way. I'm going to close this one then.

PR #1711, opened 2021-03-06
Bugfix/always configure selected columns

mringler:
Again, trying to get `ModelCriteria::addSelectQuery()`: working.

 So far, the issues I have found with `ModelCriteria::addSelectQuery()`:

- primary table is not set, leading to queries without `FROM` (addressed [here](https://github.com/propelorm/Propel2/pull/1710))
- columns from `ModelCriteria::select()` are not resolved`, so selection does not work, causing "nonaggregated columns" error in nested group by
- using `ModelCriteria::select()` to join with a subquery is quirky (a statement like `A.id = B.id` will resolve all columns by table A, so if table A is the subquery, the nested table is used even though it is not available, and if table B is the suquery, its columns are resolved by the outer table, causing naming collisions)

This addresses the second issue, when calling `->select(['ColA', 'ColB'])` is not resolved, and all columns are selected regardless.
This is a weird issue: For some reason, calling `->select()` just stored the selected columns, and the actual selection was pushed into its own function, which had to be called manually (and wasn't called for nested queries). Quick fix would have been to figure out where to call it on the subquery, however, there is really no benefit in pushing the selection downstream. Instead, you have to remember to finish initialization, and you have to deal with another class variable at several places in the code. Also, column order was not respected when doing `->select(...)->addAsColumn(...)`, where the added column was put at the beginning.
Just resolving the selection right away solves the issue, and reduces complexity overall.

Please note that this is based on the branch of [the other PR](https://github.com/propelorm/Propel2/pull/1710), which avoids a tiny yet annoying merge conflict. **Please review the commit**, instead of the whole branch.


src/Propel/Runtime/ActiveQuery/ModelCriteria.php
```diff
@@ -2144,48 +2128,9 @@ public function doSelect(?ConnectionInterface $con = null)
...
 
-        $this->configureSelectColumns();
-
         return parent::doSelect($con);
     }
 
-    /**
-     * @return void
-     */
-    public function configureSelectColumns()
```
nederdirk, line 2155: It's great that this function becomes superfluous, but since it's part of the public interface of a crucial class, it's probably better to keep it around as a no-op, with an `@deprecated` phpdoc tag.

src/Propel/Runtime/ActiveQuery/ModelCriteria.php
```diff
@@ -440,29 +434,36 @@ public function select($columnArray)
...
+        }
 
-    /**
-     * Retrieves the columns defined by a previous call to select().
-     *
-     * @see select()
-     *
-     * @return array|string A list of column names (e.g. array('Title', 'Category.Name', 'c.Content')) or a single column name (e.g. 'Name')
-     */
-    public function getSelect()
```
nederdirk, line 463: This function is also part of the public interface of this library, and we shouldn't remove it haphazardly.

nederdirk (changes requested):
Hey @mringler, I _think_ this looks great. I also think we should take a small step back with removing functions from the public interface of this library, so I'd advise returning `configureSelectColumns` and `getSelect`, and applying a `@deprecated` phpdoc tag to them.

mringler (commented):

src/Propel/Runtime/ActiveQuery/ModelCriteria.php
```diff
@@ -440,29 +434,50 @@ public function select($columnArray)
         if (empty($columnArray)) {
             throw new PropelException('You must ask for at least one column');
         }
+        $this->isSelfSelected = true;
+        if ($this->formatter === null) {
+            $this->setFormatter('\Propel\Runtime\Formatter\SimpleArrayFormatter');
```
gechetspr, line 439: minor, but we I guess we can use `SimpleArrayFormatter::class` here

gechetspr (approved):
One minor thing, but the rest looks good

src/Propel/Runtime/ActiveQuery/ModelCriteria.php
mringler: That's code I just moved around, but you are right, I should have replaced it (and now I have)

gechetspr (approved):

PR #1715, opened 2021-03-16
Make fromArray() chainable (aka return $this)

mringler:
Just when I thought I was done with Propel.

In general, methods of ActiveRecods are chainable, but `->fromArray()` isn't, meaning you have to write:
```php
$book = BookQuery::create()->findPk($id); // get the object reference
$book->fromArray($data);            // call void method
$book->setChanger($userId)->save();           // finish
```
when you could just do:
```php
BookQuery::create()
  ->findPk($id)
  ->fromArray($data)
  ->setChanger($userId)
  ->save();
```

Some people prefer the former, which is fine, and they can still use it, but lots of folks prefer the latter, lean version, and it is consistent with the other setters.
There really is no reason why the first one should be disallowed.


dereuromark (approved):

dereuromark:
We need more reviewers on this one for it to be mergable.

demkos (approved):

PR #1716, opened 2021-03-18
Bugfix: remove undefined index notice from profiler

mringler:
When using the Profiler to check time and memory usage, php shows three notices:

> **Notice**: Undefined index: microtime in **vendor/propel/propel/src/Propel/Runtime/Util/Profiler.php** on line **242**
> **Notice**: Undefined index: microtime in **vendor/propel/propel/src/Propel/Runtime/Util/Profiler.php** on line  **253**
> **Notice**: Undefined index: memoryUsage in **vendor/propel/propel/src/Propel/Runtime/Util/Profiler.php** on line  **261** 

The problem is outlined [here](https://github.com/propelorm/Propel2/pull/1647#issuecomment-801380568).

If merged, this replaces #1647

### DIFF
--- a/src/Propel/Runtime/Util/Profiler.php
+++ b/src/Propel/Runtime/Util/Profiler.php
@@ -31,9 +31,9 @@ class Profiler
     protected $outerGlue;
 
     /**
-     * @var array
+     * @var array|null
      */
-    protected $snapshot = [];
+    protected $snapshot;
 
     /**
      * @var array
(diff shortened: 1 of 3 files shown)


src/Propel/Runtime/Util/Profiler.php
```diff
@@ -213,7 +213,11 @@ public function isSlow()
      */
     public function getProfile()
     {
-        return $this->getProfileBetween($this->snapshot, self::getSnapshot());
+        $endSnapshot = self::getSnapshot();
+        $startSnapshot = (empty($this->snapshot)) ? $endSnapshot : $this->snapshot;
```
dereuromark, line 217:
```suggestion
        $startSnapshot = ($this->snapshot === null) ? $endSnapshot : $this->snapshot;
```
?
when you make the empty case the null one?

dereuromark (approved):

dereuromark:
Do we have a resolution on the open comments? Is this mergable?

mringler:
> Do we have a resolution on the open comments? Is this mergable?

Marked my walkthrough comments as resolved.

demkos (approved):

PR #1719, opened 2021-04-07
Feature: use heredoc in migration file

mringler:
Found some stuff while cleaning up my project, that other might find useful.

This changes the migration files to use heredoc instead of string literals. The advantage is that you can copy lines from and to a database editor, without having to quote/unquote strings, which I had to do a lot for more complex migrations.

I also moved the migration template from the generating code into its own file, similar to other Propel templates.

With this change, migrations look like this:
```php
    public function getUpSQL()
    {
        $connection_default = <<< 'EOT'
ALTER TABLE `payment`

  CHANGE `status` `status` set('granted','denied','in_submission', 'deleted') NOT NULL;
EOT;

        return array(
            'default' => $connection_default,
        );
    }
```

instead of:
```php
    public function getUpSQL()
    {
        return array (
  'default' => '
ALTER TABLE `payment`

  CHANGE `status` `status` set(\'granted\',\'denied\',\'in_submission\', \'deleted\') NOT NULL;
',
);
    }
```

### DIFF
--- a/src/Propel/Generator/Manager/MigrationManager.php
+++ b/src/Propel/Generator/Manager/MigrationManager.php
@@ -11,6 +11,7 @@ namespace Propel\Generator\Manager;
 use Exception;
 use PDO;
 use PDOException;
+use Propel\Generator\Builder\Util\PropelTemplate;
 use Propel\Generator\Exception\InvalidArgumentException;
 use Propel\Generator\Model\Column;
 use Propel\Generator\Model\Database;
@@ -382,11 +383,12 @@ class MigrationManager extends AbstractManager
     public function getMigrationObject($timestamp)
     {
         $className = $this->getMigrationClassName($timestamp);
-        require_once sprintf(
+        $filename = sprintf(
             '%s/%s.php',
             $this->getWorkingDirectory(),
             $className
         );
+        require_once $filename;
 
         return new $className();
     }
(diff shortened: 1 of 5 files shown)


mringler:
Hmpf, phpstan complains about completely unrelated files. Weird.
Fixed those errors in #1721

src/Propel/Generator/Manager/MigrationManager.php
```diff
@@ -402,72 +404,57 @@ public function getMigrationObject($timestamp)
...
+            foreach ($connectionNames as $index => $connectionName) {
+                if (array_key_exists($connectionName, $connectionToVariableName)) {
+                    continue;
+                }
+                $alphNums = preg_replace('/\W/', '', $connectionName);
+                if (strlen($alphNums) === 0) {
+                    $alphNums = $index;
+                }
+                $variableName = '$connection_' . $alphNums;
+                while (in_array($variableName, $connectionToVariableName)) {
```
dereuromark, line 450: can we use 3rd arg `true` here please?

mringler: Do you mean `in_array($variableName, $connectionToVariableName, true)`?
If so, why do you prefer it? Both `$variableName` and `$connectionToVariableName` are guaranteed to be strings and not null, so I am wondering about the upside of checking against type.

dereuromark: Yes, this helps to communicate this.

mringler: Oh, didn't think you'd respond, so I changed my comment. But yes, makes sense to use strict type checks here. Thanks for the tip!

dereuromark (approved):

stereomon (approved):

PR #1720, opened 2021-04-07
cleanup of migration tests

mringler:
During #1719, the migration command tests broke my test database several times to the point where I had to reset the database. My fault, obviously, but they are written very Ctrl+C/Ctrl+V, and quite fragile. Got frustrated and cleaned them up.


mringler:
Hmpf, phpstan complains about completely unrelated files. Weird.
Fixed those errors in #1721

PR #1722, opened 2021-04-08
Test and fixes for #1491 - resolve external schema location relative to file

mringler:
Had this laying around: Test case for #1491 and some fixes

### DIFF
--- a/src/Propel/Generator/Manager/AbstractManager.php
+++ b/src/Propel/Generator/Manager/AbstractManager.php
@@ -386,7 +386,7 @@ abstract class AbstractManager
      *
      * @throws \Propel\Generator\Exception\BuildException
      *
-     * @return int
+     * @return int number of included external schemas
      */
     protected function includeExternalSchemas(DOMDocument $dom, $srcDir)
     {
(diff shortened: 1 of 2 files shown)


src/Propel/Generator/Manager/AbstractManager.php
```diff
@@ -386,30 +386,33 @@ protected function loadDataModels()
      *
      * @throws \Propel\Generator\Exception\BuildException
      *
-     * @return int
+     * @return int number of included external schemas
      */
-    protected function includeExternalSchemas(DOMDocument $dom, $srcDir)
+    protected function includeExternalSchemas(DOMDocument $dom, string $srcDir): int
```
dereuromark, line 391: Is it a good idea to add the types here? The next release could be another "major" one, otherwise this could be a BC break we might want to avoid.

mringler: Makes sense. I removed the type hints again

dereuromark (approved):

gechetspr (approved):

PR #1713, opened 2021-03-09
Re-submission of #1660: Unable to use Collection suffix for table name

nederdirk:
This is an updated version of #1660

PR #1724, opened 2021-04-09
Tests & Fixes for #1660, #1713 - Bugfix: getCollectionClassName() should only return Collection classes

mringler:
Tests & fixes for #1660, #1713: Simply make sure that `getCollectionClassName()` returns a class extending `Collection`, otherwise bad things can happen.

If merged, this replaces #1660, #1713


nederdirk (approved):
Looks great, thanks for helping out!

mringler:
Hey @dereuromark , you can close #1660 now, it's a dysfunctional fix for the same issue

PR #1727, opened 2021-04-29
Show failed queries and SQL error in debug mode exception

mringler:
Failed queries are shown with placement variables in the Exception:
```
Unable to execute UPDATE statement [UPDATE payment SET receiver_id=:p1, date_changed=:p2 WHERE payment.id=:p3]
```
That makes it often hard to see what went wrong.

To see the database error, you have to check the log:
```
Error: SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'receiver_id' cannot be null
```
The query did not go to the log, but at least it is possible to puzzle together what went wrong.

However, there is a nice formatter available, which writes full queries to the log.

New Exception message looks like this:
```
Unable to execute statement [UPDATE payment SET receiver_id=NULL, date_changed='2021-04-28 22:15:55.450739' WHERE payment.id=1491]
Reason: [SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'receiver_id' cannot be null]
```
There is another PR #1728, which makes sure that failed queries are written to log, too.

### DIFF
--- a/src/Propel/Runtime/ActiveQuery/Criteria.php
+++ b/src/Propel/Runtime/ActiveQuery/Criteria.php
@@ -17,6 +17,8 @@ use Propel\Runtime\ActiveQuery\Criterion\InCriterion;
 use Propel\Runtime\ActiveQuery\Criterion\LikeCriterion;
 use Propel\Runtime\ActiveQuery\Criterion\RawCriterion;
 use Propel\Runtime\Connection\ConnectionInterface;
+use Propel\Runtime\Connection\ConnectionWrapper;
+use Propel\Runtime\Connection\StatementWrapper;
 use Propel\Runtime\Exception\LogicException;
 use Propel\Runtime\Exception\PropelException;
 use Propel\Runtime\Propel;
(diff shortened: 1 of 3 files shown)


src/Propel/Runtime/ActiveQuery/Criteria.php
```diff
@@ -2867,22 +2850,41 @@ public function doSelect(?ConnectionInterface $con = null)
...
+     * @param \Propel\Runtime\Connection\StatementInterface|\PDOStatement $stmt
+     * @param string|null $sql
+     *
+     * @throws \Propel\Runtime\Exception\PropelException
+     *
+     * @return void
+     */
+    protected function handleStatementException(Exception $e, $stmt, ?string $sql = '')
+    {
+        if (is_a($stmt, StatementWrapper::class)) {
```
dereuromark, line 2878: Can't we use instance of here?

src/Propel/Runtime/ActiveQuery/Criteria.php
```diff
@@ -2867,22 +2850,41 @@ public function doSelect(?ConnectionInterface $con = null)
...
+     *
+     * @param \Exception $e
+     * @param \Propel\Runtime\Connection\StatementInterface|\PDOStatement $stmt
+     * @param string|null $sql
+     *
+     * @throws \Propel\Runtime\Exception\PropelException
+     *
+     * @return void
+     */
+    protected function handleStatementException(Exception $e, $stmt, ?string $sql = '')
```
dereuromark, line 2876: If you add a new method, we could add `: void` here now.

src/Propel/Runtime/ActiveQuery/Criteria.php
```diff
@@ -2867,22 +2850,41 @@ public function doSelect(?ConnectionInterface $con = null)
...
+    {
+        if (is_a($stmt, StatementWrapper::class)) {
+            $sql = $stmt->getExecutedQueryString();
+        }
+        $initialMessage = $e->getMessage();
+        Propel::log($initialMessage, Propel::LOG_ERR);
+        $newMessage = sprintf("Unable to execute statement [%s]\nReason: [%s]", $sql ?? '', $initialMessage);
```
dereuromark, line 2883: We can use single quotes here.

src/Propel/Runtime/ActiveQuery/Criteria.php
mringler: Right, yes, good point. Changed it.

src/Propel/Runtime/ActiveQuery/Criteria.php
mringler: Added return type. Thank you for finding!

src/Propel/Runtime/ActiveQuery/Criteria.php
mringler: Then the '\n' wouldn't work. But I changed it so that it doesn't use sprintf anymore

dereuromark: No, the sprintf was fine, I didnt see the \n usage. All good either way.

mringler:
Ok, apparently so database systems return a prepared statement even when the query is faulty (MySQL) and others don't (SQLite, Postgres). However, a prepared statement is needed to get the full query for the exception message. This isn't a problem in itself, because if your query syntax is broken, you do not need to see the values. Those are only interesting when the execution failed, for example due to a constraint violation.
But I had to take apart the tests that checked if an exception is thrown, and that the exception contains the query. This is now done independently, so it can be checked independent of platform.

While trying out stuff to go around the issue, I refactored some lines in `Criteria::doUpdate()`. I left those changes in, even though they are not related to the objective of this PR, simply because it is an improvement. Hope that is alright.

stereomon (changes requested):
AFAIK the current approach was used wisely to not show/output sensitive data. Also, exception messages might be sent over the network and will show up in monitoring tools like NewRelic. We should NOT make the messages more verbose for security reasons.

dereuromark:
@stereomon We would need to make sure that the additional (useful) data is only added in development mode, rather than production mode maybe.

mringler:
Hey @stereomon, that's a good point, hadn't thought of that. I like @dereuromark's idea, to be careful in production mode, but helpful in development/debug mode.
So now the exception message in prod mode is the exact same as before, and in dev mode you get full query and SQL error message. Do you think this will work?

mringler:
I decided to split this into two PRs, to make it more manageable. This PR contains the changes to the exception output that we have been discussing.
Remote branch name doesn't make sense now, but I don't think I can change the branch name of a PR (can I?), and I don't want to lose the discussion.

src/Propel/Runtime/ActiveQuery/Criteria.php
```diff
@@ -2867,22 +2843,48 @@ public function doSelect(?ConnectionInterface $con = null)
...
+     * @throws \Propel\Runtime\Exception\PropelException
+     *
+     * @return void
+     */
+    protected function handleStatementException(Exception $e, ?string $sql, ?ConnectionInterface $con = null, $stmt = null): void
+    {
+        $internalMessage = $e->getMessage();
+        Propel::log($internalMessage, Propel::LOG_ERR);
+
+        $useDebugMode = ($con instanceof ConnectionWrapper && $con->useDebug);
```
stereomon, line 2876: Please extract this into its own method for better readability.

stereomon:
I'm fine with the separation for dev and prod. 

Thank you for your effort!

src/Propel/Runtime/ActiveQuery/Criteria.php
mringler: Sure. Moved the debug check into its own method.

stereomon (approved):

dereuromark (approved):

PR #1728, opened 2021-05-01
Bugfix: Failed queries should be written to log & Fix PdoConnection::exec()

mringler:
When a query fails during execution, only the exception message is written to log, not the query itself. That makes it hard to figure out what went wrong from the log.
With this fix, an exception is stored and then rethrown after logging. The query cannot be logged before execution, as this would give wrong execution time and memory usage in the Profiler.  

### ALSO

I fixed an error in `PdoConnection::exec()`. This is likey not a big deal, since it only fixes something that was broken, but you might want to check because of **backwards compatibility**, hence the double highlighting.

The `PdoConnection::exec()` method wraps `PDO::exec()`, which, according to the [manual](https://www.php.net/manual/en/pdo.exec.php) "does not return results from a SELECT statement". Instead, it returns the number of affected rows.
However, in `PdoConnection`, that returned int was wrapped into a `DataFetcher`, which only works with result sets. So now you can call `fetch()` on an int and admire the exception. I think you can call `DataFetcher::getDataObject()` to get the int back, but it is an ordeal, let alone confusing.
So I removed the DataFetcher and have `PdoConnection::exec()` return the int from `PDO::exec()` directly. If someone went the extra mile to get to the return value using `DataFetcher::getDataObject()`, that does not work anymore. 

I would like to have these two into own PRs, but the tests for query logging depend on the fix in PdoConnection (it's how I found the issue). They are different commits though.

### DIFF
--- a/src/Propel/Runtime/Connection/PdoConnection.php
+++ b/src/Propel/Runtime/Connection/PdoConnection.php
@@ -137,13 +137,11 @@ class PdoConnection implements ConnectionInterface
     /**
      * @inheritDoc
      *
-     * @return \Propel\Runtime\DataFetcher\DataFetcherInterface
+     * @return int
      */
     public function exec($statement)
     {
-        $stmt = $this->pdo->exec($statement);
-
-        return $this->getDataFetcher($stmt);
+        return $this->pdo->exec($statement);
     }
 
     /**
--- a/src/Propel/Runtime/Connection/ConnectionInterface.php
+++ b/src/Propel/Runtime/Connection/ConnectionInterface.php
@@ -148,7 +148,7 @@ interface ConnectionInterface
      * @param string $statement The SQL statement to prepare and execute.
      *                          Data inside the query should be properly escaped.
      *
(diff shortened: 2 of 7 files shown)


src/Propel/Runtime/Connection/PdoConnection.php
```diff
@@ -137,13 +137,11 @@ public function query($statement)
     /**
      * @inheritDoc
      *
-     * @return \Propel\Runtime\DataFetcher\DataFetcherInterface
+     * @return int
```
gechetspr, line 140: could you also adjust an interface if this change is really needed?

gechetspr (commented):
Except the interface everything looks good to me

src/Propel/Runtime/Connection/PdoConnection.php
mringler: Right, yes, good point. Adjusted the interface declaration in ConnectionInterface

dereuromark (approved):

gechetspr (approved):

PR #1729, opened 2021-05-06
fix doubly namespaces during database:reverse / write absolute namespaces to schema.xml

### LINKED ISSUE #1233: Invoking database:reverse incorrectly duplicates namespace in schema.xml table element

mringler:
Lots of users, including me, have stumbled over doubled namespaces after `database:reverse` (see #1233). For example, calling `prople database:reverse --namespace=My\Namespace ...` will lead to a schema like
```xml
<database namespace="My\Namespace">
  <table PhpName="MyTable" namespace="My\Namespace">
  </table>
</database>
```
 which will generate files in the namespace `My\Namespace\My\Namespace`, not the specified `My\Namespace`, and hence the confusion.
 
 The reason for this is actually quite straight forward. Propel differentiates between absolute namespaces and relative namespaces (as documented [here](http://propelorm.org/documentation/cookbook/namespaces.html#namespace-declaration-and-inheritance)). Absolute namespaces begin with a backslash, while relative namespaces do not. So if there is a namespace on the database element (doesn't matter if absolute or relative), and a table has a relative namespace, the names are combined. This is what causes the doubled namespace in above example and the output from `database:reverse`.
 
So, simple workaround is to use an absolute namespace in the reverse command, i.e. `propel database:reverse --namespace=\\My\\Namespace ...` (or `--namespace='\My\Namespace'`, but not `--namespace=\My\Namespace`). However, most user probably don't do a deep dive into the documentation while occupied with reverse engineering (and the issue runs a bit deeper).

Anyway, fix is quite simple: Let XmlDumper write absolute namespaces to schema.xml. Once tables are added to their database, they store the full namespace anyway, just without the leading backslash. Adding those leading backslashes again removes the pitfall of accidentally causing namespace extension. Beneficial side-effect is that users will see how absolute path look like, and probably identify the error should it ever come up again.

Of course, users can still use namespace extension in their schema. And there is no issue with backwards compatibility for existing schemas. Running `database:reverse` will generate a different schema than before, but I cannot think of a scenario where this causes issues with BC. 

If merged, this resolves #1233 and #1235, and fixes #1241

### DIFF
--- a/src/Propel/Generator/Builder/Om/AbstractOMBuilder.php
+++ b/src/Propel/Generator/Builder/Om/AbstractOMBuilder.php
@@ -361,21 +361,23 @@ abstract class AbstractOMBuilder extends DataModelBuilder
      * check if the current $class need an alias or if the class could be used with a shortname without conflict
      *
      * @param string $class
-     * @param string $namespace
+     * @param string $classNamespace
      *
      * @return bool
      */
-    protected function needAliasForClassName($class, $namespace)
+    protected function needAliasForClassName($class, $classNamespace)
     {
-        if ($namespace == $this->getNamespace()) {
+        $builderNamespace = trim($this->getNamespace(), '\\');
+
+        if ($classNamespace == $builderNamespace) {
             return false;
         }
 
-        if (str_replace('\\Base', '', $namespace) == str_replace('\\Base', '', $this->getNamespace())) {
+        if (str_replace('\\Base', '', $classNamespace) == str_replace('\\Base', '', $builderNamespace)) {
             return true;
         }
 
-        if (empty($namespace) && $this->getNamespace() === 'Base') {
+        if (empty($classNamespace) && $builderNamespace === 'Base') {
             if (str_replace(['Query'], '', $class) == str_replace(['Query'], '', $this->getUnqualifiedClassName())) {
                 return true;
             }
@@ -385,14 +387,14 @@ abstract class AbstractOMBuilder extends DataModelBuilder
             }
 
             // force alias for model without namespace
-            if (array_search($class, $this->whiteListOfDeclaredClasses, true) === false) {
+            if (!in_array($class, $this->whiteListOfDeclaredClasses, true)) {
                 return true;
             }
         }
 
-        if ($namespace === 'Base' && $this->getNamespace() === '') {
+        if ($classNamespace === 'Base' && $builderNamespace === '') {
             // force alias for model without namespace
-            if (array_search($class, $this->whiteListOfDeclaredClasses, true) === false) {
+            if (!in_array($class, $this->whiteListOfDeclaredClasses, true)) {
                 return true;
             }
         }
--- a/src/Propel/Generator/Model/ScopedMappingModel.php
+++ b/src/Propel/Generator/Model/ScopedMappingModel.php
@@ -76,10 +76,16 @@ abstract class ScopedMappingModel extends MappingModel
     /**
      * Returns the namespace.
      *
-     * @return string
+     * @param bool $getAbsoluteNamespace
+     *
+     * @return string|null
      */
-    public function getNamespace()
+    public function getNamespace(bool $getAbsoluteNamespace = false): ?string
     {
+        if ($getAbsoluteNamespace) {
+            return $this->makeNamespaceAbsolute($this->namespace);
+        }
+
         return $this->namespace;
     }
 
@@ -116,7 +122,23 @@ abstract class ScopedMappingModel extends MappingModel
      */
     public function isAbsoluteNamespace($namespace)
     {
-        return strpos($namespace, '\\') === 0;
+        return ($namespace && substr($namespace, 0, 1) === '\\');
+    }
+
+    /**
+     * Prepends a backslash to a namespace if there is none.
+     *
+     * A namespace with a backslash is considered absolute.
+     *
+     * @param string|null $namespace
+     *
+     * @return string|null
+     */
+    protected function makeNamespaceAbsolute(?string $namespace): ?string
+    {
+        $prependBackslash = ($namespace && !$this->isAbsoluteNamespace($namespace));
+
+        return ($prependBackslash) ?  "\\$namespace" : $namespace;
     }
 
     /**
--- a/tests/Propel/Tests/Resources/blog-database.xml
+++ b/tests/Propel/Tests/Resources/blog-database.xml
@@ -1,10 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
-<database name="acme_blog" defaultIdMethod="native" package="Acme" schema="acme" namespace="Acme\Model" baseClass="Acme\Model\ActiveRecord" defaultPhpNamingMethod="underscore" heavyIndexing="true" tablePrefix="acme_">
+<database name="acme_blog" defaultIdMethod="native" package="Acme" schema="acme" namespace="\Acme\Model" baseClass="\Acme\Model\ActiveRecord" defaultPhpNamingMethod="underscore" heavyIndexing="true" tablePrefix="acme_">
   <vendor type="mysql">
     <parameter name="Engine" value="InnoDB"/>
     <parameter name="Charset" value="utf8"/>
   </vendor>
-  <table name="blog_post" phpName="BlogPost" package="Acme.Blog" namespace="Acme\Model\Blog" baseClass="Acme\Model\ActiveRecord" description="The list of posts">
+  <table name="blog_post" phpName="BlogPost" package="Acme.Blog" namespace="\Acme\Model\Blog" baseClass="\Acme\Model\ActiveRecord" description="The list of posts">
     <column name="id" phpName="Id" type="integer" size="7" primaryKey="true" autoIncrement="true" required="true"/>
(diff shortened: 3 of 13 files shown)


tests/Propel/Tests/Resources/blog-database.xml
```diff
@@ -35,23 +35,23 @@
...
+  <table name="blog_author" phpName="BlogAuthor" package="Acme.Blog" namespace="\Acme\Model\Blog" baseClass="Acme\Model\ActiveRecord" description="The list of authors">
     <column name="id" phpName="Id" type="smallint" size="3" primaryKey="true" autoIncrement="true" required="true"/>
     <column name="username" phpName="Username" type="varchar" size="15" required="true"/>
     <column name="password" phpName="Password" type="varchar" size="40" required="true"/>
     <unique name="author_password_unique_idx">
       <unique-column name="username" size="8"/>
     </unique>
   </table>
-  <table name="blog_category" phpName="BlogCategory" package="Acme.Blog" namespace="Acme\Model\Blog" baseClass="Acme\Model\ActiveRecord" description="The list of categories">
+  <table name="blog_category" phpName="BlogCategory" package="Acme.Blog" namespace="\Acme\Model\Blog" baseClass="Acme\Model\ActiveRecord" description="The list of categories">
```
dereuromark, line 46: Do we need to adjust the presention here?
Couldnt we adjust the Interpretation instead to be fqcn?

mringler: I don't understand what you mean, could you please elaborate?

If you mean to always interpret the `namespace` attribute as fully qualified, no matter the trailing backslash, that would remove the relative namespace feature, which I think is quite nice.

dereuromark: But what about the baseLcass config?

    baseClass="Acme\Model\ActiveRecord"

vs 

    baseClass="\Acme\Model\ActiveRecord"
    
Would this not be inconsistent to keep that one then without leading slash?

mringler: Yes, that makes sense. Consistency is king. I changed it, so `baseClass` and `baseQueryClass` always start with a backslash.
When generating model files, the preceding slash will be removed (in `AbstractOmBuilder::declareClass()`, tested in `AbstractOmBuilderNamespaceTest::testGetDeclareClass()`, same as before.
Users can still use either format.

dereuromark: Are there any BC breaks to expect from changing the syntax on the XML files?
Would existing projects and declarations run into some troubles? Or is all compatible?

mringler: The changes should not affect existing schema.xml files or model classes generated from it.

The only actual change is when Propel writes a schema.xml file, which I think is only ever done during `database:reverse`. Then you get a different, more "understandable" output now (with all the absolute vs relative namespaces issue). And no more doubly namespaces. The resulting model classes do not differ.

mringler:
> tests are failing

Yes, since last Thursday, all test in the project dangle in queued state until they are automatically aborted after 24 hours. See comment  [here](https://github.com/propelorm/Propel2/pull/1730#issuecomment-834203629).
Has to be a configuration issue. Nothing I can do about it.
Can you please have a look at that?

src/Propel/Generator/Model/ScopedMappingModel.php
```diff
@@ -76,10 +76,16 @@ protected function setupObject()
     /**
      * Returns the namespace.
      *
-     * @return string
+     * @param bool $getAbsoluteNamespace
+     *
+     * @return string|null
      */
-    public function getNamespace()
+    public function getNamespace(bool $getAbsoluteNamespace = false)
```
dmytro-dymarchuk, line 83: Could you add a return type-hint?

src/Propel/Generator/Model/ScopedMappingModel.php
```diff
@@ -119,6 +125,22 @@ public function isAbsoluteNamespace($namespace)
...
+     *
+     * @param string|null $namespace
+     *
+     * @return string|null
+     */
+    public static function makeNamespaceAbsolute(?string $namespace): ?string
+    {
+        $prependBackslash = ($namespace && substr($namespace, 0, 1) !== '\\');
+
+        return ($prependBackslash) ? "\\$namespace" : $namespace;
```
dmytro-dymarchuk, line 141: Unnecessary parentheses or even variable.

mringler: The variable is there to keep the return statement readable and to self-comment on what is being checked.
The parentheses are there to highlight that this is an expression, not a simple assignment.

src/Propel/Generator/Model/ScopedMappingModel.php
mringler: Ok, added the type hint.

src/Propel/Generator/Model/ScopedMappingModel.php
dmytro-dymarchuk: I said about `($prependBackslash)`

mringler: Oh, those parentheses are there to make it easy to see that this is a ternary condition. And, same as above, I think it is helpful to emphasize that this is an expression.
I am well aware that I could cram the whole thing into a single line and omit the parens, but I genuinely consider that bad style. Granted, the kind of bad style that you can live with when working with other programmers.

src/Propel/Generator/Builder/Om/AbstractOMBuilder.php
```diff
@@ -361,21 +361,23 @@ public function declareClassNamespace($class, $namespace = '', $alias = false)
...
      *
      * @return bool
      */
-    protected function needAliasForClassName($class, $namespace)
+    protected function needAliasForClassName($class, $classNamespace)
     {
-        if ($namespace == $this->getNamespace()) {
+        $builderNamespace = trim($this->getNamespace(), '\\');
+
+        if ($classNamespace == $builderNamespace) {
```
gechetspr, line 372: not a blocker, but would be nice to have strict comparison where we can

src/Propel/Generator/Model/ScopedMappingModel.php
```diff
@@ -119,6 +125,22 @@ public function isAbsoluteNamespace($namespace)
...
+    /**
+     * Prepends a backslash to a namespace if there is none.
+     *
+     * A namespace with a backslash is considered absolute.
+     *
+     * @param string|null $namespace
+     *
+     * @return string|null
+     */
+    public static function makeNamespaceAbsolute(?string $namespace): ?string
```
gechetspr, line 137: Why this method is static? As I can see we use it non static methods, so I guess we can make it protected and nonstatic

src/Propel/Generator/Builder/Om/AbstractOMBuilder.php
mringler: This is how it was initially, my guess is that it is on purpose, so that empty strings and null match.

src/Propel/Generator/Model/ScopedMappingModel.php
mringler: Why would you make a method that does not access object properties nonstatic?

dereuromark: Wouldnt a normal method and `$this->makeNamespaceAbsolute()` call suffice, is I think, here the question.

mringler: Are you saying you want this to be an instance method, even when it does not work with the instance?
I am curious to know why, particularly if there is architectural concern behind it, or if you just like one better.
Either way, I am happy to change it, if it suits you better.

dereuromark: Instance and protected invocation are always preferred over public and static.
The question at the beginning was for a reason on the latter. Otherwise, yes, let's use protected normal.

src/Propel/Generator/Builder/Om/AbstractOMBuilder.php
dereuromark: According to docblock null is not a valid input. But trim() could maybe return false in some cases.
I guess we need to keep this as is for now as we are not aware of possible side effects.

src/Propel/Generator/Model/ScopedMappingModel.php
mringler: Wait, are we talking about visibility (public vs protected) or association (class vs instance)?
Nevermind, I changed it to protected and nonstatic.

asaulenko (approved):

PR #1732, opened 2021-05-15
Bugfixes for abstract tables

mringler:
I want to mark a table abstract, since it is only used for inheritance:
```xml
<table name="person" skipSql="true" abstract="true">...</table>
```
But the classes built from that schema have errors in them. Marking a table abstract seems to do exactly two things, both cause errors with PHP.

Current behavior 1: The user model class is set to abstract
Introduced problem: The query class is still set to instantiate objects from a now abstract class, which is not allowed
Proposed solution: If table is abstract, calling methods that usually create instances (`findPkSimple()`) throw an exception (unless when the table uses  concrete_inheritance, which creates instances from other classes).

Current behavior 2:  In the table map, set `OM_CLASS` to empty string and make the function `getOMClass()` abstract
Introduced problem: The table map class is not declared abstract, so it is not allowed to use abstract methods
Proposed solution: Instead of making it abstract, let the method throw an exception

On the way, I also did some housecleaning (mostly removal of unused variables).

### DIFF
--- a/src/Propel/Generator/Builder/Om/QueryBuilder.php
+++ b/src/Propel/Generator/Builder/Om/QueryBuilder.php
@@ -67,8 +67,16 @@ class QueryBuilder extends AbstractOMBuilder
     public function getParentClass()
     {
         $parentClass = $this->getBehaviorContent('parentClass');
+        if ($parentClass) {
+            return $parentClass;
+        }
 
-        return $parentClass === null ? ($this->getTable()->getBaseQueryClass() != '' ? $this->getTable()->getBaseQueryClass() : 'ModelCriteria') : $parentClass;
+        $baseQueryClass = $this->getTable()->getBaseQueryClass();
+        if ($baseQueryClass) {
+            return $baseQueryClass;
+        }
+
+        return 'ModelCriteria';
     }
 
     /**
@@ -627,10 +635,9 @@ abstract class " . $this->getUnqualifiedClassName() . ' extends ' . $parentClass
             return null;
         }";
         if ($table->hasCompositePrimaryKey()) {
-            $pks = [];
-            foreach ($table->getPrimaryKey() as $index => $column) {
-                $pks[] = "\$key[$index]";
-            }
+            $numberOfPks = count($table->getPrimaryKey());
+            $pkIndexes = range(0, $numberOfPks - 1);
+            $pks = preg_filter('/(\d+)/', '$key[${1}]', $pkIndexes); // put ids into "$key[]"
         } else {
             $pks = '$key';
         }
@@ -672,6 +679,19 @@ abstract class " . $this->getUnqualifiedClassName() . ' extends ' . $parentClass
 
         // this method is not needed if the table has no primary key
         if (!$table->hasPrimaryKey()) {
+            return;
+        }
+
+        $usesConcreteInheritance = $table->usesConcreteInheritance();
+        if ($table->isAbstract() && !$usesConcreteInheritance) {
+            $tableName = $table->getPhpName();
+            $script .= "
+    protected function findPkSimple(\$key, ConnectionInterface \$con)
+    {
+        throw new PropelException('$tableName is declared abstract, you cannot query it.');
+    }
+";
+
             return;
         }
 
@@ -741,7 +761,7 @@ abstract class " . $this->getUnqualifiedClassName() . ' extends ' . $parentClass
         \$obj = null;
         if (\$row = \$stmt->fetch(\PDO::FETCH_NUM)) {";
 
-        if ($table->getChildrenColumn()) {
+        if ($usesConcreteInheritance) {
             $script .= "
             \$cls = {$tableMapClassName}::getOMClass(\$row, 0, false);
             /** @var $ARClassName \$obj */
@@ -1224,7 +1244,6 @@ abstract class " . $this->getUnqualifiedClassName() . ' extends ' . $parentClass
      */
     protected function addFilterByArrayCol(&$script, Column $col)
     {
-        $colPhpName = $col->getPhpName();
         $singularPhpName = $col->getPhpSingularName();
         $colName = $col->getName();
         $variableName = $col->getCamelCaseName();
@@ -1306,7 +1325,6 @@ abstract class " . $this->getUnqualifiedClassName() . ' extends ' . $parentClass
             '\Propel\Runtime\Collection\ObjectCollection',
             '\Propel\Runtime\Exception\PropelException'
         );
-        $table = $this->getTable();
         $queryClass = $this->getQueryClassName();
         $fkTable = $fk->getForeignTable();
         $fkStubObjectBuilder = $this->getNewStubObjectBuilder($fkTable);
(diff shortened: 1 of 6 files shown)


tests/Propel/Tests/Generator/Builder/Om/QueryBuilderTest.php
```diff
@@ -1389,4 +1393,82 @@ public function testPruneCompositeKey()
...
+{
+    /**
+     * Build a TestableQueryBuilder for a table given in schema xml format.
+     *
+     * @param string $schemaXml
+     * @param string $tableName
+     *
+     * @return self
+     */
+    public static function forTableFromXml(string $schemaXml, string $tableName): self
```
dereuromark, line 1448: self and static (return) types should be removed. They should be documented only.

tests/Propel/Tests/Generator/Builder/Om/QueryBuilderTest.php
```diff
@@ -1389,4 +1390,44 @@ public function testPruneCompositeKey()
...
+
+    /**
+     * @return void
+     */
+    public function testFindPkSimpleIsAbstractWhenTableIsAbstract(): void
+    {
+        $databaseXml = '
+<database>
+    <table name="my_table" abstract="true">
+        <column name="id" type="integer" primaryKey="true" />
```
dereuromark, line 1402: Let's use the consistent version without trailing space.
```suggestion
        <column name="id" type="integer" primaryKey="true"/>
```

dereuromark (approved):

tests/Propel/Tests/Generator/Builder/Om/TestableQueryBuilder.php
```diff
@@ -0,0 +1,55 @@
...
+     * @param string $tableName
+     *
+     * @return self
+     */
+    public static function forTableFromXml(string $schemaXml, string $tableName): self
+    {
+        $reader = new SchemaReader();
+        $schema = $reader->parseString($schemaXml);
+        $table = $schema->getDatabase()->getTable($tableName);
+        $builder = new TestableQueryBuilder($table);
```
dereuromark, line 34: I am surprised this isnt found by cs sniffer as `self` usage requirement.

tests/Propel/Tests/Generator/Builder/Om/QueryBuilderTest.php
mringler: Oh, the code sniffer actually changed the return type from `TestableQueryBuilder` to `self`, and now does not complain anymore. Do you still want me to change it?

I moved the class into its own file. While doing so, I found that there already is a file that collects other classes used in the test file (`QueryBuilderTestClasses.php`). I think I have seen this pattern in other tests as well. The folder has 40 files now, which makes it all too easy to overlook something like this.
Note that autoloading classes that extend Bookstore will cause CI to fail during agnostic tests. I think it is because the fixtures are not generated. Those files have to be included manually. So autoloading is not a general solution, or needs at least needs an update of how CI works.

tests/Propel/Tests/Generator/Builder/Om/QueryBuilderTest.php
mringler: Ok

tests/Propel/Tests/Generator/Builder/Om/TestableQueryBuilder.php
mringler: I am not sure how many PHP developers know the difference between `self` and `static`, so the sniffer probably doesn't want them to accidentally introduce errors?

mringler:
I missed that the query model class stub has to be declared abstract as well, when the `findPkSimple()` method is declared abstract.
Makes sense too. If you cannot instantiate the table objects directly, neither can you instantiate the query directly.
The test case for this is pretty annoying, but what can you do.

src/Propel/Generator/Model/Table.php
```diff
@@ -801,6 +801,16 @@ public function getChildrenColumn()
...
     }
 
+    /**
+     * Checks whether the table uses concrete inheritance
+     *
+     * @return bool
+     */
+    public function usesConcreteInheritance(): bool
+    {
+        return ($this->inheritanceColumn !== null);
```
asaulenko, line 811: There's no need for brackets here.

dereuromark: I also dont think it hurts too much. A code sniffer some day could auto assert these things.

src/Propel/Generator/Builder/Om/QueryBuilder.php
```diff
@@ -67,8 +67,16 @@ public function getUnprefixedClassName()
     public function getParentClass()
     {
         $parentClass = $this->getBehaviorContent('parentClass');
+        if ($parentClass !== null) {
+            return $parentClass;
+        }
+
+        $baseQueryClass = $this->getTable()->getBaseQueryClass();
+        if ($baseQueryClass != '') {
```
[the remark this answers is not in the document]
mringler: I don't know, can we? This is how it was in the original statement.
I think $baseQueryClass can be null, too, in which case, no, we cannot make this comparison strict.
I can replace it with `if ($baseQueryClass) {`, if you are sure that nothing breaks?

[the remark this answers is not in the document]
dereuromark: All those `!= ''` are sure some code smells, as `if(...)` would definitely replace it with the same (falsely) meaning.

src/Propel/Generator/Model/Table.php
mringler: @asaulenko I am aware that PHP does not need brackets around expressions. Do you believe this has serious negative impact?

src/Propel/Generator/Builder/Om/QueryBuilder.php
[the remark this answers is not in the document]
mringler: Please check the updated version

src/Propel/Generator/Builder/Om/QueryBuilder.php
```diff
@@ -627,10 +635,9 @@ public function findPk(\$key, ConnectionInterface \$con = null)
             return null;
         }";
         if ($table->hasCompositePrimaryKey()) {
-            $pks = [];
-            foreach ($table->getPrimaryKey() as $index => $column) {
-                $pks[] = "\$key[$index]";
-            }
+            $numberOfPks = count($table->getPrimaryKey());
+            $pkIndexes = range(0, $numberOfPks - 1);
+            $pks = preg_filter('/(\d+)/', '$key[${1}]', $pkIndexes); // put ids into "$key[]"
```
gechetspr, line 640: Well, this is pretty cool :)
I've compared it with regular foreach loop and it's really faster

gechetspr (approved):

src/Propel/Generator/Builder/Om/QueryBuilder.php
mringler: Nice! I read on stackoverflow, that it is faster, but was too lazy to test (I changed it because I find it more readable).
And now I know it really is faster. Thanks!

PR #1734, opened 2021-05-17
Bugfix: Escape quotes in behavior 

mringler:
I am getting errors in the generated map- and model-files when using AggregateMultipleColumnsBehavior with single quotes in the filter condition. Problem is that strings are written to file without escaping those quotes.

In the model-file, it is a problem with the behavior, when the SQL query is not escaped.
In the map-file, it is a problem with the TableMapBuilder, which only escapes in arrays.

Also, `addGetBehaviors()` function in TableMapBuilder replaced all spaces in stringified arrays, probably to adjust the format of `var_export`. Because, what could go wrong...
Also a newline was missing at the beginning of the template added after the `getBehaviors()` function.

I did not change AggregateColumnsBehavior, even though it suffers from the same problem. Reason is that it would break the current workaround, which is to escape the single quotes in schema.xml, and I am not sure if you prefer the fix over backward compatibility. Let me know if I should change it, too.
AggregateMultipleColumnsBehavior is new, so BC is no concern.

### DIFF
--- a/tests/Propel/Tests/Generator/Builder/Om/TableMapBuilderTest.php
+++ b/tests/Propel/Tests/Generator/Builder/Om/TableMapBuilderTest.php
@@ -11,6 +11,7 @@ namespace Propel\Tests\Generator\Builder\Om;
 use Propel\Generator\Builder\Om\TableMapBuilder;
 use Propel\Generator\Builder\Util\SchemaReader;
 use Propel\Generator\Config\QuickGeneratorConfig;
+use Propel\Generator\Model\Table;
 use Propel\Runtime\Map\RelationMap;
 use Propel\Runtime\Propel;
 use Propel\Tests\Bookstore\Behavior\Map\Table1TableMap;
@@ -378,11 +379,12 @@ class TableMapBuilderTest extends BookstoreTestBase
         $schema = $reader->parseString($databaseXml);
         $table = $schema->getDatabase()->getTable('email');
 
-        $tableMapBuilder = new class($table) extends TableMapBuilder {
+        $tableMapBuilder = new class ($table) extends TableMapBuilder {
             public function getNormalizedColumnNameMapDefinition(): string
             {
                 $script = '';
                 $this->addNormalizedColumnNameMap($script);
+
                 return $script;
             }
         };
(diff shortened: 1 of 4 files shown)


dereuromark (approved):

tests/Propel/Tests/Generator/Builder/Om/TableMapBuilderTest.php
```diff
@@ -378,11 +379,12 @@ public function testNormalizedColumnMapHasOnlyUniqueKeys()
         $schema = $reader->parseString($databaseXml);
         $table = $schema->getDatabase()->getTable('email');
 
-        $tableMapBuilder = new class($table) extends TableMapBuilder {
+        $tableMapBuilder = new class ($table) extends TableMapBuilder {
```
asaulenko, line 382: PSR seems to not say anything about the style of an anonymous class declaration, hence I would follow the examples from the PHP manual. And those do not have spaces after `class`.

tests/Propel/Tests/Generator/Builder/Om/TableMapBuilderTest.php
```diff
@@ -399,4 +401,48 @@ public function getNormalizedColumnNameMapDefinition(): string
...
+    public function testStringify($input, string $message): void
+    {
+        $builder = new class (new Table('any')) extends TableMapBuilder{
+            public function doStringify($value): string
+            {
+                return $this->stringify($value);
+            }
+        };
+        $inputAsString = $builder->doStringify($input);
+        eval("\$output = $inputAsString;");
```
asaulenko, line 444: Not really sure about the purpose of this `eval` call. And shouldn't there be an escaping backslash before `$inputAsString`. Otherwise, it would be interpolated, I assume.

tests/Propel/Tests/Generator/Builder/Om/TableMapBuilderTest.php
dereuromark: There should be a code sniffer for for one or the other.

mringler: The code sniffer/fixer actually adds the space after class. I don't like it, but fighting the formatter is pointless.

tests/Propel/Tests/Generator/Builder/Om/TableMapBuilderTest.php
mringler: Have you looked at what the test case does? Would you prefer `'$output = ' .  $inputAsString . ';'`?
I prefer it how it is, since the concatenation hurts readability.
Yes, the mixing of stringified php code with actual php code is confusing. Unfortunately, that is the subject of the test case.
Instead of using `eval`, I could validate against stored strings. But I think that is a lesser solution, as it also tests for specific formatting, which might change over time, and really does not matter, as long as the snipped produces correct data.

dereuromark: I also think that a concat could suffice here. Since this is a test case, this is fine though IMO.

mringler: I have changed the variable names, hope that makes things clearer

gechetspr (approved):

PR #1735, opened 2021-05-17
Add test for integration BC.

dereuromark:
We would like to use an active demo project to verify that integration of the latest propel code as well as PRs work without unknown or hidden bc breaks.
So this check may fail, if BC breaks are known or expected.
Otherwise we will have to check in detail if a change is applicable into a minor or patch release or needs to be postponed until a major can be done again.

Feedback?

WIP
will require the current branch to be checked out using
```
composer require propel/propel:"dev-... as ..."
```
not sure how we can access that dynamically, or if we can just copy over the existing checked out propel code into the vendor dir?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -191,13 +191,13 @@ jobs:
               run: cd suite && composer install
 
             - name: Generate transfer objects
-              run: cd suite && vendor/bin/console transfer:generate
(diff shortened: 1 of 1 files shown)


dereuromark:
This is not to be approved :) It is WIP and contains, as mentioned above, further work for actually doing what it should be doing.

dereuromark:
Anyone any ideas on how we can now implement this integration testing on a large live project like this?
Or should we just close?

PR #1383, opened 2017-06-01
Data Dictionary Markdown Generator

crossan007:
Adds a generator for creating markdown data dictionaries.

usage: ```propel --config-dir=propel datadictionary:generate```

Proof of concept usage here: https://github.com/ChurchCRM/CRM/blob/5be16c7f1016ac117556dab623e5c9efdf7d2065/generated-datadictionary/default.schema.md

PR #1737, opened 2021-05-20
Feature/generate data dictionary documentation

mringler:
Stumbled over #1383 and think it is really nice. Most people will use a more powerful program to generate a data dictionary, but it is a nice feature to have.

Cleaned it up a bit and added a test case.
The test only checks that the command exits without error, I don't think it makes sense to validate the output. But I can add a check if you disagree.

Also, I wrapped the exception that is thrown when a command is run without `--config-dir` parameter, or when the config misses elements. The initial error message is only useful if you know where it was thrown.


mringler:
Example output (shortened):
# Data Dictionary for bookstore-behavior

<a name=\"TOC\"></a>
# Table of Contents
1. [add_class_table](#add_class_table)
2. [aggregate_comment](#aggregate_comment)

---

## Table **add_class_table** 
<a name="add_class_table"></a>

### Columns:
|Column Name|PHP Name|Type|Length|PK|AI|NN|U|FK|Default|Description|
|---|---|---|---|---|---|---|---|---|---|---|
|id|Id|INTEGER||*|*|*|||||
|title|Title|VARCHAR|100||||||||

[to table of contents](#TOC)

---

## Table **aggregate_comment** 
<a name="aggregate_comment"></a>

### Columns:
|Column Name|PHP Name|Type|Length|PK|AI|NN|U|FK|Default|Description|
|---|---|---|---|---|---|---|---|---|---|---|
|id|Id|INTEGER||*|*|*|||||
|post_id|PostId|INTEGER||||||[aggregate_post](#aggregate_post)|||

[to table of contents](#TOC)

src/Propel/Common/Config/ConfigurationManager.php
```diff
@@ -8,9 +8,11 @@
 
 namespace Propel\Common\Config;
 
+use Exception;
```
dereuromark, line 11: Where does this one come from?

mringler: Oh, initially, the catch was for all exceptions, and a new PropelException was thrown. But the tests relied on InvalidConfigurationException, so I changed it back. Forgot to remove the import - thank you for finding!

src/Propel/Generator/Manager/DataDictionaryExportManager.php
```diff
@@ -0,0 +1,262 @@
...
+    {
+        return ($val) ? '*' :  '';
+    }
+
+    /**
+     * @param string $name
+     *
+     * @return string
+     */
+    private function toAlphnum(string $name): string
```
dereuromark, line 242: maybe we can use real words here? also protected instead of private maybe.

dereuromark (commented):
Can we have a demo MD file content maybe added here into the PR as comment/descr?
Just so we can easier review that one?

src/Propel/Generator/Manager/DataDictionaryExportManager.php
mringler: Ok, changed it

src/Propel/Generator/Command/DataDictionaryExportCommand.php
```diff
@@ -0,0 +1,85 @@
...
+     * {@inheritDoc}
+     *
+     * @return void
+     */
+    protected function configure()
+    {
+        parent::configure();
+
+        $this
+            ->addOption('output-dir', null, InputOption::VALUE_REQUIRED, 'The output directory', self::DEFAULT_OUTPUT_DIRECTORY)
```
dmytro-dymarchuk, line 34: Would be better to move `output-dir` to const, since this value is used in several places.

dereuromark: You can also use `static::` here for class const access

mringler: Ok, extracted the option name into a const.

I don't think it matters here if I use `static::` or `self::`, as I can't think of a scenario where the value is overridden by a child class (which is the only difference between self and static, right?). I changed it anyway.

src/Propel/Generator/Manager/DataDictionaryExportManager.php
```diff
@@ -0,0 +1,263 @@
...
+use Propel\Generator\Model\Table;
+
+/**
+ * Manager for Markdown Data Dictionary
+ *
+ * @author Charles Crossan <crossan007@gmail.com>
+ */
+class DataDictionaryExportManager extends AbstractManager
+{
+    public const DICTIONARYCOLUMNS = ['Column Name', 'PHP Name', 'Type', 'Length', 'PK', 'NN', 'UQ', 'AI', 'FK', 'Default', 'Description'];
```
dereuromark, line 22: I wonder if those should be two words: DICTIONARY_COLUMNS

dereuromark (approved):
Overall I think this is fine.

There are some questions to whether this should be maybe an additional opt-in repo instead.
Or if other formats like json, XML formatting could or should also be possible.

But I think for now we can continue.

src/Propel/Generator/Manager/DataDictionaryExportManager.php
mringler: Good point, that wasn't named very well. Changed it to `COLUMN_TABLE_HEADERS`

src/Propel/Generator/Manager/DataDictionaryExportManager.php
```diff
@@ -0,0 +1,263 @@
...
+
+    /**
+     * @param \Propel\Generator\Model\Column $column
+     *
+     * @return string
+     */
+    protected function buildLinksToForeignTables(Column $column): string
+    {
+        $foreignKeys = $column->getForeignKeys();
+        if (count($foreignKeys) === 0) {
```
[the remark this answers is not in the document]
dereuromark: We usually just use if (!$...) in that case for simplicity even.

src/Propel/Generator/Manager/DataDictionaryExportManager.php
```diff
@@ -0,0 +1,263 @@
...
+
+            return $tables;
+    }
+
+    /**
+     * @param array $tables
+     *
+     * @return string
+     */
+    protected function buildToc(array $tables): string
```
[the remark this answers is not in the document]
dereuromark: This is an acronym.
I think for those the lines are a bit blury. The long version takes quite a bit of space, whereas the short form often already is quite descriptive IMO.

src/Propel/Generator/Command/DataDictionaryExportCommand.php
```diff
@@ -0,0 +1,85 @@
...
+        $manager->setGeneratorConfig($generatorConfig);
+        $schemaDir = $generatorConfig->getConfigProperty('paths.schemaDir');
+        if (!$schemaDir) {
+            throw new MissingInputException('Path to schema directory is missing. Use the --schema-dir option or the propel.paths.schemaDir configuration property to set it.');
+        }
+        $recursive = $generatorConfig->getConfigProperty('generator.recursive');
+        $schemas = $this->getSchemas($schemaDir, $recursive);
+        $manager->setSchemas($schemas);
+        $manager->setLoggerClosure(function ($message) use ($input, $output) {
+            if ($input->getOption('verbose')) {
```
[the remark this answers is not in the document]
mringler: I don't know, but it is used in all commands the same way. The initial author probably just copied it.

src/Propel/Generator/Manager/DataDictionaryExportManager.php
```diff
@@ -0,0 +1,263 @@
...
+        return preg_replace('/\W/', '-', $name);
+    }
+
+    /**
+     * @param string $title
+     * @param string|null $anchor
+     *
+     * @return string
+     */
+    protected function buildLink(string $title, ?string $anchor = null): string
```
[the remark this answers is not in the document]
mringler: Good point, I missed removing this. Thanks

src/Propel/Generator/Manager/DataDictionaryExportManager.php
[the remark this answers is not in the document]
mringler: changed it to ` if (!$...)`

src/Propel/Generator/Manager/DataDictionaryExportManager.php
```diff
@@ -0,0 +1,263 @@
...
+
+        $tableItems = [];
+        foreach ($tables as $table) {
+            $tableName = $table->getName();
+            $anchorName = $this->buildAnchorName($tableName);
+            $tableItems[] = "$tableCount. [$tableName](#$anchorName)";
+            $tableCount++;
+        }
+
+        return implode("\n", $tableItems);
```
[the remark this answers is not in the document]
mringler: Good point

src/Propel/Generator/Manager/DataDictionaryExportManager.php
```diff
@@ -0,0 +1,263 @@
...
+$tocMd
+
+$tableSectionMd
+EOT;
+    }
+
+    /**
+     * @param \Propel\Generator\Model\Database $database
+     *
+     * @return array
```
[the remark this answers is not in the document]
mringler: ok

src/Propel/Generator/Manager/DataDictionaryExportManager.php
```diff
@@ -0,0 +1,263 @@
...
+     * @return string
+     */
+    protected function buildLink(string $title, ?string $anchor = null): string
+    {
+        if ($anchor === null) {
+            $anchor = $title;
+        }
+        $anchorName = $this->buildAnchorName($anchor);
+
+        return "[$title](#$anchorName)";
```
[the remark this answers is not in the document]
mringler: fine with me

src/Propel/Generator/Command/DataDictionaryExportCommand.php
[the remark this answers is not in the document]
dmytro-dymarchuk: In this case, I think we should add option definition to `AbstractCommand::configure()`

src/Propel/Generator/Command/DataDictionaryExportCommand.php
```diff
@@ -0,0 +1,88 @@
...
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * @author Charles Crossan <crossan007@gmail.com>
+ */
+class DataDictionaryExportCommand extends AbstractCommand
+{
+    public const DEFAULT_OUTPUT_DIRECTORY = 'generated-datadictionary';
+
+    private const OPTION_OUTPUT_DIR = 'output-dir';
```
dereuromark, line 24: Can we make those protected both?
Last thing from my side.

src/Propel/Generator/Command/DataDictionaryExportCommand.php
```diff
@@ -0,0 +1,88 @@
...
+     * {@inheritDoc}
+     *
+     * @return void
+     */
+    protected function configure()
+    {
+        parent::configure();
+
+        $this
+            ->addOption(self::OPTION_OUTPUT_DIR, null, InputOption::VALUE_REQUIRED, 'The output directory', static::DEFAULT_OUTPUT_DIRECTORY)
```
dereuromark, line 37: Those would then also be static::

src/Propel/Generator/Command/DataDictionaryExportCommand.php
mringler: ok

src/Propel/Generator/Command/DataDictionaryExportCommand.php
mringler: ok

crossan007:
@mringler Wow!  The timing on this is impeccable   I had lost track of my original merge request (totally forgot about it), but today I was looking for something that did this, and I found you've been working on it!  Thank you!

Issue #1695, opened 2021-01-11
Unable to use php74 preloading because of how table maps are dumped


prgTW:
![image](https://user-images.githubusercontent.com/203249/104244628-1d008180-5463-11eb-9a9a-b5f2714bc1a2.png)

prgTW:
references #361

mringler:
I have created a PR (#1742) which should fix this issue. If you have time to test it, I'd be happy to get some feedback

emicks:
I am using propelorm/Propel2 together with SkyFoxvn/PropelBundle in a Symfony 4.4 application. PR #1742 breaks the application even if the model has been rebuilt, because the database connections are configured via PropelBundle (config coming from config/packages/propel.yaml in the symfony application) and the generated loader script is not included. Is there any workaround for this?

mringler:
Interesting, thank your for reporting.
What error message are you seeing?

emicks:
I'm getting a "Database map was not initialized. Please check the database loader script included by your conf" error. After  (manually) including the generated loadDatabase.php in PropelBundle::configureConnections(), it seems to work. As far as I understood PR #1742, the generated loader is included at the end of the (PHP) config script. However, with PropelBundle, the config is loaded from YAML and connections are  initialized in PropelBundle::boot().

mringler:
Great that you figured it out and it is working. I'd still like to find out what went wrong, so we can fix it.
So, initially, was there no `require_once` at the end of conf.php? Or did it point to another file?
At the top of `conf.php` (line 3), does it check for version 2 (instead of "propel2-dev"):
```
$serviceContainer->checkVersion(2);
```

emicks:
AFAIK, PropelBundle does not use a `conf.php`. With PropelBundle, configuration is loaded from a .yaml file according to the bundle's config info which does not recognize `loaderScriptDir` : https://github.com/SkyFoxvn/PropelBundle/blob/5.0/DependencyInjection/Configuration.php
The connections are then initialized in the boot code of the bundle. If I `require_once` the generated loader in PropelBundle::configureConnections() (line 59 in  https://github.com/SkyFoxvn/PropelBundle/blob/5.0/PropelBundle.php )  it works.  
Perhaps your initial assumption that "_the configuration is the only file that is imported directly into a Propel project_" could be true only for standalone projects, but not for Symfony projects with PropelBundle.

mringler:
jinx

mringler:
[...]

Issue #361, opened 2013-04-21
[RFC] remove static initialization of TableMap


jaugustin:
We need to find a good way to initialize the `TableMap`classes and remove the static call 
`TableMap::buildTableMap()` at the bottom of each `TableMap`classes

Maybe we could use :
- composer class mapping (to auto discover classes)
- use the propel container
- ...

dereuromark:
@mringler Can we close this with latest changes done?

mringler:
Yes, we handled this. Let's close the issue.

PR #1742, opened 2021-05-27
Add table map initializer

mringler:
This PR fixes the issue with missing table registration when using PHP 7.4 preloading (see #1695).

This is a big one, and I am not sure how much I like it. But there is probably no way around it.

### tl;dr
- Instead of relying on automatic registration of tables, which does not work with preloading, the tables are now registered explicitly through a script.
- Script creation and importing is done in a way where it does not impact existing projects.
- However, users have to rebuild their configuration and their models with the `config:convert` and `model:build` command.
- If they are not aware of that, exceptions will tell them what to do.

### Background
The metadata for tables is stored in table map files, which are generated by the `model:build` command (i.e. BookTableMap.php). They are registered in database maps in the global ServiceContainer. Every reference to a table in Propel is resolved through these files. Without the table map files, Propel knows nothing about the table and will not be able to work with it.

Currently, the table map files register themselves when being loaded. This is done through a static call, which is inserted after the class declaration in each table map file:

```php
class BookTableMap extends TableMap
{
    ...

     /**
     * Add a TableMap instance to the database for this tableMap class.
     */
    public static function buildTableMap()
    {
        $dbMap = Propel::getServiceContainer()->getDatabaseMap(BookTableMap::DATABASE_NAME);
        if (!$dbMap->hasTable(BookTableMap::TABLE_NAME)) {
            $dbMap->addTableObject(new BookTableMap());
        }
    }
} // BookTableMap
// This is the static code needed to register the TableMap for this table with the main Propel class.
//
BookTableMap::buildTableMap()
```
 This approach is often used to bridge the issue of missing static class initializers in PHP. It is pretty nice, because no additional structure is needed - the classes just make themselves known to the ServiceContainer.

### The problem
 When using PHP 7.4 preloading, the static call is executed during preloading, but the result is removed when preloading finishes. The static calls are not run again when restoring from cache (see for example [here](https://stackoverflow.com/a/61577684/4883195) or #1695), leading to an empty database schema. This means that it is currently impossible to preload the table map files.

### Proposed solution
So, there is an easy part and a hard part to fixing this.

The easy part is to extract the static table map initialization calls into a standalone script, which can be run manually during Propel initialization. This can easily be done during the 'model:build' command.

The hard part is figuring out how to run the script without having users to adjust their code base. Because while it would be easy to have them import the file manually, like the configuration, it means that they have to be aware of the change and know what to do, which will lead to endless bug reports.

I think the least annoying approach is to add an import statement at the end of the configuration. It is the only file directly imported into a Propel project, making it the only candidate for piggybacking. It is also auto-generated, so we can change it in a structured manner.
Very nice, however, creating the configuration and creating the table loader script is done in different commands - `config:convert`  and `model:build`. So after building the configuration, it contains an import statement for a file, which is yet to be created. And  `model:build` needs to know where the configuration expects the table loader script, but doesn't even know where the configuration file is. Also, the process has to work without users necessarily being aware of it.

Long story short, this is what I came up with:
- running `config:convert` creates an updated configuration which imports the table loader script, and creates a dummy file at the expected location
- running `model:build` will create the actual table loader script at the expected location
- The expected location is either shared through the `paths` property of the raw configuration (like `propel.json`), or by passing a value via parameter to both commands:
    - either the parameter `--loader-script-dir` has to be passed to both commands for manual setup
    - setting the new `paths.loaderScriptDir` property in the raw configuration allows to set a custom location
    - if none of the above is given, the script will be put at the location of `paths.phpConfigDir` (which default to `./generated-conf/`)
 - if a user updates Propel without rebuilding the configuration, an exception is thrown upon reading it, telling them to run `config:convert`
 - if a user has updated the configuration, but has not rebuild the models, the service container will throw an exception when accessing the database map, telling him to check the script, and if it is the dummy script, it will tell them to run `model:build`

So this changes some substantial mechanisms in Propel, and I a bit anxious about that. Probably why the topic was gracefully dropped in #361. Then again, I don't think there is a way around it. I have looked for easier ways, like fiddling with the composer autoload mechanism, or just searching for all table map classes during initialization, but those introduce performance issues among other problems.

Of course, I am always happy to discuss.

### DIFF
--- a/src/Propel/Generator/Builder/Om/TableMapBuilder.php
+++ b/src/Propel/Generator/Builder/Om/TableMapBuilder.php
@@ -173,8 +173,6 @@ class " . $this->getUnqualifiedClassName() . " extends TableMap
             $this->addGetTableMap($script);
         }
 
-        $this->addBuildTableMap($script);
-
         $this->addDoDelete($script);
         $this->addDoDeleteAll($script);
 
@@ -496,37 +494,11 @@ class " . $this->getUnqualifiedClassName() . " extends TableMap
      */
     protected function addClassClose(&$script)
     {
+        $unqualifiedClassName = $this->getUnqualifiedClassName();
         $script .= "
-} // " . $this->getUnqualifiedClassName() . "
-// This is the static code needed to register the TableMap for this table with the main Propel class.
-//
-" . $this->getUnqualifiedClassName() . "::buildTableMap();
+} // $unqualifiedClassName
 ";
         $this->applyBehaviorModifier('tableMapFilter', $script, '');
-    }
-
-    /**
-     * Adds the buildTableMap() method.
-     *
-     * @param string $script The script will be modified in this method.
-     *
-     * @return void
-     */
-    protected function addBuildTableMap(&$script)
-    {
-        $this->declareClassFromBuilder($this->getTableMapBuilder());
-        $script .= "
-    /**
-     * Add a TableMap instance to the database for this tableMap class.
-     */
-    public static function buildTableMap()
-    {
-        \$dbMap = Propel::getServiceContainer()->getDatabaseMap(" . $this->getTableMapClass() . "::DATABASE_NAME);
-        if (!\$dbMap->hasTable(" . $this->getTableMapClass() . "::TABLE_NAME)) {
-            \$dbMap->addTableObject(new " . $this->getTableMapClass() . "());
-        }
-    }
-";
     }
 
     /**
--- a/src/Propel/Generator/Builder/Om/TableMapLoaderScriptBuilder.php
+++ b/src/Propel/Generator/Builder/Om/TableMapLoaderScriptBuilder.php
@@ -1,0 +1,113 @@
+<?php
+
+/**
+ * MIT License. This file is part of the Propel package.
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Propel\Generator\Builder\Om;
+
+use Propel\Generator\Builder\Util\PropelTemplate;
+use Propel\Generator\Config\GeneratorConfigInterface;
+use Propel\Generator\Model\Table;
+use SplFileInfo;
+
+/**
+ * Generates a database loader file, which is used to register all table maps with the DatabaseMap.
+ */
+class TableMapLoaderScriptBuilder
+{
+    public const FILENAME = 'loadDatabase.php';
+
+    /**
+     * @var \Propel\Generator\Config\GeneratorConfigInterface $generatorConfig
+     */
+    protected $generatorConfig;
+
+    /**
+     * @param \Propel\Generator\Config\GeneratorConfigInterface $generatorConfig
+     */
+    public function __construct(GeneratorConfigInterface $generatorConfig)
+    {
+        $this->generatorConfig = $generatorConfig;
+    }
+
+    /**
+     * @param array $schemas
+     *
+     * @return string
+     */
+    public function build(array $schemas): string
+    {
+        $vars = $this->buildVars($schemas);
+
+        return $this->renderTemplate($vars);
+    }
+
+    /**
+     * @param \Propel\Generator\Model\Schema[] $schemas
+     *
+     * @return array
+     */
+    protected function buildVars(array $schemas): array
+    {
+        $databaseNameToTableMapNames = [];
+
+        foreach ($schemas as $schema) {
+            foreach ($schema->getDatabases(false) as $database) {
+                $databaseName = $database->getName();
+                $tableMapNames = array_map([$this, 'getFullyQualifiedTableMapClassName'], $database->getTables());
+                if (array_key_exists($databaseName, $databaseNameToTableMapNames)) {
+                    $existing = $databaseNameToTableMapNames[$databaseName];
+                    $tableMapNames = array_merge($existing, $tableMapNames);
+                }
+                sort($tableMapNames);
+                $databaseNameToTableMapNames[$databaseName] = $tableMapNames;
+            }
+        }
+
+        return [
+            'databaseNameToTableMapNames' => $databaseNameToTableMapNames,
+        ];
+    }
+
+    /**
+     * @param \Propel\Generator\Model\Table $table
+     *
+     * @return string
+     */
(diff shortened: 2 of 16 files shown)


mringler:
Alright, test look better on my machine... This will take some time to fix

src/Propel/Generator/Manager/ModelManager.php
```diff
@@ -194,4 +196,33 @@ protected function doBuild(AbstractOMBuilder $builder, $overwrite = true)
...
+        $file = $builder->getFile();
+        $filePath = $file->getPathname();
+
+        $action = $file->isFile() ? 'Updating' : 'Creating';
+        $this->log("\t-> $action $filePath");
+
+        if ($file->isFile() && file_get_contents($filePath) === $fileContent) {
+            $this->log("\t\t(no change)");
+
+            return 0;
```
dereuromark, line 220: What does 0 mean - is that a console return code? maybe we can (re)use the class const CODE_SUCCESS / CODE_ERROR for those as more speaking ones instead of magic numbers?

src/Propel/Generator/Manager/ModelManager.php
```diff
@@ -194,4 +196,33 @@ protected function doBuild(AbstractOMBuilder $builder, $overwrite = true)
...
+        }
+        $this->log('Generating script for loading table maps at ' . $file->getRealPath());
+        $this->filesystem->mkdir($file->getPath());
+        file_put_contents($filePath, $fileContent);
+
+        return 1;
```
dereuromark, line 226: Is that an error? Otherwise I would recommend using a non 1 code, e.g 2 - ideally as CODE_CHANGES class const somewhere.

src/Propel/Generator/Util/QuickBuilder.php
```diff
@@ -426,6 +427,10 @@ public function buildClasses(?array $classTargets = null): void
         foreach ($includes as $tempFile) {
             include($tempFile);
         }
+
+        if (in_array('tablemap', $classes)) {
```
dereuromark, line 431: we should set 3rd arg to true here for stringish compares.

src/Propel/Generator/Builder/Om/TableMapBuilder.php
```diff
@@ -173,8 +173,6 @@ protected function addClassBody(&$script)
             $this->addGetTableMap($script);
         }
 
-        $this->addBuildTableMap($script);
```
dereuromark, line 176: Can we clarify the BC breaking impact of removing this part?
Also looking into projects that extend the table map builder etc.

src/Propel/Generator/Builder/Om/TableMapLoaderScriptBuilder.php
```diff
@@ -0,0 +1,113 @@
...
+ * Generates a database loader file, which is used to register all table maps with the DatabaseMap.
+ */
+class TableMapLoaderScriptBuilder
+{
+    public const FILENAME = 'loadDatabase.php';
+
+    /**
+     * @var \Propel\Generator\Config\GeneratorConfigInterface $generatorConfig
+     */
+    private $generatorConfig;
```
dereuromark, line 26:
```suggestion
    protected $generatorConfig;
```

src/Propel/Generator/Manager/ModelManager.php
mringler: That's the number of changed files. It is collected by the calling method for user output. I'll add a description to the return parameter.

src/Propel/Generator/Manager/ModelManager.php
mringler: see above

src/Propel/Generator/Util/QuickBuilder.php
mringler: ok

src/Propel/Generator/Builder/Om/TableMapLoaderScriptBuilder.php
mringler: changed it

src/Propel/Generator/Builder/Om/TableMapBuilder.php
mringler: That whole thing is pretty much internal Propel business. There is no designated way to override table map classes, and the way they are tightly coupled with the base model- and query-classes should make it incredibly difficult to do so.
Also, changing this method would have broken the initialization, unless the whole mechanism was rewritten to an extend, where that Propel instance cannot be updated anymore.
I am pretty sure we are safe here.

tests/Propel/Tests/Runtime/ServiceContainer/StandardServiceContainerTest.php
```diff
@@ -190,31 +192,106 @@ public function testSetAdaptersRemovesExistingAdaptersForAllDatasources($value =
...
      */
-    public function testCheckValidVersion()
+    public function testUninitializedDatabaseMapThrowsException(): void
     {
-        $logger = $this->getMockBuilder('Monolog\Logger')
-            ->setMethods(['warning'])
-            ->setConstructorArgs(['mylogger'])
-            ->getMock();
-        $logger->expects($this->never())->method('warning');
+        $sc = new StandardServiceContainer();
```
gechetspr, line 220: same, but for tests it's ok

gechetspr (approved):
Few things with variable names, but the rest looks good

src/Propel/Generator/Builder/Om/TableMapBuilder.php
```diff
@@ -496,39 +494,13 @@ protected function addNormalizedColumnNameMap(&$script): void
      */
     protected function addClassClose(&$script)
     {
+        $uqClassName = $this->getUnqualifiedClassName();
```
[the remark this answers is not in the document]
mringler: I agree and changed it

tests/Propel/Tests/Runtime/ServiceContainer/StandardServiceContainerTest.php
mringler: I agree, but `sc` is used throughout the whole class, so for the sake of consistency, I think `$sc` works best here.
If you insist, I can easily replace all occurrences. Though the PR is quite large as it is.

PR #1746, opened 2021-05-31
Tests: remove unused fixture

mringler:
While working on #1742, I stumbled over an unused fixture ('bookstore-packaged'), so I removed it. Two tests (which I wrote) used schema.xml files, but they can easily use others.

Also, I ran into `TestCaseFixtures`, which was pretty bad (look for the `$skip` variable if you want to see something funny), and I was in the mood for refactoring.


mringler:
Hmm, first run of tests failed, because I used `Command::SUCCESS`, which apparently is only available on symfony "5-max" version. But the first run is stuck now, some test suits are not starting (see https://github.com/propelorm/Propel2/actions/runs/893667080). Don't know why, no changes to the code should be able to compromise the test runner.
Unless someone stops the last run, it will be aborted in 24 hours, at which point the updated test are started, which should go through smoothly.

PR #1747, opened 2021-06-04
Feature/untangle criteria class

mringler:
I was in the mood for some refactoring, and decided it is time to go for the white whale, aka the `Criteria` classes.

I believe those are Propel's achilles heel, since they are the center classes of ActiveQuery and parent class to all Query classes, and completely unmaintainable - with together almost 6000 lines of code, a fuzzy extension hierarchy, and them doing literally dozens of things at once.
In practice, that is acceptable, as long as everything works perfectly. But almost all bugs I have fixed were in those three classes, and I am pretty sure there are more. Also, I am pretty sure that a lot of performance is lost by hacky workarounds, of which I have seen plenty. I think those bugs and workarounds show that the information flow in those classes is not understandable anymore. 

 So here is a proposal to start getting things back in order by removing sql query generation and execution out of those classes. This already removes more than 500 lines of code.

The basic idea is to have the `doSelect()`, `doInsert()`, `do...()` methods delegate to dedicated classes. Here is an overview of the new classes (not sure if the `tree` output renders in all browsers):

```
ActiveQuery
├── BaseModelCriteria.php
├── Criteria.php
├── ModelCriteria.php
├── ...
├── QueryExecutor
│   ├── AbstractQueryExecutor.php
│   ├── CountQueryExecutor.php
│   ├── DeleteAllQueryExecutor.php
│   ├── DeleteQueryExecutor.php
│   ├── InsertQueryExecutor.php
│   ├── QueryExecutionException.php
│   ├── SelectQueryExecutor.php
│   └── UpdateQueryExecutor.php
└── SqlBuilder
    ├── AbstractSqlQueryBuilder.php
    ├── CountQuerySqlBuilder.php
    ├── DeleteQuerySqlBuilder.php
    ├── InsertQuerySqlBuilder.php
    ├── SelectQuerySqlBuilder.php
    └── UpdateQuerySqlBuilder.php

``` 
The changes are very thoroughly tested through the Criteria classes and a few new tests.

### DIFF
--- a/src/Propel/Runtime/ActiveQuery/Criteria.php
+++ b/src/Propel/Runtime/ActiveQuery/Criteria.php
@@ -2046,9 +2046,10 @@ class Criteria
      */
     public function createSelectSql(&$params)
     {
-        [$sql, $params] = SelectQuerySqlBuilder::createSelectSql($this, $params);
+        $preparedStatementDto = SelectQuerySqlBuilder::createSelectSql($this, $params);
+        $params = $preparedStatementDto->getParameters();
 
-        return $sql;
+        return $preparedStatementDto->getSqlStatement();
     }
 
     /**
--- a/src/Propel/Runtime/ActiveQuery/Criteria.php
+++ b/src/Propel/Runtime/ActiveQuery/Criteria.php
@@ -16,11 +16,15 @@ use Propel\Runtime\ActiveQuery\Criterion\CustomCriterion;
 use Propel\Runtime\ActiveQuery\Criterion\InCriterion;
 use Propel\Runtime\ActiveQuery\Criterion\LikeCriterion;
 use Propel\Runtime\ActiveQuery\Criterion\RawCriterion;
+use Propel\Runtime\ActiveQuery\QueryExecutor\CountQueryExecutor;
+use Propel\Runtime\ActiveQuery\QueryExecutor\DeleteAllQueryExecutor;
+use Propel\Runtime\ActiveQuery\QueryExecutor\DeleteQueryExecutor;
+use Propel\Runtime\ActiveQuery\QueryExecutor\InsertQueryExecutor;
+use Propel\Runtime\ActiveQuery\QueryExecutor\SelectQueryExecutor;
+use Propel\Runtime\ActiveQuery\QueryExecutor\UpdateQueryExecutor;
+use Propel\Runtime\ActiveQuery\SqlBuilder\SelectQuerySqlBuilder;
 use Propel\Runtime\Connection\ConnectionInterface;
-use Propel\Runtime\Connection\ConnectionWrapper;
-use Propel\Runtime\Connection\StatementWrapper;
 use Propel\Runtime\Exception\LogicException;
-use Propel\Runtime\Exception\PropelException;
 use Propel\Runtime\Propel;
 use Propel\Runtime\Util\PropelConditionalProxy;
 
@@ -2042,205 +2046,7 @@ class Criteria
      */
     public function createSelectSql(&$params)
     {
-        $adapter = Propel::getServiceContainer()->getAdapter($this->getDbName());
-        $dbMap = Propel::getServiceContainer()->getDatabaseMap($this->getDbName());
-
-        $fromClause = [];
-        $joinClause = [];
-        $joinTables = [];
-        $whereClause = [];
-        $orderByClause = [];
-
-        $orderBy = $this->getOrderByColumns();
-
-        // get the first part of the SQL statement, the SELECT part
-        $selectSql = $adapter->createSelectSqlPart($this, $fromClause);
-        $this->replaceNames($selectSql);
-
-        // Handle joins
-        // joins with a null join type will be added to the FROM clause and the condition added to the WHERE clause.
-        // joins of a specified type: the LEFT side will be added to the fromClause and the RIGHT to the joinClause
-        foreach ($this->getJoins() as $join) {
-            $join->setAdapter($adapter);
-
-            // add 'em to the queues..
-            if (!$fromClause) {
-                $fromClause[] = $join->getLeftTableWithAlias();
-            }
-            $joinTables[] = $join->getRightTableWithAlias();
-            $joinClauseString = $join->getClause($params);
-            $this->replaceNames($joinClauseString);
-            $joinClause[] = $joinClauseString;
-        }
-
-        // add the criteria to WHERE clause
-        // this will also add the table names to the FROM clause if they are not already
-        // included via a LEFT JOIN
-        foreach ($this->keys() as $key) {
-            $criterion = $this->getCriterion($key);
-            $table = null;
-            foreach ($criterion->getAttachedCriterion() as $attachedCriterion) {
-                $tableName = $attachedCriterion->getTable();
-
-                $table = $this->getTableForAlias($tableName);
-                if ($table !== null) {
-                    $fromClause[] = $table . ' ' . $tableName;
-                } else {
-                    $fromClause[] = $tableName;
-                    $table = $tableName;
-                }
-
-                if (
-                    $this->isIgnoreCase() && method_exists($attachedCriterion, 'setIgnoreCase')
-                    && $dbMap->getTable($table)->getColumn($attachedCriterion->getColumn())->isText()
-                ) {
-                    $attachedCriterion->setIgnoreCase(true);
-                }
-            }
-
-            $criterion->setAdapter($adapter);
-
-            $sb = '';
-            $criterion->appendPsTo($sb, $params);
-            $this->replaceNames($sb);
-            $whereClause[] = $sb;
-        }
-
-        // Unique from clause elements
-        $fromClause = array_unique($fromClause);
-        $fromClause = array_diff($fromClause, ['']);
-
-        // tables should not exist in both the from and join clauses
-        if ($joinTables && $fromClause) {
-            foreach ($fromClause as $fi => $ftable) {
-                if (in_array($ftable, $joinTables)) {
-                    unset($fromClause[$fi]);
-                }
-            }
-        }
-
-        $having = $this->getHaving();
-        $havingString = null;
-        if ($having !== null) {
-            $sb = '';
-            $having->appendPsTo($sb, $params);
-            $this->replaceNames($sb);
-            $havingString = $sb;
-        }
-
-        if (!empty($orderBy)) {
-            foreach ($orderBy as $orderByColumn) {
-                // Add function expression as-is.
-                if (strpos($orderByColumn, '(') !== false) {
-                    $orderByClause[] = $orderByColumn;
-
-                    continue;
-                }
-
-                // Split orderByColumn (i.e. "table.column DESC")
-                $dotPos = strrpos($orderByColumn, '.');
-
-                if ($dotPos !== false) {
-                    $tableName = substr($orderByColumn, 0, $dotPos);
-                    $columnName = substr($orderByColumn, $dotPos + 1);
-                } else {
-                    $tableName = '';
-                    $columnName = $orderByColumn;
-                }
-
-                $spacePos = strpos($columnName, ' ');
-
-                if ($spacePos !== false) {
-                    $direction = substr($columnName, $spacePos);
-                    $columnName = substr($columnName, 0, $spacePos);
-                } else {
-                    $direction = '';
-                }
-
-                $tableAlias = $tableName;
-                $aliasTableName = $this->getTableForAlias($tableName);
-                if ($aliasTableName) {
-                    $tableName = $aliasTableName;
-                }
-
-                $columnAlias = $columnName;
-                $asColumnName = $this->getColumnForAs($columnName);
-                if ($asColumnName) {
-                    $columnName = $asColumnName;
-                }
-
-                $column = $tableName ? $dbMap->getTable($tableName)->getColumn($columnName) : null;
-
-                if ($this->isIgnoreCase() && $column && $column->isText()) {
-                    $ignoreCaseColumn = $adapter->ignoreCaseInOrderBy("$tableAlias.$columnAlias");
-                    $this->replaceNames($ignoreCaseColumn);
-                    $orderByClause[] = $ignoreCaseColumn . $direction;
-                    $selectSql .= ', ' . $ignoreCaseColumn;
-                } else {
-                    $this->replaceNames($orderByColumn);
-                    $orderByClause[] = $orderByColumn;
-                }
-            }
-        }
-
-        // tables should not exist as alias of subQuery
-        if ($this->hasSelectQueries()) {
-            foreach ($fromClause as $index => $ftable) {
-                if (strpos($ftable, ' ') !== false) {
-                    [, $tableName] = explode(' ', $ftable);
-                } else {
-                    $tableName = $ftable;
-                }
-                if ($this->hasSelectQuery($tableName)) {
-                    unset($fromClause[$index]);
-                }
-            }
-        }
-
-        // from / join tables quoted if it is necessary
-        $fromClause = array_map([$this, 'quoteIdentifierTable'], $fromClause);
-        $joinClause = $joinClause ? $joinClause : array_map([$this, 'quoteIdentifierTable'], $joinClause);
-
-        // add subQuery to From after adding quotes
-        foreach ($this->getSelectQueries() as $subQueryAlias => $subQueryCriteria) {
-            $fromClause[] = '(' . $subQueryCriteria->createSelectSql($params) . ') AS ' . $subQueryAlias;
-        }
-
-        if (empty($fromClause) && $this->getPrimaryTableName()) {
-            $fromClause[] = $this->getPrimaryTableName();
-        }
-
-        // build from-clause
-        $from = '';
-        if (!empty($joinClause) && count($fromClause) > 1) {
-            $from .= implode(' CROSS JOIN ', $fromClause);
-        } else {
-            $from .= implode(', ', $fromClause);
-        }
-
-        $from .= $joinClause ? ' ' . implode(' ', $joinClause) : '';
-
-        // Build the SQL from the arrays we compiled
-        $sql = $selectSql
-            . ' FROM ' . $from
-            . ($whereClause ? ' WHERE ' . implode(' AND ', $whereClause) : '');
-
-        $groupBy = $adapter->getGroupBy($this);
-        if ($groupBy) {
-            $this->replaceNames($groupBy);
-            $sql .= $groupBy;
-        }
-
-        $sql .= ($havingString ? ' HAVING ' . $havingString : '')
-             . ($orderByClause ? ' ORDER BY ' . implode(',', $orderByClause) : '');
-
-        if ($this->getLimit() >= 0 || $this->getOffset()) {
-            $adapter->applyLimit($sql, $this->getOffset(), $this->getLimit(), $this);
-        }
-
-        if (null !== $this->lock) {
-            $adapter->applyLock($sql, $this->lock);
-        }
+        [$sql, $params] = SelectQuerySqlBuilder::createSelectSql($this, $params);
 
         return $sql;
     }
@@ -2289,38 +2095,6 @@ class Criteria
 
                     return $adapter->quote($string);
                 }
-            }
-        }
-
-        return $string;
-    }
-
-    /**
-     * @param string $string
-     *
-     * @return string
-     */
-    public function quoteIdentifierTable($string)
-    {
-        $realTableName = $string;
-        if (($pos = strrpos($string, ' ')) !== false) {
-            $realTableName = substr($string, 0, $pos);
-        }
-
-        $dbMap = Propel::getServiceContainer()->getDatabaseMap($this->getDbName());
-
-        if ($this->isIdentifierQuotingEnabled()) {
-            $adapter = Propel::getServiceContainer()->getAdapter($this->getDbName());
-
-            return $adapter->quoteIdentifierTable($string);
-        }
-
-        if ($dbMap->hasTable($realTableName)) {
-            $tableMap = $dbMap->getTable($realTableName);
-            if ($tableMap->isIdentifierQuotingEnabled()) {
-                $adapter = Propel::getServiceContainer()->getAdapter($this->getDbName());
-
-                return $adapter->quoteIdentifierTable($string);
             }
         }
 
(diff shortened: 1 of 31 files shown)


mringler:
Hmm, tests fail before they even start:
```
Tests started in temp /tmp.
Could not read "tests/sqlite.phpunit.xml".
Error: Process completed with exit code 2.
```
I know it is what everybody says, but I don't see how this is my fault. I have not changed the phpunit.xml files.

Please let me know how to proceed.

mringler:
Added test fixed from #1751. I'll remove them and rebase when that is resolved.

src/Propel/Runtime/ActiveQuery/QueryExecutor/CountQueryExecutor.php
```diff
@@ -0,0 +1,47 @@
...
+
+        return $executor->runCount();
+    }
+
+    /**
+     * Execute a count statement.
+     *
+     * @return \Propel\Runtime\DataFetcher\DataFetcherInterface
+     */
+    public function runCount(): DataFetcherInterface
```
dmytro-dymarchuk, line 40: This public method is not used anywhere except in this class so it could be `protected`.

dmytro-dymarchuk: Please check all children of `AbstractQueryExecutor`.

src/Propel/Runtime/ActiveQuery/QueryExecutor/InsertQueryExecutor.php
```diff
@@ -0,0 +1,127 @@
...
+        $pkFullName = $this->primaryKeyColumn->getFullyQualifiedName();
+        if ($this->criteria->keyContainsValue($pkFullName)) {
+            return;
+        }
+
+        $keyInfo = $this->primaryKeyColumn->getTable()->getPrimaryKeyMethodInfo();
+        $id = null;
+        try {
+            $id = $this->adapter->getId($this->con, $keyInfo);
+        } catch (Exception $e) {
```
dmytro-dymarchuk, line 104: `Throwable` will cover more cases

src/Propel/Runtime/ActiveQuery/SqlBuilder/CountQuerySqlBuilder.php
```diff
@@ -0,0 +1,67 @@
...
+                throw new LogicException($errorMessage);
+            }
+
+            $this->adapter->turnSelectColumnsToAliases($this->criteria);
+        }
+        [$baseSelectSql, $params] = SelectQuerySqlBuilder::createSelectSql($this->criteria);
+
+        $countStatment = "SELECT COUNT(*) FROM ($baseSelectSql) propelmatch4cnt";
+
+        return [$countStatment, $params];
```
dmytro-dymarchuk, line 65: I suggest for all similar places create DTO that contains statement and params to make return type more strict and more clear.

src/Propel/Runtime/Adapter/SqlAdapterInterface.php
```diff
@@ -55,10 +55,11 @@ public function cleanupSQL(&$sql, array &$params, Criteria $values, DatabaseMap
      * @param string $sql
      * @param int $offset
      * @param int $limit
+     * @param \Propel\Runtime\ActiveQuery\Criteria|null $criteria
      *
      * @return void
      */
-    public function applyLimit(&$sql, $offset, $limit);
+    public function applyLimit(&$sql, $offset, $limit, $criteria = null);
```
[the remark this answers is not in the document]
mringler: Oh, context for the change is that this method was declared here as
```php
public function applyLimit(&$sql, $offset, $limit);
```
and was implemented like this in most adapters, but not in `OracleAdapter`, where it is implemented like this:
```php
public function applyLimit(&$sql, $offset, $limit, $criteria = null)
```
Which confused me, since I had not looked at `OracleAdapter`, but saw the method being called with a fourth parameter, which should be wrong, according to the interface.
So I just added the parameter to reflect the actual signature of the method. But it is implemented as given here, without a type hint. I don't mind adding it, but it does mean adding it to existing public methods, and you guys usually don't want to do that. Should I still add it?

src/Propel/Runtime/ActiveQuery/QueryExecutor/CountQueryExecutor.php
mringler: This means that the query executors can only be used through their static `execute` method. So I have also changed all constructors to `protected`. Time will tell how this will fare in the long run.

dmytro-dymarchuk: > This means that the query executors can only be used through their static execute method.
So you don't need another public method if you want to call executor only via static method.

> So I have also changed all constructors to protected.
Then we even don't have the possibility to get an instance of executor anywhere and we will not have the ability to call `runCount` even in the future.

mringler: You just repeated what I said, right?

dmytro-dymarchuk: Sorry, It looks like repetition now. But I understood your phrase "Time will tell how this will fare in the long run." that you want to keep it public for the future. And I wanted to underline that it should not be. Now I see that it is protected and my comment is excess. Sorry, for the misunderstanding.

mringler: Ah, I see. Thank you for clearing it up.
Had to change the constructor from `AbstractQueryExecutor` (which most of the classes inherit directly) back to `public`, so that tests can override methods. I don't see a way around that.

src/Propel/Runtime/Adapter/SqlAdapterInterface.php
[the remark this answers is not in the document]
dereuromark: I think this is fine then as is.

dereuromark:
Small conflict btw.

mringler:
This is probably not going to get merged, so I am just going to close the PR.
Should you actually want to see this merged, let me know, and I'll fix the conflict and re-open.

dereuromark:
From what I can see it cleans up the baseline quite a bit
So it seems worth finishing this up IMO.

mringler:
Great, I have fixed the conflict

src/Propel/Runtime/ActiveQuery/SqlBuilder/InsertQuerySqlBuilder.php
```diff
@@ -0,0 +1,85 @@
...
+        $parameterPlaceholdersCsv = $this->buildParameterPlaceholdersCsv($numberOfColumns);
+
+        $insertStatement = "INSERT INTO $tableName ($columnCsv) VALUES ($parameterPlaceholdersCsv)";
+        $params = $this->buildParams($qualifiedColumnNames);
+
+        return new PreparedStatementDto($insertStatement, $params);
+    }
+
+    /**
+     * @param array $qualifiedColumnNames
```
dereuromark, line 57: string[]

src/Propel/Runtime/ActiveQuery/SqlBuilder/DeleteQuerySqlBuilder.php
```diff
@@ -0,0 +1,96 @@
...
+            $sql[] = "FROM $realTableName AS $tableName";
+        }
+
+        return implode(' ', $sql);
+    }
+
+    /**
+     * Build WHERE clause from the given column names.
+     *
+     * @param array $columnNames
```
dereuromark, line 80: Same here and in other places probably

src/Propel/Runtime/ActiveQuery/QueryExecutor/InsertQueryExecutor.php
```diff
@@ -0,0 +1,123 @@
...
+        $this->primaryKeyColumn = $criteria->getPrimaryKey();
+    }
+
+    /**
+     * @see InsertQuerySqlBuilder::runInsert()
+     *
+     * @param \Propel\Runtime\ActiveQuery\Criteria $criteria
+     * @param \Propel\Runtime\Connection\ConnectionInterface|null $con
+     *
+     * @return mixed|null
```
dereuromark, line 41: is it really mixed? isnt it only `string|int|null`?

dereuromark:
@nederdirk can also maybe help with the review for this to be merged today?

src/Propel/Runtime/ActiveQuery/QueryExecutor/AbstractQueryExecutor.php
```diff
@@ -0,0 +1,148 @@
...
+     * @param \Propel\Runtime\Connection\ConnectionInterface|null $con
+     */
+    public function __construct(Criteria $criteria, ?ConnectionInterface $con = null)
+    {
+        $this->criteria = $criteria;
+
+        $dbName = $criteria->getDbName();
+        $serviceContainer = Propel::getServiceContainer();
+
+        $this->con = ($con) ? $con : $this->getConnection($serviceContainer, $dbName, static::NEEDS_WRITE_CONNECTION);
```
dereuromark, line 63:
```suggestion
        $this->con = $con ?: $this->getConnection($serviceContainer, $dbName, static::NEEDS_WRITE_CONNECTION);
```

src/Propel/Runtime/ActiveQuery/SqlBuilder/InsertQuerySqlBuilder.php
mringler: Good point, I refined those array type declarations

src/Propel/Runtime/ActiveQuery/QueryExecutor/InsertQueryExecutor.php
mringler: I think you are right, but the return chain ends with AdapterInterface::getId(), where return type is set to `mixed`
```php
/**
     * Gets the generated ID (either last ID for autoincrement or next sequence ID).
     *
     * @param \Propel\Runtime\Connection\ConnectionInterface $con
     * @param string|null $name
     *
     * @return mixed
     */
    public function getId(ConnectionInterface $con, $name = null);
```
I don't think I can change this, can I?

dereuromark: We can change the docblock (and the actual truth of values expected to be returned). This will not have any consequences on the code itself.

src/Propel/Runtime/ActiveQuery/ModelCriteria.php
```diff
@@ -1044,17 +1052,32 @@ public function addSelectQuery(Criteria $subQueryCriteria, $alias = null, $addAl
      *
      * @param bool $force To enforce adding columns for changed alias, set it to true (f.e. with sub selects)
      *
-     * @return $this The current object, for fluid interface
+     * @return $this|\Propel\Runtime\ActiveQuery\ModelCriteria
```
nederdirk, line 1055: I'm usually pretty wary of this PHPstorm-ism. This type-hint (reads like 'this OR a `ModelCriteria` object, meaning 'this function might return an object reference to a `ModelCriteria` _other than_ `$this`) is less specific than `$this`. With the [pure-intersection-types RFC](https://wiki.php.net/rfc/pure-intersection-types) around the corner for PHP8.1 we should probably use either `$this` (even though it's nonstandard PHPdoc) or `\Propel\Runtime\ActiveQuery\ModelCriteria&$this` (to help phpdoc attribute interpreters that handle neither intersection types and the special value `$this`

Wow, that was a long piece of text for such a small issue.
TL;DR: don't assume PHPstorm is the only editor in the world 😜

src/Propel/Runtime/ActiveQuery/ModelCriteria.php
```diff
@@ -1044,17 +1052,32 @@ public function addSelectQuery(Criteria $subQueryCriteria, $alias = null, $addAl
...
+        $tableMapClassName = $this->modelTableMapName;
+        $alias = ($this->useAliasInSQL) ? $this->modelAlias : null;
+
+        return $this->addSelfSelectColumnsFromTableMapClass($tableMapClassName, $alias);
+    }
+
+    /**
+     * Adds the select columns for the given table.
+     *
+     * @param string $tableMapClassName
```
nederdirk, line 1073: I'm not privy to the coding standards in this project, but technically these `@param` annotations are redundant with the (stronger, therefore better :+1: ) native-PHP typehints in the function signature.

src/Propel/Runtime/ActiveQuery/QueryExecutor/InsertQueryExecutor.php
mringler: I'll be damned, stan and psalm don't complain. Nice!

src/Propel/Runtime/ActiveQuery/QueryExecutor/AbstractQueryExecutor.php
mringler: Yes, great, I love that shorthand form

src/Propel/Runtime/ActiveQuery/QueryExecutor/AbstractQueryExecutor.php
```diff
@@ -0,0 +1,148 @@
...
+use Propel\Runtime\ActiveQuery\Criteria;
+use Propel\Runtime\ActiveQuery\SqlBuilder\PreparedStatementDto;
+use Propel\Runtime\Connection\ConnectionInterface;
+use Propel\Runtime\Connection\ConnectionWrapper;
+use Propel\Runtime\Connection\StatementWrapper;
+use Propel\Runtime\Propel;
+use Propel\Runtime\ServiceContainer\ServiceContainerInterface;
+use Throwable;
+
+abstract class AbstractQueryExecutor
```
nederdirk, line 20: This may be a good opportunity to add comments about 'what is the structure of the subsystem these classes represent', i.e.
```
`QueryExecutor`'s responsibility is to translate a `Criteria` object to the specific RDBMS dialect and execute the query.
Use it like this: <insert example here>
```

src/Propel/Runtime/ActiveQuery/QueryExecutor/AbstractQueryExecutor.php
```diff
@@ -0,0 +1,148 @@
...
+    /**
+     * @var \Propel\Runtime\Map\DatabaseMap
+     */
+    protected $dbMap;
+
+    /**
+     * @param \Propel\Runtime\ActiveQuery\Criteria $criteria
+     * @param \Propel\Runtime\Connection\ConnectionInterface|null $con
+     */
+    public function __construct(Criteria $criteria, ?ConnectionInterface $con = null)
```
nederdirk, line 56: with abstract classes, it might be good to think about whether the constructor should be `final`: given a `class-string<AbstractQueryExecutor> $className`, should users of the Propel library be able to do `new $className($criteria, $connection);`?

If concrete implementing classes (i.e. `extends AbstractQueryExecutor`s) should be able to override the constructor, consider having `AbstractQueryExecutor` implement an `interface ConstructableQueryExecutor { function __construct(Criteria $criteria, ?ConnectionInterface $con = null); }`, so implementers are bound to that interface, too.

src/Propel/Runtime/ActiveQuery/QueryExecutor/AbstractQueryExecutor.php
```diff
@@ -0,0 +1,148 @@
...
+    /**
+     * Retrieves a read or write connection from the service container.
+     *
+     * @param \Propel\Runtime\ServiceContainer\ServiceContainerInterface $sc
+     * @param string $dbName
+     * @param bool $getWritableConnection
+     *
+     * @return \Propel\Runtime\Connection\ConnectionInterface
+     */
+    protected function getConnection(ServiceContainerInterface $sc, string $dbName, bool $getWritableConnection = true): ConnectionInterface
```
nederdirk, line 81: (⚠strong opinions following)
getters and/or setters usually hint at a broken design, leaking implementation details and making maintenance tricky as users of the Propel library _will_ start to depend on the specific exposed implementation.

Also: this `getConnection` function doesn't reference `$this` at all. Should it really be a method of this class?

nederdirk (commented):
I'm not done yet, but it's lunch time for me now :smile:

src/Propel/Runtime/ActiveQuery/ModelCriteria.php
mringler: Oh, I'm not using PhpStorm. If I remember correctly, one of stan or psalm complained about this somewhere, so I switched to always set both. It seems to work if I remove it here

src/Propel/Runtime/ActiveQuery/QueryExecutor/AbstractQueryExecutor.php
mringler: This isn't really a getter, the name was poorly chosen (changed it to `retrieveConnection`). I would have liked to make it a static method, but those are frowned upon in the project

src/Propel/Runtime/ActiveQuery/QueryExecutor/AbstractQueryExecutor.php
```diff
@@ -78,7 +78,7 @@ public function __construct(Criteria $criteria, ?ConnectionInterface $con = null
      *
      * @return \Propel\Runtime\Connection\ConnectionInterface
      */
-    protected function getConnection(ServiceContainerInterface $sc, string $dbName, bool $getWritableConnection = true): ConnectionInterface
+    private function retrieveConnection(ServiceContainerInterface $sc, string $dbName, bool $getWritableConnection = true): ConnectionInterface
```
dereuromark, line 81: Can we keep protected and not use private in this context?

mringler: Right, forgot about that, sorry

src/Propel/Runtime/ActiveQuery/ModelCriteria.php
mringler: The error is back:
```
1057 | ERROR | Chainable method (@return $this) cannot have multiple
      |       | return types in code.
      |       | (Spryker.Commenting.DocBlockReturnSelf.InvalidChainable)
```
I'll change it back

src/Propel/Runtime/ActiveQuery/QueryExecutor/AbstractQueryExecutor.php
mringler: I am not sure, what's the advantage here?

src/Propel/Runtime/ActiveQuery/Criteria.php
```diff
@@ -2295,38 +2102,6 @@ public function quoteIdentifier($string, $tableName = '')
         return $string;
     }
 
-    /**
-     * @param string $string
-     *
-     * @return string
-     */
-    public function quoteIdentifierTable($string)
```
gechetspr, line 2303: I would consider this as a BC break.
This is a internal class, but at least for us Criteria class is widely used.
@dereuromark I haven't found usage of it in our code base, but still.

Just to be sure, this is part of the major release, right?

src/Propel/Runtime/Adapter/Pdo/MssqlAdapter.php
```diff
@@ -130,19 +130,20 @@ public function random($seed = null)
...
+     * @param \Propel\Runtime\ActiveQuery\Criteria|null $criteria
      *
      * @throws \Propel\Runtime\Exception\InvalidArgumentException
      * @throws \Propel\Runtime\Adapter\Exception\ColumnNotFoundException
      * @throws \Propel\Runtime\Adapter\Exception\MalformedClauseException
      *
      * @return void
      */
-    public function applyLimit(&$sql, $offset, $limit)
+    public function applyLimit(&$sql, $offset, $limit, $criteria = null)
```
gechetspr, line 146: mb we can add typehints for new params?

src/Propel/Runtime/Adapter/Pdo/PdoAdapter.php
```diff
@@ -357,7 +367,7 @@ public function getGroupBy(Criteria $criteria)
     {
         $groupBy = $criteria->getGroupByColumns();
         if ($groupBy) {
-            return ' GROUP BY ' . implode(',', $groupBy);
+            return 'GROUP BY ' . implode(',', $groupBy);
```
gechetspr, line 370: is it covered on other place? Looks like this space was added in purpose

src/Propel/Runtime/Adapter/Pdo/PdoAdapter.php
```diff
@@ -451,27 +435,29 @@ public function createSelectSqlPart(Criteria $criteria, &$fromClause, $aliasAll
...
+                    // COUNT(DISTINCT books.price)
+                    $tableName = substr($columnName, $parenPos + 1, $dotPos - ($parenPos + 1));
+                    $lastSpace = strrpos($tableName, ' ');
+                    if ($lastSpace !== false) { // COUNT(DISTINCT books.price)
+                        $tableName = substr($tableName, $lastSpace + 1);
                     }
                 }
+                // is it a table alias?
+                $tableName2 = $criteria->getTableForAlias($tableName);
+                if ($tableName2 !== null) {
```
gechetspr, line 456: I can see that this is a copy paste, but personally I don't like `else` statements :)

if ($tableName2 !== null) {
$tableName2 = $tableName2 . '  ';
}
$fromClause[] = (string)$tableName2 . $tableName;

But this is not a mandatory to fix

gechetspr (commented):
Few minor things. Not sure if they are blocker, but IMO nice to fix

src/Propel/Runtime/ActiveQuery/Criteria.php
dereuromark: Yes, a beta release may do small BC breaks where justified. We outline those in the release notes.

dereuromark:
I see 1-2 important comments, otherwise we can probably continue for the beta release.
Lets see if we can merge this today then.

src/Propel/Runtime/Adapter/Pdo/PdoAdapter.php
mringler: Yes, that formatting is done in SelectQueryBuilder line 63-78;
```php
        $groupBy = $this->adapter->getGroupBy($this->criteria);
        if ($groupBy) {
            $this->criteria->replaceNames($groupBy);
            $sqlClauses[] = $groupBy;
        }

        $havingSql = $this->buildHavingClause($params);
        if ($havingSql) {
            $sqlClauses[] = $havingSql;
        }

        if ($orderBySql) {
            $sqlClauses[] = $orderBySql;
        }

        $sql = implode(' ', $sqlClauses);
```
I really disliked how those spaces were scattered around the codebase

dereuromark (approved):

gechetspr (approved):

dereuromark:
This one fail

> $DB_USER not set. Using 'root'.
/usr/bin/mysql  Ver 14.14 Distrib 5.7.35, for Linux (x86_64) using  EditLine wrapper

looks unrelated

dereuromark:
Should this be a squash merge?

mringler:
One of the mysql tests always fails, but I think it is a problem with the pipeline (it's a different one each time)
https://github.com/propelorm/Propel2/pull/1747/checks?check_run_id=3302733292
https://github.com/propelorm/Propel2/runs/3302662191?check_suite_focus=true

mringler:
>Should this be a squash merge?
I'd prefer it

PR #1749, opened 2021-06-09
Fixed usage of default on-update and on-delete behavior

### LINKED ISSUE #1238: Migrations with onDelete Restrict
In the schema, if I have a foreign key defined with 
`onDelete="RESTRICT"`

Each time a diff command is called, a migration class is generated to update the FK, even if the current onDelete value is already RESTRICT.

To fix this, I omit the onDelete tag, as Restrict is the default, but I would think that it should be possible to specify a RESTRICT value in the schema.

with propel 2.0.0-dev mysql  5.5.49

djairhogeuens:
Fixes #1238

I did not manage to run the unit tests as I always get an out of memory error :/ If anyone can help me with this, I'm happy to look into it.


djairhogeuens:
@dereuromark I see you have added the BC-break 2.0 label but I don't see why this change is breaking backwards compatibility?

mringler:
Hey @djairhogeuens,

first off, thank you for your contribution! This is pretty nice. I'll try to help get it merged.

I think the `BC-break` label was added because you change signatures of public methods. My guess is that test currently would fail because of it. So we need to get tests running on your end.
You say you get an out of memory error. On my machine, phpunit takes about 150 MB to run the tests. You can try to change memory limit by running it with the according parameter, both `phpunit -d memory_limit=150M` and `composer run-script test:mysql -- -d  memory_limit=150M` should work.
But before you do that, please make sure you also get the error when running against current master, so we know the memory problem does not come from your changes. For example, the mentioned changed signature might cause loops during setUp.
If you don't get an out of memory error, you will most likely see an error after the fixtures were built (see #1748). In that case, just restart the tests, you will not see the error again until you rebuild the fixtures..

If you haven't done it, you might also want to run `composer run-script stan`, `composer run-script psalm` and `composer run-script cs-check` (and `cs-fix`) for static analysis and code style check and fix. Your PR will need to pass theses checks before merging, but it might give you a hint if something is not working now.

Let me know when tests run and pass on your end, then I'll review the PR. Question right now: I see that you added `$fromDatabase` and `$toDatabase` members and method arguments, where `$fromTable` and `$toTable` is also available. Since Table objects have a `->getDatabase()` method, I am wondering if there is a reason why you are not using those?

Best,
Moritz

djairhogeuens:
> But before you do that, please make sure you also get the error when running against current master, so we know the memory problem does not come from your changes.

Hi @mringler I have checked whether I can run the tests against current master and that doesn't work either:
`PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 471040 bytes) in C:\local\Propel2\src\Propel\Generator\Builder\Om\AbstractOMBuilder.php on line 1113`

mringler:
Ok, seems like your system caps memory consumption at around 135 MB. Have you tried `composer run-script test:mysql -- -d memory_limit=150M`?

djairhogeuens:
@mringler I managed to get the memory limit a bit higher by adapting my local php.ini (so I don't need to pass the argument manually every time). But I still can't get the tests to pass properly. I'm currently running them on the latest master commit (so no changes from my side).

Just a quick snippet of the last lines of my logs (which are incredibly long):
```
In PdoConnection.php line 82:

  [PDOException (2002)]
  SQLSTATE[HY000] [2002] No connection could be made because the target machine actively refused it


Exception trace:
  at C:\local\Propel2\src\Propel\Runtime\Connection\PdoConnection.php:82
 PDO->__construct() at C:\local\Propel2\src\Propel\Runtime\Connection\PdoConnection.php:82
 Propel\Runtime\Connection\PdoConnection->__construct() at C:\local\Propel2\src\Propel\Runtime\Adapter\Pdo\PdoAdapter.php:69
 Propel\Runtime\Adapter\Pdo\PdoAdapter->getConnection() at C:\local\Propel2\src\Propel\Runtime\Connection\ConnectionFactory.php:40
 Propel\Runtime\Connection\ConnectionFactory::create() at C:\local\Propel2\src\Propel\Generator\Manager\SqlManager.php:224
 Propel\Generator\Manager\SqlManager->getConnectionInstance() at C:\local\Propel2\src\Propel\Generator\Manager\SqlManager.php:187
 Propel\Generator\Manager\SqlManager->insertSql() at C:\local\Propel2\src\Propel\Generator\Command\SqlInsertCommand.php:69
 PHPUnit\TextUI\TestRunner->run() at C:\local\Propel2\vendor\phpunit\phpunit\src\TextUI\Command.php:143
 PHPUnit\TextUI\Command->run() at C:\local\Propel2\vendor\phpunit\phpunit\src\TextUI\Command.php:96
 PHPUnit\TextUI\Command::main() at C:\local\Propel2\vendor\phpunit\phpunit\phpunit:61

test:prepare [--vendor VENDOR] [--dsn DSN] [-u|--user USER] [-p|--password [PASSWORD]] [--exclude-database]

C:\local\Propel2\tests\Propel\Tests\TestCaseFixtures.php:101
C:\local\Propel2\tests\Propel\Tests\TestCaseFixtures.php:81
C:\local\Propel2\tests\Propel\Tests\Helpers\Bookstore\BookstoreTestBase.php:37

ERRORS!
Tests: 1224, Assertions: 1161, Errors: 65, Failures: 1156.
Script phpunit --colors=always handling the test event returned with error code 2
Script @test -c tests/mysql.phpunit.xml was called via test:mysql
```

Am I missing something here? Sorry for my noobness ;)

mringler:
Have you tried googling the error message? Looking for ```No connection could be made because the target machine actively refused it``` gives me
>Either there is a firewall blocking the connection or the process that is hosting the service is not listening on that port, this may be because it is not running at all or because it is listening on a different port.

Propel test connects to MySql at 127.0.0.1 at the default port (3306) as user "root" with an empty password (not great, but this is how it is at the moment).
If you can connect like this from a MySql client, Propel test should be able to do it, too. So I would start there. Before running the tests, the database has to be initialized by the script at  `tests/bin/setup.mysql.sh' (or `\tests\bin\setup.mysql.bat` on windows), and you can test with that if Propel can access MySql.

The setup with SQLite is much easier. If I remember correctly, you only need to have the php module and SQLite installed. There is also an init script (`tests/bin/setup.sqlite.sh'), but no further configuration is necessary. If you expand your changes to that adapter, it might save you some trouble. Then again, who doesn't love to hate the occasional hour spent wrestling with MySql setup. Good luck!

djairhogeuens:
Okay great, I did not realize I needed to do any specific database setup
for running the tests... Is this stuff documented anywhere because I feel
like I'm asking things that I could just read somewhere as well. I tried
looking for information for new contributors but apart from coding
conventions, I could not find anything...

mringler:
Right, yes, I shouldn't have assumed you have already found that page:
http://propelorm.org/documentation/cookbook/working-with-test-suite.html
Some of it is outdated though, but should point you in the right direction.

djairhogeuens:
@mringler Thank you so much for pointing me to that page and for your help in general! I got the local setup working with a local MySQL db now.

I also improved the implementation as per your suggestion (I had indeed missed the getDatabase() and getTable() methods which allowed me to strip a lot of the code I had before). All tests should be passing now as well as the statical code analysis and code style checks.

mringler (commented):
Happy to hear you go the tests running!
Now, time to pick some nit. Hope it helps.

src/Propel/Generator/Platform/MysqlPlatform.php
```diff
@@ -952,6 +952,26 @@ public function doQuoting($text)
...
     }
 
+    /**
+     * Get the default On Delete behavior for foreign keys when not explicity set.
+     *
+     * @return string
+     */
+    public function getDefaultForeignKeyOnDeleteBehavior()
+    {
+        return ForeignKey::RESTRICT;
```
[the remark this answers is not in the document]
djairhogeuens: It seems like the docs for version 8 are incorrect. I am running MySQL Community Server 8.0.25 (the latest version) and when I execute a CREATE TABLE statement with a foreign key without specifying the ON DELETE or ON UPDATE behavior, it sets it to RESTRICT, not NO ACTION. I have actually implemented the distinction between version 5 and 8 as you suggested but when I test it on my database, the bug that I am trying to solve is still there because of the behavior I just mentioned. So I will revert that "compatibility" fix.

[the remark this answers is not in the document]
djairhogeuens: I did an extra test where I now executed a CREATE TABLE statement with a foreign key with an explicit ON DELETE behavior set to NO ACTION. Turns out that this is just created as a RESTRICT behavior. So this does not seem to be a problem as specifying the NO ACTION behavior is not supported according to the Propel XML schema docs (http://propelorm.org/documentation/reference/schema.html#foreign-key-element).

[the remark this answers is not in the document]
mringler: Hmm, that's weird. I am running MySql server version 8.0.25-0ubuntu0.20.04.1, and it behaves exactly as specified in the docs:
```sql
create table t(
	id1 int,
    foreign key (id1) references book(id) on delete no action on update no action,
    foreign key (id1) references book(id) on delete restrict on update restrict,
    foreign key (id1) references book(id) on delete set null on update set null,
    foreign key (id1) references book(id) on delete cascade on update cascade
);
show create table t;
```
gives
```
'CREATE TABLE `t` (
  `id1` int DEFAULT NULL,
  KEY `id1` (`id1`),
  CONSTRAINT `t_ibfk_1` FOREIGN KEY (`id1`) REFERENCES `book` (`id`),
  CONSTRAINT `t_ibfk_2` FOREIGN KEY (`id1`) REFERENCES `book` (`id`) ON DELETE RESTRICT ON UPDATE RESTRICT,
  CONSTRAINT `t_ibfk_3` FOREIGN KEY (`id1`) REFERENCES `book` (`id`) ON DELETE SET NULL ON UPDATE SET NULL,
  CONSTRAINT `t_ibfk_4` FOREIGN KEY (`id1`) REFERENCES `book` (`id`) ON DELETE CASCADE ON UPDATE CASCADE
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci'
```
So it does indeed omit the `NO ACTION`, and not `RESTRICT`. I suppose we cannot rely on version alone. Also, MariaDB uses the same platform, but with different versioning (and it seems to use `RESTRICT` consistently).

If you want to go through the extra trouble, you can easily adjust your code to be able to detect both options as default (i.e. by changing `getDefaultOnDeleteBehavior()` to `isDefaultOnDeleteBehavior($detectedBehavior))`. Considering how there is an easy workaround for the issue - not declaring the default behavior - I think this is ok then as it is.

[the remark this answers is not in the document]
mringler: Oh, just for reference, the propel doc is misleading. You actually can use `onDelete="no action"`, there just is no alias like `setnull`.

mringler:
Alright, almost done. Finally, could you please squash your commits into one, and force push that? It keeps the commit history free from clutter.

dereuromark:
Personally, I like PRs more that are not squashed, especially after review (or partial review before more commits).
It makes re-review much easier, and we could always squash merge the PR itself then.

Right now, the problem with many PRs that are repeatedly squashed: It is impossible to only review the changed part, as it got lost now in between all other changes.

mringler:
@dereuromark Alright, thanks for clearing that up

mringler (approved):

src/Propel/Generator/Platform/MysqlPlatform.php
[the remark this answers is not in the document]
djairhogeuens: Wait, I think you misunderstood me. The behavior you are seeing is also what I am seeing. So if you create a FK with behavior RESTRICT, NO ACTION or without defining it, it actually sets RESTRICT. So, if the Propel XML declares either of those three options, it should match RESTRICT. Currently, it will only give a match if RESTRICT is set or if the behavior is not set (empty or undefined). So if we indeed want to support declaring `onDelete="no action"` then the implementation should be further extended. But then indeed I'm not sure whether that will also be compatible with MySQL 5 and MariaDB.  It depends on how they behave when a NO ACTION behavior is specified. If it also results in a RESTRICT, it should be fine. If not, the version should be checked first.

djairhogeuens:
@dereuromark I think the BC-break 2.0 label can be removed now as well, after my last changes :)

src/Propel/Generator/Platform/MysqlPlatform.php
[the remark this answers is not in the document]
djairhogeuens: Ok, just tested and it appears that MariaDB explicitly sets NO ACTION if that is specified. So that means that if we want to support the `onDelete="no action"` declaration in the XML that we will have to differentiate between MySQL and MariaDB at least OR just interpret both behaviors as equal (e.g. RESTRICT = NO ACTION, also for MariaDB). Or, we can leave it as is and indeed advise to omit it in the XML instead. Let me know which option you prefer so we can wrap this up.

dereuromark:
This now actually might need rebase etc, I dont see CI running or green.

djairhogeuens:
@dereuromark I just did a rebase and it's still not starting it seems...

dmytro-dymarchuk:
> This now actually might need rebase etc, I dont see CI running or green.

As described below, It is because: "1 workflow awaiting approval. First-time contributors need a maintainer to approve running workflows." [Link](https://docs.github.com/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks)

src/Propel/Generator/Model/ForeignKey.php
```diff
@@ -168,19 +168,20 @@ protected function doNaming()
      * Returns the normalized input of onDelete and onUpdate behaviors.
      *
      * @param string|null $behavior
+     * @param string|null $default
      *
      * @return string
      */
-    public function normalizeFKey($behavior)
+    public function normalizeFKey($behavior, $default = null)
```
dmytro-dymarchuk, line 175: Small suggestion. You can use type-hint `?string` since it is a new code and it doesn't break BC. It is actual and for other places with new code.

dmytro-dymarchuk (approved):

dereuromark:
Thank you, from now on any commit should run fine for you :)

src/Propel/Generator/Platform/DefaultPlatform.php
```diff
@@ -1442,6 +1442,26 @@ public function getDateFormatter()
         return 'Y-m-d';
     }
 
+    /**
+     * Get the default On Delete behavior for foreign keys when not explicity set.
+     *
+     * @return string
+     */
+    public function getDefaultForeignKeyOnDeleteBehavior()
```
dereuromark, line 1450: here also return type could be added etc for new code

djairhogeuens: Done

src/Propel/Generator/Model/ForeignKey.php
djairhogeuens: Done

src/Propel/Generator/Platform/MysqlPlatform.php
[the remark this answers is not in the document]
djairhogeuens: Ok, after some struggling with unit tests I think I managed to implement a solution that will solve all issues for both MySQL 5 and 8 and also for MariaDB.

The situation before this PR can be described as follows:
|db|xml|result in db|reverse parsed as|comparison|
|---|---|---|---|---|
|mysql5|null|null|null|match|
|mysql5|restrict|null|null|no match|
|mysql5|no action|no action|no action|match|
|mysql8|null|null|null|match|
|mysql8|restrict|restrict|null|no match|
|mysql8|no action|null|null|no match|
|mariadb|null|null|null|match|
|mariadb|restrict|null|null|no match|
|mariadb|no action|no action|no action|match|

This PR takes away a mapping that is done during the reverse parsing and adds a fallback to a DB version-specific default:
|db|xml|result in db|reverse parsed as|comparison|
|---|---|---|---|---|
|mysql5|null|null|null|match (null defaults to restrict)|
|mysql5|restrict|null|null|**match (null defaults to restrict)**|
|mysql5|no action|no action|no action|match|
|mysql8|null|null|null|match (null defaults to no action)|
|mysql8|restrict|restrict|**restrict**|**match**|
|mysql8|no action|null|null|**match (null defaults to no action)**|
|mariadb|null|null|null|match (null defaults to restrict)|
|mariadb|restrict|null|null|**match (null defaults to restrict)**|
|mariadb|no action|no action|no action|match|

djairhogeuens:
@dereuromark apparently an approval is needed again after new commits :)

src/Propel/Generator/Platform/MysqlPlatform.php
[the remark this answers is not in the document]
djairhogeuens: Okay no, I reverted this as I realized (after testing more intensively) that the schema that is created based on the XML to compare with the reversed schema from the actual database does not contain an actual database connection so it has no way to know which database version should be used for the XML, resulting in a null error when trying to get the database version attribute for the XML database. I honestly don't see a way to address this problem anymore :/

If you still have ideas or suggestions, please share but otherwise I would propose to leave the PR as-is for now, knowing that the NO ACTION behavior won't work correctly when using it with MySQL 8.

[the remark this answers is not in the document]
djairhogeuens: Okay, sorry for the spamming but I think I finally found a permanent solution for the connection problem. So hopefully, this PR can be wrapped up now ;)

src/Propel/Generator/Platform/MysqlPlatform.php
```diff
@@ -952,6 +952,42 @@ public function doQuoting($text)
...
+     * Get the default On Delete behavior for foreign keys when not explicity set.
+     *
+     * @return string
+     */
+    public function getDefaultForeignKeyOnDeleteBehavior(): string
+    {
+        preg_match('/^(\d+)\.(\d+)\.(\d+)-([^-]+)-?(.*)/', $this->con->getAttribute(PDO::ATTR_SERVER_VERSION), $versionComponents);
+        if (isset($versionComponents[4]) && strcasecmp($versionComponents[4], 'mariadb') != 0) {
+            // assume MySQL
+            if ($versionComponents[1] == '8') {
```
mringler, line 965: Several remarks here:
- this code is not doing `getDefaultForeignKeyOnDeleteBehavior` - it should not be in this method
- the server version string is set by the package builder, trying to match it this closely is extremely fragile
- for now, I would assume that MySQL will stick with `NO ACTION` in later versions, not just 8

Please change this to something like:
```php
class MysqlPlatform extends DefaultPlatform
{
    protected $serverVersion;

    protected function getServerVersion(): ?string
    {
        if(!$this->serverVersion && $this->con) {
            $this->serverVersion = $this->con->getAttribute(PDO::ATTR_SERVER_VERSION);
        }

        return $this->serverVersion;
    }

    protected function getMayorVersionNumber(): ?int
    {
        $serverVersion = $this->getServerVersion();
        if (!$serverVersion) {
            return null;
        }
        $dotPos = strpos($serverVersion, '.');
        if ($dotPos === false) {
            return null;
        }

        return (int) substr($serverVersion, 0, $dotPos - 1);
    }

    protected function isMariaDB(): bool
    {
        $serverVersion = $this->getServerVersion() ?? '';

        return (stripos($serverVersion, 'mariadb') !== false);
    }

    public function getDefaultForeignKeyOnDeleteBehavior(): string
    {
        $mayorVersion = $this->getMayorVersion;

        return ($mayorVersion && $mayorVersion >= 8 && !$this->isMariaDB()) : ForeignKey::NOACTION : ForeignKey::RESTRICT;
    }

    public function getDefaultForeignKeyOnUpdateBehavior(): string
    {
        $mayorVersion = $this->getMayorVersion;

        return ($mayorVersion && $mayorVersion >= 8 && !$this->isMariaDB()) : ForeignKey::NOACTION : ForeignKey::RESTRICT;
    }
```

src/Propel/Generator/Platform/MysqlPlatform.php
```diff
@@ -952,6 +952,42 @@ public function doQuoting($text)
...
+
+        return ForeignKey::RESTRICT;
+    }
+
+    /**
+     * Get the default On Update behavior for foreign keys when not explicity set.
+     *
+     * @return string
+     */
+    public function getDefaultForeignKeyOnUpdateBehavior(): string
```
mringler, line 978: see above

src/Propel/Generator/Reverse/MysqlSchemaParser.php
```diff
@@ -410,30 +410,22 @@ protected function addForeignKeys(Table $table)
...
+                    'ON UPDATE' => null,
                 ];
 
                 if ($fkey) {
-                    // split foreign key information -> search for ON DELETE and afterwords for ON UPDATE action
+                    // split foreign key information -> search for ON DELETE and afterwards for ON UPDATE action
                     foreach (array_keys($fkactions) as $fkaction) {
                         $result = null;
-                        preg_match('/' . $fkaction . ' (' . ForeignKey::CASCADE . '|' . ForeignKey::SETNULL . ')/', $fkey, $result);
+                        preg_match('/' . $fkaction . ' (' . ForeignKey::CASCADE . '|' . ForeignKey::SETNULL . '|' . ForeignKey::RESTRICT . '|' . ForeignKey::NOACTION . ')/', $fkey, $result);
```
mringler, line 422: Might as well clean it up while you are working on it:
```php
$availableActions = [ForeignKey::CASCADE, ForeignKey::SETNULL, ForeignKey::RESTRICT, ForeignKey::NOACTION];
$pipedActionsString = implode('|', $availableActions);
$regex = sprintf('/ %s (%s)/',$fkaction,  $pipedActionsString);
preg_match($regex, $fkey, $result);
```
Personally, I would put this into its own method `getForeignKeyBehaviors()`, but this might be asking too much.

tests/Propel/Tests/Generator/Platform/PlatformTestBase.php
```diff
@@ -9,16 +9,20 @@
...
-abstract class PlatformTestBase extends TestCase
+abstract class PlatformTestBase extends TestCaseFixturesDatabase
 {
     protected function getDatabaseFromSchema($schema)
     {
-        $xtad = new SchemaReader($this->getPlatform());
+        $platform = $this->getPlatform();
+        require_once __DIR__ . '/../../../../Fixtures/platform/build/conf/platform-conf.php';
+        $platform->setConnection(Propel::getConnection('platform'));
+        $xtad = new SchemaReader($platform);
```
mringler, line 25: With the changes above, this is not necessary anymore. Please undo it.
If there was a need to add a connection, we should use an existing one (like `tests/Fixtures/bookstore/build/conf/bookstore-conf.php`)

mringler (changes requested):
Seems like we are getting somewhere. Great that you figured out a way to check database type and version.

There are, however, several severe style-nonos in there. I added the code that (I think) is needed to fix this in my comments, as I don't want to keep you guessing. Feel free to add you personal spin on it, but make sure to strive for similar cleanliness. There is a lot of smelly code in Propel, let's try to make it better, and not add to it.

Let me know if you want me to go into details why those changes are necessary.

src/Propel/Generator/Model/Diff/ForeignKeyComparator.php
```diff
@@ -63,10 +63,12 @@ public static function computeDiff(ForeignKey $fromFk, ForeignKey $toFk, $caseIn
         }
 
         // compare on
-        if ($fromFk->normalizeFKey($fromFk->getOnUpdate()) !== $toFk->normalizeFKey($toFk->getOnUpdate())) {
+        $fromPlatform = $fromFk->getTable()->getDatabase() ? $fromFk->getTable()->getDatabase()->getPlatform() : null;
+        $toPlatform = $toFk->getTable()->getDatabase() ? $toFk->getTable()->getDatabase()->getPlatform() : null;
+        if ($fromFk->normalizeFKey($fromFk->getOnUpdate(), $fromPlatform ? $fromPlatform->getDefaultForeignKeyOnUpdateBehavior() : null) !== $toFk->normalizeFKey($toFk->getOnUpdate(), $toPlatform ? $toPlatform->getDefaultForeignKeyOnUpdateBehavior() : null)) {
```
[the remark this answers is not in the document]
djairhogeuens: Done

src/Propel/Generator/Platform/MysqlPlatform.php
djairhogeuens: Done

tests/Propel/Tests/Generator/Platform/PlatformTestBase.php
djairhogeuens: Done

src/Propel/Generator/Reverse/MysqlSchemaParser.php
djairhogeuens: Done

src/Propel/Generator/Platform/MysqlPlatform.php
djairhogeuens: Done

src/Propel/Generator/Model/Diff/ForeignKeyComparator.php
```diff
@@ -63,10 +63,12 @@ public static function computeDiff(ForeignKey $fromFk, ForeignKey $toFk, $caseIn
...
             return true;
         }
-        if ($fromFk->normalizeFKey($fromFk->getOnDelete()) !== $toFk->normalizeFKey($toFk->getOnDelete())) {
+        if ($fromFk->normalizeFKey($fromFk->getOnDelete(), $fromPlatform ? $fromPlatform->getDefaultForeignKeyOnDeleteBehavior() : null) !== $toFk->normalizeFKey($toFk->getOnDelete(), $toPlatform ? $toPlatform->getDefaultForeignKeyOnDeleteBehavior() : null)) {
```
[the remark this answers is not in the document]
djairhogeuens: Done

mringler (approved):
Awesome. Looks really good to me now.

djairhogeuens:
@mringler @dereuromark seems like the workflow for this PR is still waiting for an approval?

djairhogeuens:
@mringler @dereuromark checks are passing so this can be merged. Don't know who has write access to do this?

mringler:
Awesome! It might take a few days until the maintainers will get to a final review and then merging it. But it really should only be a matter of time now.
Again, thank you for contributing! It was fun seeing this through with you.
There are always bugs to fix or things to review, so let me say, you are welcome to keep participating.

dereuromark (approved):

PR #1750, opened 2021-06-10
cleaning XmlToArrayConverter

ikeyan:


src/Propel/Common/Config/XmlToArrayConverter.php
```diff
@@ -36,17 +36,17 @@ public static function convert($xmlToParse)
         $isFile = file_exists($xmlToParse);
 
         //Empty xml file returns empty array
```
mringler, line 38: The comment does not make sense here. Would you mind removing it?

mringler (commented):
Really nice, thank you.

The ill complaint from stan is fixed in #1751, when it is merged and you have rebases all should be green.

And on the subject of rebasing - would you mind squashing your commits into one and force-pushing the remaining single commit? This really helps to keep the commit history from completely cluttering up.

ikeyan:
@mringler
I'd like to add ext-dom, ext-simplexml and ext-libxml to the composer.json.
I see ext-* under `require-dev`, but I think they should be moved into `require` because Propel2 actually requires them to work properly.

dereuromark:
I would make such ext-* changes in a separate PR, to not halt this one.
It is a separate topic/discussion.

src/Propel/Common/Config/XmlToArrayConverter.php
ikeyan: moved to proper position

mringler (approved):

dmytro-dymarchuk (approved):

dereuromark:
Nice work on this one, everyone!
We finished a full PR in <1 day with 2+ reviews/approvals :)

PR #1751, opened 2021-06-10
another test fix: reset counters before building test fixtures

mringler:
Another bug when running test, and again I am not sure why it shows up now.

AggregateMultipleColumnsBehavior has an map where it keeps track of names of inserted methods, so that it can avoid inserting the same name multiple times. The counter has to be static, since it has to work across instances of the class.
When fixtures are reinitialized during a test run, this map is not reset, so when the methods are inserted into the new model, they are adjusted, as if they were duplicates. And then tests fail, when the actual method is not available.

The fix is quite simple, just reset the map when running tests.

The problem cannot occur outside tests, since there is no mechanism to run 'model:build' twice without reloading the classes.

I don't really understand, why the problem did not occur before. Might be because the tests were executed in a different order, and the  tests for AggregateMultipleColumnsBehavior ran before the second initialization? Anyway, it needs to be stable one way or another.


mringler:
And when it rains, it pours: Now stan first ignores two error pattern from the baseline file, and then complains about the occurrences (see [here](https://github.com/propelorm/Propel2/pull/1751/checks?check_run_id=2794909350)).

This only happens on CI, not when I run stan locally, and it complains about a file that was not changed.

Fixing the errors should solve the problem.

dereuromark (approved):

dmytro-dymarchuk (approved):

PR #1754, opened 2021-06-16
refactor getExecutedQueryString

ikeyan:


src/Propel/Runtime/Connection/StatementWrapper.php
```diff
@@ -220,24 +220,15 @@ public function execute($inputParameters = null)
      *
      * @return string
      */
-    public function getExecutedQueryString($inputParameters = null)
+    public function getExecutedQueryString(?array $inputParameters = null): string
```
dereuromark, line 223: If this is planned to be released as part of the upcoming beta release, we could think about keeping those, and also add other missing ones.
For a pure bugfix release it would be a bit too much.

cc @mringler

mringler: That's a good point. That mean there is no better time to add them than now, doesn't it?

Tell me if I am missing something: the big problem with adding type hints is how easily they cause errors in seemingly unrelated parts of the code, making it hard to see the true impact of such a change. With stan, psalm and the tests, we should be relatively safe, but those are famous last words. I would guess that every new type hint is still a crapshoot to some extend. So if we add enough, at some point, we will inevitably end up with a bug report.
Then again, the type hints won't add themselves. And while they can make things less flexible, it is probably still better to have them in almost all cases.

Personally, I think I like seeing them added. But I must admit, it is dicey and I am not sure if I am even completely aware of the extend of trouble it can create. I tend to say let's find out, and accept the risk of having to add a few quick fixes sometimes in the near future. It is the best time for that. But I'd completely understand if you decide that the risk outweighs the benefit.

mringler (approved):

mringler:
Really nice!
Not sure why the one test fails, but I have experienced the same error. Just pushing the commit again works for me:
```
git commit --amend --no-edit
git push -f
```

src/Propel/Runtime/Connection/StatementWrapper.php
gechetspr: It should not break anything or at least if it breaks smth, that means that smth was broken from the beginning.

But mb we can also add this method to the interface if we started to refactor it?

dereuromark:
OK, so we agree that we want to fix up the signatures for the upcoming beta release to contain
- param types
- return types

wherever this is feasible and reasonable - that means it should not break any code.
If current code could be in the "wrong type" we should rather skip then to avoid too many bc breaking issues.

PR #1757, opened 2021-06-21
add howto url to missing loader script exception

mringler:
In order to reduce confusion and resulting bug reports about an outdated configuration or missing database import, I have created a howto page in the Propel2 wiki at https://github.com/propelorm/Propel2/wiki/Exception-Target:-Loading-the-database, and I have added the URL to the exception message.

I am not sure if the wiki is the correct location, but I am not sure if the official documentation is still maintained, so the wiki seems like the next best thing. Please let me know if the howto should go somewhere else. I'd prefer a perma link

### DIFF
--- a/src/Propel/Runtime/ServiceContainer/StandardServiceContainer.php
+++ b/src/Propel/Runtime/ServiceContainer/StandardServiceContainer.php
@@ -39,9 +39,12 @@ class StandardServiceContainer implements ServiceContainerInterface
     /**
      * Used in exception when the configuration is outdated.
      *
+     * @see StandardServiceContainer::checkVersion()
+     * @see StandardServiceContainer::getDatabaseMap()
+     *
      * @var string
      */
-    public const HOWTO_FIX_MISSING_LOADER_SCRIPT_URL = 'https://github.com/propelorm/Propel2/wiki/Exception-Target:-Loading-the-database';
+    protected const HOWTO_FIX_MISSING_LOADER_SCRIPT_URL = 'https://github.com/propelorm/Propel2/wiki/Exception-Target:-Loading-the-database';
 
     /**
      * @var \Propel\Runtime\Adapter\AdapterInterface[] List of database adapter instances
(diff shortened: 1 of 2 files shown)


dereuromark (approved):
Given, that we want to start stabilizing the (beta) 2.0 release, we should now make a release that cleans up a lot.
This seems helpful to get there.

gechetspr (changes requested):
visibility of constant can definitely be protected, but the rest of changes are not mandatory

src/Propel/Runtime/ServiceContainer/StandardServiceContainer.php
```diff
@@ -230,7 +240,12 @@ public function checkVersion($generatorVersion)
             return;
         }
 
-        throw new PropelException('Your configuration is outdated. Please rebuild it with the config:convert command.');
+        $message = 'Your configuration is outdated. Please rebuild it with the config:convert command.';
+        if (!is_int($generatorVersion) || $generatorVersion < 2) {
+            $message .= sprintf(' Visist %s for information on how to fix this.', self::HOWTO_FIX_MISSING_LOADER_SCRIPT_URL);
```
dereuromark, line 245:
```suggestion
            $message .= sprintf(' Visit %s for information on how to fix this.', self::HOWTO_FIX_MISSING_LOADER_SCRIPT_URL);
```

src/Propel/Runtime/ServiceContainer/StandardServiceContainer.php
```diff
@@ -279,7 +294,11 @@ public function getDatabaseMap($name = null)
             $name = $this->getDefaultDatasource();
         }
         if ($this->databaseMaps === null) {
-            throw new PropelException('Database map was not initialized. Please check the database loader script included by your conf');
+            $messageFormat = 'Database map was not initialized. Please check the database loader script included by your conf. '
+                . 'Visist %s for information on how to fix this.';
```
dereuromark, line 298: Same here

mringler:
@gechetspr Thank you for the review. I updated the code as you requested.
@dereuromark Thanks for finding the typo

First CI run after push failed, when the connection to the database could not be established (see [here](https://github.com/propelorm/Propel2/pull/1757/checks?check_run_id=2884657122)). Currently, this happens every other time I push a commit.

PR #1760, opened 2021-06-23
Remove call_user_func(), call_user_func_array() and constant()

ikeyan:


mringler (approved):
Really nice. Just some small suggestions

mringler (approved):

src/Propel/Runtime/ActiveQuery/ModelCriteria.php
```diff
@@ -1955,8 +1962,12 @@ public function doUpdate($updateValues, ConnectionInterface $con, $forceIndividu
                 // this criteria updates only one object defined by a concrete primary key,
                 // therefore there's no need to remove anything from the pool
             } else {
-                call_user_func([$this->modelTableMapName, 'clearInstancePool']);
-                call_user_func([$this->modelTableMapName, 'clearRelatedInstancePool']);
+                $modelTableMapName = $this->modelTableMapName;
+                if (!isset($modelTableMapName)) {
```
dereuromark, line 1966: should be !== to avoid cloaking call on a must-be-initialized var, and use simple strict null comparison.

src/Propel/Runtime/Connection/ConnectionWrapper.php
```diff
@@ -457,11 +455,14 @@ public function query($statement)
      */
     public function callUserFunctionWithLogging(callable $callback, ?array $args, string $sqlForLog)
     {
+        if (!isset($args)) {
```
dereuromark, line 458: Same here

src/Propel/Runtime/ActiveQuery/ModelCriteria.php
ikeyan: done

src/Propel/Runtime/Connection/ConnectionWrapper.php
ikeyan: done

PR #1761, opened 2021-06-23
revert fix stan

ikeyan:
psalm's error was because `$k` can be float or bool.

### DIFF
--- a/src/Propel/Common/Config/XmlToArrayConverter.php
+++ b/src/Propel/Common/Config/XmlToArrayConverter.php
@@ -20,8 +20,6 @@ class XmlToArrayConverter
     /**
      * Create a PHP array from the XML file
      *
-     * @psalm-return array<scalar, mixed>
-     *
      * @param string $xmlToParse The XML file or a string containing xml to parse
      *
      * @throws \Propel\Common\Config\Exception\XmlParseException if parse errors occur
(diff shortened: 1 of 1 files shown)


src/Propel/Common/Config/XmlToArrayConverter.php
```diff
@@ -80,8 +78,6 @@ public static function convert($xmlToParse)
      *
      * @author Christophe VG (based on code form php.net manual comment)
      *
-     * @psalm-return array<scalar, mixed>
-     *
      * @param \SimpleXMLElement $xml SimpleXML object.
      *
      * @return array Array representation of SimpleXML object.
```
dereuromark, line 83: Can't you just use

    @return array<int|string,mixed>

instead etc?
Then we can avoid a non functional cast (as this is clear for keys given anyway) - we would only do it for phpstan.

mringler: Or just remove the call to `getConvertedXmlValue()`, and replace it with something like
```php
$k = (ctype_digit($av)) ? (int)$av : $av;
```
If you set an id to `trUe`, you don't want it to become `true`.

Nice find though! I remember not really understanding the psalm error.

ikeyan: done

mringler (approved):

dereuromark (approved):

PR #1763, opened 2021-06-24
Format insert DATE values as Y-m-d instead of 'Y-m-d H:i:s.u'

nederdirk:
In our system, we sometimes use `9999-12-31` as a special 'until end of time' value for date-columns. (We're developing against MySQL, which has `9999-12-31` as maximum for date columns. However, if the `DateTime` object accidentally had a time component, this would cause the column to overflow.

This PR contains a fix for this issue, ensuring that DATE columns will be inserted into the database in the `Y-m-d` format, dropping the time-component which is not expected on columns with a native DATE type.

### DIFF
--- a/tests/Propel/Tests/Generator/Builder/Om/GeneratedObjectDateColumnTypeTest.php
+++ b/tests/Propel/Tests/Generator/Builder/Om/GeneratedObjectDateColumnTypeTest.php
@@ -49,12 +49,12 @@ XML;
 
     public function testInsertDateColumn(): void
     {
-        assert(class_exists(GeneratedObjectDateColumnTypeEntity::class));
-        $e = new GeneratedObjectDateColumnTypeEntity();
-        $this->assertTrue(method_exists($e, 'setDatecolumn'));
-        $this->assertTrue(\method_exists($e, 'save'));
+        assert(\class_exists(GeneratedObjectDateColumnTypeEntity::class));
+        $entity = new GeneratedObjectDateColumnTypeEntity();
+        $this->assertTrue(\method_exists($entity, 'setDatecolumn'));
+        $this->assertTrue(\method_exists($entity, 'save'));
         $dateValue = new DateTimeImmutable('2021-06-25 12:26');
-        $e->setDatecolumn($dateValue);
+        $entity->setDatecolumn($dateValue);
 
         $insertStatement = $this->createMockInsertStatement();
         $insertStatement
@@ -76,7 +76,7 @@ XML;
                 );
                 return $insertStatement;
             });
-        $e->save($con);
+        $entity->save($con);
     }
 
     /**
--- a/tests/Propel/Tests/Generator/Builder/Om/GeneratedObjectDateColumnTypeTest.php
+++ b/tests/Propel/Tests/Generator/Builder/Om/GeneratedObjectDateColumnTypeTest.php
@@ -1,0 +1,117 @@
+<?php
+
+/**
+ * MIT License. This file is part of the Propel package.
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Propel\Tests\Generator\Builder\Om;
+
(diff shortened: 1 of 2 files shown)


dereuromark:
As a pragmatic solution, wouldnt having `9999-12-30` or even `9999-01-01` as date solve this issue?
Not like anyone else will ever use a date close to that. I wouldnt operate on the very limits of the DB here with this issue known.

dereuromark (approved):
That said, this sounds like a good fix
Can we maybe have a small test case for this edge case?

nederdirk:
> That said, this sounds like a good fix
> Can we maybe have a small test case for this edge case?

Hey @dereuromark,

thank you for your approval! I just pushed a test for this, although I'd be happy to remove that monster again :sweat_smile: I went for a no-database-needed Unit test, but it involved a tad more ugly mocking than I anticipated. Point is: the code I'm exercising here is database-independent, but the surrounding code is so incredibly stateful-database call heavy :shrug:

dereuromark:
Given that the tests still exec in ~40s, it overall doesn't add too much, the value of having this regression check in place weighs higher none the less :)

tests/Propel/Tests/Generator/Builder/Om/GeneratedObjectDateColumnTypeTest.php
```diff
@@ -0,0 +1,117 @@
...
+XML;
+            QuickBuilder::buildSchema($schema);
+        }
+    }
+
+    public function testInsertDateColumn(): void
+    {
+        assert(class_exists(GeneratedObjectDateColumnTypeEntity::class));
+        $e = new GeneratedObjectDateColumnTypeEntity();
+        $this->assertTrue(method_exists($e, 'setDatecolumn'));
```
stereomon, line 54: You sometimes prefix functions with a `\` and sometimes not. @dereuromark do we have a CS fixer to align those?

stereomon (changes requested):

tests/Propel/Tests/Generator/Builder/Om/GeneratedObjectDateColumnTypeTest.php
```diff
@@ -0,0 +1,117 @@
...
+    </table>
+</database>
+XML;
+            QuickBuilder::buildSchema($schema);
+        }
+    }
+
+    public function testInsertDateColumn(): void
+    {
+        assert(class_exists(GeneratedObjectDateColumnTypeEntity::class));
```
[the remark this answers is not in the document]
nederdirk: This assert is here to satisfy the psalm linting phase, among other things.

tests/Propel/Tests/Generator/Builder/Om/GeneratedObjectDateColumnTypeTest.php
dereuromark: So far we are not using this yet. If we would, any fixer should normalize this then later.

dereuromark: But as long as there is no sniffer for it to fix it in either direction, I would not bother about this now.

nederdirk: @stereomon I processed your feedback, could you greenlight this?

Also: you might be interested in https://github.com/slevomat/coding-standard#slevomatcodingstandardnamespacesfullyqualifiedglobalfunctions- for the global-namespaced functions :smile:

dereuromark: @nederdirk Yeah, we are so far not doing this as this is somewhat a nano-opt (small benefit) while having a lot of more noise in the use statements.
So far, we are using normal function usage (that also does allow overwriting for exception cases when needed).

dereuromark:
Given that all comments have been resolved, lets move forward.

Issue #1733, opened 2021-05-17
First beta release coming up


dereuromark:
Hi
Especially to the ones actively contributing:
It is time we start to make a new release happen.

What do you think we should include in our first beta release?
We should probably
- Identify all BC breaking issues and label them as such ( https://github.com/propelorm/Propel2/labels/BC-break%202.0 )
- Identify all critical and important issues to include in such a beta release - and add them to the milestone ( https://github.com/propelorm/Propel2/milestone/4 )
- If some can be descoped as requirement we can also do that in order to have a realistic release soon

Maybe you can help me here?
And we can get a milestone planned, including release date?

mringler:
Oh, exiting!

Some thoughts on the milestones:
- #606 (perform migrations through php instead of SQL): That's a fun one, but a ton of work. It says "max 1 day of work" in the description, and I seriously doubt that. Also, it probably makes it harder to see and fix errors in generated migrations. Unless someone jumps at it, I don't see this happening anytime soon (or at all).
- #1577 (modern MySQL adapter): I think this is a misunderstanding and can be closed. I added a comment there with my reasoning.
- #687 (migrations with multiple connections): Another fun one. Problem is that the migration command files really were not written to be rewritten. Hard to test, too. Again, a lot of work. More likely than #606, and an actual bug, but probably not going to happen anytime soon.

Personally, I would like to see Propel 2 get out of dev-limbo, so descoping the open issues makes sense to me.

Of my PRs, I think having `EXISTS` queries adds credibility to the major version bump, and really shouldn't be missing in any ORM, considering how important it is in search queries.
And I would include all the bugfixes, of course, since almost all of them fix performance issues, or, even worse, problems that make it hard or joyless to work with propel. Particularly the changes around nested queries are useful, since that feature was pretty much unusable, when you ran into a bug whichever way you wrote it.
But that is obviously up to you to decide.

Hope that helps, best
Moritz

dereuromark:
I updated your post and added bullet points to toggle once done.
Can you please check 1st item? The ticket seems to be a different one than you meant.
Also, for the 2nd part, can you maybe also link the others in the same style, you can just edit your post.

This way, once we see all are done, we can close this ticket later a bit more easily.

SourceCode:
Migrations in PHP seems like a really poor idea.  Why and what would make a migration depart or need to depart from the native language of the storage engine you are using.

dereuromark:
[...]

PR #1765, opened 2021-06-25
Try PHP 8.1

dereuromark:
Part of https://github.com/propelorm/Propel2/issues/1733

Evaluate propel compat with 8.1 dev/alpha

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,9 +14,9 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php-version: [ '7.2', '8.1' ]
-                db-type: [ sqlite, mysql, pgsql, agnostic ]
-                symfony-version: [ '3-min', '3-max', '4-min', '4-max', '5-min', '5-max' ]
+                php-version: ['8.0', '8.1']
+                db-type: [sqlite, mysql, pgsql, agnostic]
+                symfony-version: ['3-min', '3-max', '4-min', '4-max', '5-min', '5-max']
                 exclude:
                   - php-version: '8.0'
                     symfony-version: '4-min'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php-version: [ '7.2', '7.4', '8.0' ]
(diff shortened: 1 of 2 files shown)


dereuromark:
Good news: It looks like propel works with 8.1 just fine
Only one weird issue:
https://github.com/propelorm/Propel2/pull/1765/checks?check_run_id=2913036548
Mysql has a fail:
```
There was 1 failure:

1) Propel\Tests\Runtime\ActiveQuery\ModelCriteriaTest::testHavingWithColumn
having() accepts a string clause
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     0 => Array (
         'table' => null
-        'type' => 2
+        'type' => 3
         'value' => 'foo'
     )
 )

/home/runner/work/Propel2/Propel2/tests/Propel/Tests/Runtime/ActiveQuery/ModelCriteriaTest.php:58
/home/runner/work/Propel2/Propel2/tests/Propel/Tests/Runtime/ActiveQuery/ModelCriteriaTest.php:588
```
The other dbs dont

dereuromark:
Can anyone explain why 8.1 vs 8.0 changes the functional outcome of a simple query?
The language itself shouldn't be able to cause this, should it?

mringler:
Yes, this is really weird.
The test case is defined to only run on MySQL, but it is probably independent of DB vendor. The test sets the type of a having clause to string:
```
        $c->having('title_start = ?', 'foo', PDO::PARAM_STR);
```
And then tests if the resulting param declaration has "type" set to 2 (which is the value of `PDO::PARAM_STR`):
```
$(expected)params = [
            ['table' => null, 'type' => 2, 'value' => 'foo'],
        ];
```
I am pretty sure that the type value is only propagated, not evaluated or changed. A value of 3 indicates `PARAM_LOB` which would not make sense anyway.
So the easiest, yet unlikely reason is that PDO in PHP 8.1 changed the value to 3. I don't have PHP 8.1 yet, so I cannot check it more thoroughly.
What happens if you change the expected return type to `PDO::PARAM_STR`, i.e. in `ModelCriteriaTest::testHavingWithColumn()`:
```
$params = [
            ['table' => null, 'type' => PDO::PARAM_STR, 'value' => 'foo'],
        ];
```
?

PR #1766, opened 2021-07-12
Add phpstan- prefixed ObjectCollection&\Traversable<> annotations

nederdirk:
In the query baseclass generator, some `@psalm-method` lines were used,
because PHPstan doesn't support prefixed `@method`-annotations currently
(2021-07-12). The two tools have a good history of supporting each
other's prefixes, so chances are that PHPstan will pick up the `@psalm-` prefixed
annotations in the future.

### DIFF
--- a/src/Propel/Generator/Builder/Om/ObjectBuilder.php
+++ b/src/Propel/Generator/Builder/Om/ObjectBuilder.php
@@ -4340,6 +4340,7 @@ abstract class " . $this->getUnqualifiedClassName() . $parentClass . ' implement
      * @param      ConnectionInterface \$con optional connection object
      * @param      string \$joinBehavior optional join type to use (defaults to $joinBehavior)
      * @return ObjectCollection|{$className}[] List of $className objects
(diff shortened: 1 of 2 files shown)


src/Propel/Generator/Builder/Om/ObjectBuilder.php
```diff
@@ -5072,6 +5078,7 @@ protected function addRefFkScheduledForDeletionAttribute(&$script, ForeignKey $r
     /**
      * An array of objects scheduled for deletion.
      * @var ObjectCollection|{$className}[]
+     * @phpstan-var  ObjectCollection&\Traversable<{$className}>
```
dereuromark, line 5081: No inline alignment needed
```suggestion
     * @phpstan-var ObjectCollection&\Traversable<{$className}>
```

dereuromark (approved):
Apart from small alignment issues

nederdirk:
> [approved] Apart from small alignment issues

Fixed!

PR #1767, opened 2021-07-14
PropelConfiguration: avoid deprecation notice

gharlan:
After this PR the new way is already used for sf 4.2 - 4.4, to avoid the deprecation notice.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -22,7 +22,6 @@ parameters:
         - '%rootDir%/../../../src/Propel/Generator/Command/InitCommand.php'
     ignoreErrors:
         - '#Call to an undefined method .+Collection::.+Array\(\)#'
-        - '#Class .+TreeBuilder constructor invoked with 0 parameters, 1-3 required#'
         -
         	message: "#^Call to an undefined method Symfony\\\\Component\\\\Config\\\\Definition\\\\Builder\\\\TreeBuilder\\:\\:root\\(\\)\\.$#"
         	path: src/Propel/Common/Config/PropelConfiguration.php
--- a/psalm.xml
+++ b/psalm.xml
@@ -26,5 +26,10 @@
         <ImplementedReturnTypeMismatch errorLevel="suppress"/>
         <ImplicitToStringCast errorLevel="suppress"/>
         <UndefinedMagicMethod errorLevel="suppress"/>
+        <UndefinedMethod>
+            <errorLevel type="suppress">
+                <referencedMethod name="Symfony\Component\Config\Definition\Builder\TreeBuilder::root"/>
+            </errorLevel>
+        </UndefinedMethod>
     </issueHandlers>
 </psalm>
--- a/src/Propel/Common/Config/PropelConfiguration.php
+++ b/src/Propel/Common/Config/PropelConfiguration.php
@@ -27,14 +27,13 @@ class PropelConfiguration implements ConfigurationInterface
      */
     public function getConfigTreeBuilder()
     {
-        $isBeforeSymfony5 = method_exists(TreeBuilder::class, 'root');
+        $treeBuilder = new TreeBuilder('propel');
 
-        if ($isBeforeSymfony5) {
-            $treeBuilder = new TreeBuilder();
-            $rootNode = $treeBuilder->root('propel');
-        } else {
-            $treeBuilder = new TreeBuilder('propel');
+        if (\method_exists($treeBuilder, 'getRootNode')) {
             $rootNode = $treeBuilder->getRootNode();
+        } else {
+            // BC layer for symfony/config 4.1 and older
+            $rootNode = $treeBuilder->root('propel');
         }
 
         $this->addGeneralSection($rootNode);
--- a/src/Propel/Common/Config/PropelConfiguration.php
+++ b/src/Propel/Common/Config/PropelConfiguration.php
@@ -29,7 +29,7 @@ class PropelConfiguration implements ConfigurationInterface
     {
         $treeBuilder = new TreeBuilder('propel');
 
-        if (\method_exists($treeBuilder, 'getRootNode')) {
+        if (method_exists($treeBuilder, 'getRootNode')) {
             $rootNode = $treeBuilder->getRootNode();
         } else {
             // BC layer for symfony/config 4.1 and older


dereuromark (approved):
Small CS issue, otherwise looks good.

gharlan:
> Small CS issue

fixed

src/Propel/Common/Config/PropelConfiguration.php
```diff
@@ -27,14 +27,13 @@ class PropelConfiguration implements ConfigurationInterface
      */
     public function getConfigTreeBuilder()
     {
-        $isBeforeSymfony5 = method_exists(TreeBuilder::class, 'root');
```
mringler, line 30: Honestly, I prefer the documenting variable over the comment you added. Nobody reads comments. Do you mind restoring this as something like
```
$isSymfony5OrLater = `method_exists($treeBuilder, 'getRootNode');
if($isSymfony5OrLater){
    ...
```

src/Propel/Common/Config/PropelConfiguration.php
```diff
@@ -27,14 +27,13 @@ class PropelConfiguration implements ConfigurationInterface
...
-        if ($isBeforeSymfony5) {
-            $treeBuilder = new TreeBuilder();
-            $rootNode = $treeBuilder->root('propel');
-        } else {
-            $treeBuilder = new TreeBuilder('propel');
+        if (method_exists($treeBuilder, 'getRootNode')) {
             $rootNode = $treeBuilder->getRootNode();
+        } else {
+            // BC layer for symfony/config 4.1 and older
+            $rootNode = $treeBuilder->root('propel');
```
mringler, line 36: Doesn't matter too much, but if you are setting the same variable, a ternary if should be more readable:
```
$rootNote = ($isSymfony5OrLater) ? $treeBuilder->getRootNode() : $treeBuilder->root('propel');
```

psalm.xml
```diff
@@ -26,5 +26,10 @@
         <ImplementedReturnTypeMismatch errorLevel="suppress"/>
         <ImplicitToStringCast errorLevel="suppress"/>
         <UndefinedMagicMethod errorLevel="suppress"/>
+        <UndefinedMethod>
+            <errorLevel type="suppress">
+                <referencedMethod name="Symfony\Component\Config\Definition\Builder\TreeBuilder::root"/>
+            </errorLevel>
+        </UndefinedMethod>
```
mringler, line 33: Would you mind explaining why is this necessary?

phpstan.neon
```diff
@@ -22,7 +22,6 @@ parameters:
         - '%rootDir%/../../../src/Propel/Generator/Command/InitCommand.php'
     ignoreErrors:
         - '#Call to an undefined method .+Collection::.+Array\(\)#'
-        - '#Class .+TreeBuilder constructor invoked with 0 parameters, 1-3 required#'
```
mringler, line 25: Would you mind explaining why is this necessary?

mringler (changes requested):
I like the idea, but I am honestly not sure if it really is an improvement. But maybe I don't understand it yet, so please explain for the peanut gallery:
It seems like you are moving the constructor call out of the if, but need two rules for psalm and stan for it to work. Which seems worse than before, since rules have a long tail. Why do you need the rules? Is there a way around it? Would you prefer two rules over a safe declaration?

phpstan.neon
gharlan: because phpstan is complaining that this ignore-rule is not neccessary/used anymore: https://github.com/propelorm/Propel2/runs/3067690860

psalm.xml
gharlan: because psalm is complaining that a method is called that does not exist: https://github.com/propelorm/Propel2/runs/3067726361
(but it exists in older symfony versions)

mringler: But it did not complain before, when it was called in the `if`?
On a semi-related note: where did you see the deprecation notice?

gharlan:
> but need two rules for psalm and stan for it to work

I need a new ignore-rule for pslam, but for phpstan I actually removed one. So it is +1-1=0.

My snippet is the standard way for solving the problem, which is used in many symfony bundles. [[1]](https://github.com/doctrine/DoctrineBundle/blob/1.12.x/DependencyInjection/Configuration.php#L43-L50) [[2]](https://github.com/hautelook/AliceBundle/blob/master/src/DependencyInjection/Configuration.php#L31-L38) [[3]](https://github.com/FriendsOfSymfony/FOSUserBundle/blob/master/DependencyInjection/Configuration.php#L34-L40) ...

The advantage is: already with symfony 4.2 - 4.4 the new way is used, not only with sf 5. This avoids to get deprecation notices when using symfony 4.2 - 4.4:

Bildschirmfoto 2021-07-14 um 19 58 45

Bildschirmfoto 2021-07-14 um 19 59 11

src/Propel/Common/Config/PropelConfiguration.php
gharlan: Yes, I can change the code to using a variable if you prefer it. But the variable would be `$isSymfony42OrLater`.

psalm.xml
gharlan: Instead of changing the psalm config, I could add this to the else branch:

```php
assert(method_exists($treeBuilder, 'root'));
```

src/Propel/Common/Config/PropelConfiguration.php
mringler: Maybe `$isAtLeastSymfony42`? I really think it is much better to use variable names instead of comments, but I am happy to shut up about it if other reviewers dissagree

phpstan.neon
mringler: Right, you removed the rule. I need to check my color perception. Sorry!

mringler (approved):

psalm.xml
mringler: Interesting. But I don't think that's necessary. You are right, swapping the rule from stan to psalm seems alright.

src/Propel/Common/Config/PropelConfiguration.php
dereuromark: Since this will be removed in the future, fair enough I guess.

PR #1769, opened 2021-07-27
Quote primary table name if identifier quoting is enabled

nederdirk:
With the latest `master` build we ran into the problem that `COUNT(*)`-queries to our `order` table failed. This patch ensures that the setting for identifier quoting is properly observed.

Note that I deliberately ignored the `getSql` helper-function in the test, as this behaviour is precisely what the `getSql` helper-function hides.

### DIFF
--- a/src/Propel/Runtime/ActiveQuery/Criteria.php
+++ b/src/Propel/Runtime/ActiveQuery/Criteria.php
@@ -2207,7 +2207,7 @@ class Criteria
         }
 
         if (empty($fromClause) && $this->getPrimaryTableName()) {
-            $fromClause[] = $this->getPrimaryTableName();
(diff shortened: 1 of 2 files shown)


dereuromark (approved):
Was this a regression recently introduced?

nederdirk:
> Was this a regression recently introduced?

I think so, as we didn't encounter the same problem before. However, I'm not entirely sure and I'm not interested in the finger-pointing game :smile_cat:

dereuromark:
All good - if anything we would rather use this to improve our code review skills :) After all, we approved and merged without seeing this.
Anyway, we do need a 2nd signoff on this to merge.

yaroslav-spryker (approved):

Issue #1284, opened 2016-10-14
plural of "ox" words


kriks57:
Hi,

with StandardEnglishPluralizer, when a column name is box, it's plural form is generated as boxen, which is wrong.
In the StandardEnglishPluralizer class "ox" is in $plural and $irregular, but the latter is applied first.

maybe should there be a ^ in $pattern or $searchPattern ?

dereuromark:
Is someone able to make a PR here with suggested changes?

dereuromark:
https://github.com/propelorm/Propel2/pull/1772

PR #1772, opened 2021-08-03
[#1284] pluralise 'Box' to 'Boxes'

nederdirk:
Time to fix some old bug-reports!

Although the [Jargon File](http://www.catb.org/jargon/html/B/boxen.html) suggests that 'Boxen' should be a valid pluralisation 😜


dereuromark (approved):

dereuromark:
You might want to also check lists like https://github.com/cakephp/cakephp/blob/master/tests/TestCase/Utility/InflectorTest.php#L236 against the code base and if there are a few more left overs on this front.

gechetspr (approved):

PR #1776, opened 2021-08-10
Use object-equality instead of reference-equality to compare object properties

nederdirk:
Sometimes we use boxed primitives for types that have more specific semantics than just `string` or `int`. This allows us to define methods on the boxed primitives, enabling greater class coherence.

However, entities with boxed primitives are always reported as 'does not have only default values', because the `hasOnlyDefaultValues` function uses `!==`. Generally this is a good idea, but for objects this compares the _references_ instead of the _contents_.

This patch inspects the default-value initialiser for 'starts with `new `', as a proxy for 'is this an object type'.

There's a test to validate this behaviour. I validated that the test fails with the unpatched `ObjectBuilder` code.


dereuromark (approved):
Looks good to me.

gechetspr (approved):

PR #1779, opened 2021-08-20
1317-parent-constructor

### LINKED ISSUE #1317: Add parent::__construct to addConstructorOpen when BaseClass is defined
When i have a BaseClass in my <table> node in schema.xml the $outputDir/Base/$name.php has the right extends Expression. 

```
/**
 * Initializes internal state of $pathToModel object.
 */
 public function __construct()
 {
 }
```

But why not then there is no parent::__construct() after? Would be nice if the BaseClass is valid and added to the class defination. 

```
/**
 * Initializes internal state of $pathToModel object.
 */
 public function __construct()
 {
   parent::__construct();
 }
```

nederdirk:
Fixes #1317

I have tried to add a test for this in `tests/Propel/Tests/Generator/Behavior/ConcreteInheritance/ConcreteInheritanceBehaviorTest.php`, but because of the on-the-fly generated classes I was not able to override the constructor of `ConcreteContentSetPk`.

### DIFF
--- a/src/Propel/Generator/Builder/Om/ObjectBuilder.php
+++ b/src/Propel/Generator/Builder/Om/ObjectBuilder.php
@@ -218,12 +218,13 @@ class ObjectBuilder extends AbstractObjectBuilder
      *
      * @return string|null
      */
-    protected function getParentClass()
+    protected function getParentClass(): ?string
     {
         $parentClass = $this->getBehaviorContent('parentClass');
         if ($parentClass !== null) {
             return $parentClass;
         }
+
         return ClassTools::classname($this->getBaseClass());
     }
 
@@ -240,7 +241,8 @@ class ObjectBuilder extends AbstractObjectBuilder
         $tableName = $table->getName();
         $tableDesc = $table->getDescription();
 
-        if (($parentClass = $this->getParentClass()) !== null) {
+        $parentClass = $this->getParentClass();
+        if ($parentClass !== null) {
             $parentClass = ' extends ' . $parentClass;
         }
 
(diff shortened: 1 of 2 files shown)


src/Propel/Generator/Builder/Om/ObjectBuilder.php
```diff
@@ -226,10 +240,7 @@ protected function addClassOpen(&$script)
         $tableName = $table->getName();
         $tableDesc = $table->getDescription();
 
-        if (
-            ($parentClass = $this->getBehaviorContent('parentClass')) !== null ||
-            ($parentClass = ClassTools::classname($this->getBaseClass())) !== null
-        ) {
+        if (($parentClass = $this->getParentClass()) !== null) {
```
dereuromark, line 243: Can we avoid inline assignment here?

nederdirk: Sure, working on it!

src/Propel/Generator/Builder/Om/ObjectBuilder.php
```diff
@@ -213,6 +213,20 @@ protected function getDefaultValueString(Column $column)
         return $defaultValue;
     }
 
+    /**
+     * Return the parent class name, or null.
+     *
+     * @return string|null
+     */
+    protected function getParentClass()
```
dereuromark, line 221: Should we add return type for New methods?

dereuromark (approved):

src/Propel/Generator/Builder/Om/ObjectBuilder.php
nederdirk: done!

dereuromark:
Looks like CS is failing

dereuromark:
Any additional reviews here so we can merge this?

yaroslav-spryker (approved):

PR #1782, opened 2021-09-05
Reintroduce late select for backward compatibility

mringler:
Undoes most of the changes from https://github.com/propelorm/Propel2/pull/1711
as requested in https://github.com/propelorm/Propel2/issues/1733#issuecomment-907154887


dereuromark:
Do we still keep some of the fixes and Improvements of the original PR here?

Thank you for the efforts, I think we can release then now and can rediscuss further Improvements for next beta without hurries.

mringler:
The old PR also introduced setting the base table at create time, instead of having to initialize it during query building. That part is still there. Probably should have put it into its own PR anyway.

dereuromark:
I merged it as we will do another QA session tomorrow and finish our release then based on the result.

Issue #972, opened 2015-07-30
ErrorException: Only variable references should be returned by reference


sebastian301082:
We get this Exception while using Propel Collection class.
public function &offsetGet($offset)
Problem is the & at the beginning. It's not supported anymore by newer PHP versions. Could you please look into it?

All the best
Seb

https://github.com/propelorm/Propel2/blob/master/src/Propel/Runtime/Collection/Collection.php#L96

mreichenbach-igniti:
Yea, would be great to have that solved :+1:

BPE02:
+1

bskart:
+1

anusch-athari:
:+1:

Hadrien45:
+1

chocochaos:
Just ran into this too. A fix would be nice :)

PR #1794, opened 2021-10-26
Remove outdated & on method

dereuromark:
Refs https://github.com/propelorm/Propel2/issues/972

Lets see if that breaks something

Result:
```
There was 1 error:

1) Propel\Tests\Runtime\Collection\ArrayCollectionTest::testSave
Indirect modification of overloaded element of Propel\Runtime\Collection\ArrayCollection has no effect

/home/runner/work/Propel2/Propel2/tests/Propel/Tests/Runtime/Collection/ArrayCollectionTest.php:49
```

### DIFF
--- a/src/Propel/Runtime/Collection/Collection.php
+++ b/src/Propel/Runtime/Collection/Collection.php
@@ -99,7 +99,7 @@ class Collection implements ArrayAccess, IteratorAggregate, Countable, Serializa
      *
      * @return mixed
      */
-    public function &offsetGet($offset)
+    public function offsetGet($offset)
     {
         if (isset($this->data[$offset])) {
             return $this->data[$offset];


dereuromark:
Anyone wants to help out here?

dereuromark:
The code should also be fixed up to
```diff
      */
     public function &offsetGet($offset)
     {
-        if (isset($this->data[$offset])) {
-            return $this->data[$offset];
-        }
+        if (!isset($this->data[$offset])) {
+            return null;
+        }
+
+        return $this->data[$offset];
     }
 
     /**
```
But that is only possible if we can make this fix here happen first.

dereuromark:
I guess we can just add

    #[\ReturnTypeWillChange]

PR #1799, opened 2021-11-22
Fix issue propelorm#1345: Corrupt data returned: many-to-many without primary key

javray:


src/Propel/Runtime/ActiveQuery/InstancePoolTrait.php
```diff
@@ -30,8 +30,9 @@ public static function addInstanceToPool($object, $key = null)
             if ($key === null) {
                 $key = static::getInstanceKey($object);
             }
-
-            self::$instances[$key] = $object;
+            if (!empty($key)) {
```
dereuromark, line 33: So this is an empty string here then? as we do check for null before.
I wonder if the code calling this would check for proper valid "non empty string" instead, and doesnt call this method then in the first place?

Making this a no-op method in such an edge case might make it more difficult to debug in the future.

javray: I think it is the simplest solution, other solutions will involve adding more complexity to the development for the rare case that a table does not have a primary key.

dereuromark, line 33:
```php
            if (!Propel::isInstancePoolingEnabled()) {
                return;            
            }

            if (!$key) {
                return;
            }
            
            ...
```
Can we use bail early principle and noncloaking instead then?

javray: Done

mringler (approved):

PR #1800, opened 2021-11-23
use bail early principle and noncloaking

javray:


src/Propel/Runtime/ActiveQuery/InstancePoolTrait.php
```diff
@@ -26,13 +26,17 @@ trait InstancePoolTrait
...
-            }
-
-            self::$instances[$key] = $object;
+        if (!Propel::isInstancePoolingEnabled()) {
+            return;
+        }
+        if ($key === null) {
+            $key = static::getInstanceKey($object);
+        }
+        if (empty($key)) {
```
dereuromark, line 35: It is still cloaking check

PR #1804, opened 2021-12-03
Add return-typehint for generated buildRelations, to bring it in line with the base class typehint

nederdirk:
Our 'symfony debug autoloader' was spamming messages for every generated `TableMap`-class on every request. This patch should bring the typehint of the generated `TableMap` implementations in line with the typehint of the base class

### DIFF
--- a/src/Propel/Generator/Builder/Om/TableMapBuilder.php
+++ b/src/Propel/Generator/Builder/Om/TableMapBuilder.php
@@ -618,6 +618,8 @@ class " . $this->getUnqualifiedClassName() . " extends TableMap
         $script .= "
     /**
      * Build the RelationMap objects for this table relationships
+     *
+     * @return void
      */
     public function buildRelations()
     {";


dereuromark (approved):
I wonder how this got missed

Do we have any other methods with missing return annotations etc?

nederdirk:
> I wonder how this got missed
> 
> Do we have any other methods with missing return annotations etc?

None that I have noticed, but it's hard to tell with certainty. To be honest, I think the symfony debug autoloader is overly strict in this case, as it gives the reason 'the library you're using _might_ in the future declare the return type as `: void` instead of `@return void`'

Also: since we're targeting php7.3+, we might as well specify `: void` instead of `@return void` here, right? that's compatible, see here:
```
>>> class A { public function a() {} }
>>> class B extends A { public function a(): void {} }
>>> $b = new B;
=> B {#2672}
>>> $b->a()
=> null
```

dereuromark:
Yes, we should actually add consistent param/return types now everywhere possible for the next beta release.
So we can actually have a more safe upgrade process from there on in the future - once all types are in place, no more fatal errors because suddenly the lib added types where there were none in the future.

PR #1805, opened 2021-12-05
PHP ^7.4 Compatibility (Including 8.1)

### LINKED ISSUE #1795: PHP 8.0
./vendor/bin/propel config:convert --config-dir /var/customers/webs/xxx/generated-conf/ --output-dir /tmp/
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Propel\Generator\Command\ConfigConvertCommand::createLoadDatabaseDummyScript() must be of the type string, null given, called in /var/customers/webs/aktuell/xxx/vendor/propel/propel/src/Propel/Generator/Command/ConfigConvertCommand.php on line 60 and defined in /var/customers/webs/aktuell/xxx/vendor/propel/propel/src/Propel/Generator/Command/ConfigConvertCommand.php:115

oojacoboo:
This PR includes support for PHP >= 7.4, resolving all type errors and deprecation warnings.  

I should further add that PHP 7.3 is no longer receiving security updates - it's unmaintained.  And, 7.4 will be unmaintained next year.

It appears evident that there is a lot of nullable behavior tied to business logic throughout the lib.  It's quite unfortunate.  I was able to strictly type and enforce in many places.  However, in others, there were some business logic decisions that had to be made.  And without greater knowledge of how some of these classes are instantiated and used, these decisions were typically permissive.

I’ve gone ahead and conformed with your code-sniffer rules @dereuromark.  I will say though, I disagree with the requirement to include a docblock with PHP 8.1.  Going forward, now that we have support to fully type a function, they are self documenting.  If a function doesn’t require additional comments, for the function - itself, it’s arguments, return value, throws, etc - a docblock shouldn’t be required.

This PR actually took significantly more time than I expected.  There were a lot of little things, especially to get all the tests passing.

The docs need updating to make clear only PHP 8.0 and greater is supported.  I didn’t see them in the repo.  It’d be nice to have the docs in the repo with a static site generator.  I’ve used docusaurus on other OSS libs.

Closes #1795


dereuromark:
Isn't it a bit too early for this?
Propel was always a bit more on the conservative side of PHP dependency to allow also older projects to still be compatible with latest release.
So at this time, we are PHP7.3+ but we could discuss to at least keep the latest minor of 7.x, so PHP7.4+ maybe.
One more release at least here, and in a few months we could still discuss 8.0+ then?

phpstan.neon
```diff
@@ -6,7 +6,7 @@ parameters:
     checkMissingIterableValueType: false
     paths:
         - '%rootDir%/../../../src/'
-    excludes_analyse:
+    excludePaths:
```
dereuromark, line 9: You might have missed https://github.com/propelorm/Propel2/pull/1803 PR and you did a lot of redundant work
Might have been easier to build on top of existing work done (which now is merged anyway).

src/Propel/Common/Config/Loader/IniFileLoader.php
```diff
@@ -80,6 +80,7 @@ private function parse(array $data): array
 
         foreach ($data as $section => $value) {
             if (is_array($value)) {
+                // @phpstan-ignore-next-line
```
dereuromark, line 83: We solved this cleaner in current master

oojacoboo:
> Isn't it a bit too early for this? Propel was always a bit more on the conservative side of PHP dependency to allow also older projects to still be compatible with latest release. So at this time, we are PHP7.3+ but we could discuss to at least keep the latest minor of 7.x, so PHP7.4+ maybe. One more release at least here, and in a few months we could still discuss 8.0+ then?

@dereuromark I have no problem with continuing to target ^7.4.  Do you have suggestions on how that's going to be achieved though?  I outlined the issue in the original comment:

>Firstly, the `mixed` type isn’t supported in versions prior to 8.0, and 8.1 has strictly typed interfaces that require the `mixed` return type. We’d have to implement a custom namespaced shimmed interface instead of the core interfaces, update any logic that is typed or checking for instances of this interface, and then hope there aren’t any other side-effects in user-land from reliance on those interfaces.

dereuromark:
We will not be able to add any such types, thats fine though.
We can do that once we cross that bridge, it should be still before declaring a stable version so that we can actually add this in semver compliance.

oojacoboo:
@dereuromark here is an example.  Maybe I'm missing something here.

https://www.php.net/manual/en/class.iterator.php

As you'll notice, this interface for PHP 8.1 requires a `mixed` return type.  In order for classes that implement this interface to be compatible, they must also implement this return type.  However, the `mixed` return type wasn't introduced until 8.0.

phpstan.neon
oojacoboo: Sorry, I didn't see this PR.  Let me know if there is anything we need to do to merge these together.

dereuromark:
Yes, but since we do 7.4+, you can just omit impossible types.

src/Propel/Common/Config/Loader/IniFileLoader.php
oojacoboo: I'm not sure I understand.  What has been solved in `master`?  I just had a look and it's the same.  The only change here is skipping over the PHPStan error, since it's incorrectly determining the string to be empty here:

`$sections = explode($this->nestSeparator, $section);`

dereuromark: No, it does not skip over it.
It uses the proper annotation on it - https://github.com/propelorm/Propel2/blob/master/src/Propel/Common/Config/Loader/IniFileLoader.php#L27

oojacoboo:
@dereuromark when you say use attributes, are you suggesting an attribute or an annotation?  I haven't used attributes yet, as we're just upgrading to PHP 8, ourselves.  I'm not aware of how you'd use an attribute to declare a return typing for a method.  Do you have any info on that?

As for an annotation, no - that won't work b/c PHP doesn't parse annotations, they're just comments.

dereuromark:
See [this](https://github.com/cakephp/cakephp/pull/15708/files)
Only that it should be using FQCN, so `#[\ReturnTypeWillChange]`

oojacoboo:
@dereuromark I see - thanks.  I'll give it a go, targeting ^7.4 for this PR.  Is there anything else preventing this from being merged in?

src/Propel/Common/Config/Loader/IniFileLoader.php
```diff
@@ -151,6 +152,7 @@ private function parseKey(string $key, $rawValue, array &$config): void
                 $value = (float)$rawValue;
             }
         }
+        // @phpstan-ignore-next-line
```
dereuromark, line 155: Lets remove all of those, we never need to use them.

phpstan.neon
dereuromark: You might want to rebase on top of master now.

src/Propel/Common/Config/Loader/IniFileLoader.php
oojacoboo: There are a number of cases where `static` is being used to access a private or protected method.  I'm assuming it's safe to assume `self` is sufficient?

oojacoboo:
@dereuromark Let me know if anything else is needed.

src/Propel/Common/Config/Loader/IniFileLoader.php
dereuromark: Or we change to protected, depending on the case.

src/Propel/Common/Config/Loader/IniFileLoader.php
```diff
@@ -28,7 +28,7 @@ class IniFileLoader extends FileLoader
      *
      * @var string
      */
-    private $nestSeparator = '.';
+    private string $nestSeparator = '.';
```
dereuromark, line 31: We didnt add any such types on properties yet. This would be the first.

src/Propel/Generator/Builder/Util/SchemaReader.php
```diff
@@ -460,7 +463,9 @@ private function getLocationDescription()
             $location .= sprintf('file %s,', $this->currentXmlFile);
         }
 
+        /** @psalm-suppress InvalidArgument */
```
dereuromark, line 466: We usually have baselines for this, I am not sure we should start adding ignores and suppress then directly into code.

src/Propel/Generator/Model/Column.php
```diff
@@ -319,12 +319,12 @@ protected function setupObject()
                 $this->parentTable->getAttribute('columnPrefix'),
             );
 
-            // Accessor visibility
-            $visibility = $this->getMethodVisibility('accessorVisibility', 'defaultAccessorVisibility');
+            // Accessor visibility - no idea why this returns null, or the use case for that
+            $visibility = $this->getMethodVisibility('accessorVisibility', 'defaultAccessorVisibility') ?: '';
```
dereuromark, line 323: But is continuing with an empty string better then?
Shouldnt this throw an exception or default to a sensible value instead?

src/Propel/Runtime/Parser/CsvParser.php
```diff
@@ -234,10 +234,10 @@ public function toCSV($array, $isList = false, $includeHeading = true)
      */
     public function toArray($data, $rootKey = null, $isList = false, $includeHeading = true)
     {
-        $rows = explode($this->lineTerminator, $data);
+        $rows = explode($this->lineTerminator, $data); // @phpstan-ignore-line
```
dereuromark, line 237: I am sure those are not needed anymore

tests/Propel/Tests/Generator/Behavior/Validate/UniqueConstraintTest.php
```diff
@@ -29,6 +29,7 @@ public function testUniqueValidatorPass()
         $publisher = new Publisher();
         $publisher->setName('Happy Reading');
         $publisher->setWebsite('http://www.happyreading.com');
+        $publisher->save();
```
dereuromark, line 32: Why does it need to be saved before validating?

.github/workflows/ci.yml
```diff
@@ -5,19 +5,20 @@ on:
...
         runs-on: ubuntu-18.04
         strategy:
             fail-fast: false
             matrix:
-                php-version: [ '7.3', '7.4', '8.1' ]
+                php-version: [ '7.4', '8.0', '8.1' ]
                 db-type: [ sqlite, mysql, pgsql, agnostic ]
                 symfony-version: [ '4-min', '4-max', '5-min', '5-max' ]
                 exclude:
-                  - php-version: '8.1'
```
dereuromark, line 19: Why removing 8.1 here?

phpstan-baseline.neon
```diff
@@ -1,4 +1,5 @@
 parameters:
+	reportUnmatchedIgnoredErrors: false
```
dereuromark, line 2: this file should only be touched by tools
Are you sure this shouldnt be in the main neon file?

src/Propel/Common/Config/Loader/IniFileLoader.php
dereuromark: I am not against it, just outlining that it can have risks.
As a first step the public API (methods that are non private) should be the focus.

oojacoboo: I'll remove it if you want, but I don't see the issue in a gradual approach to more strict typing.

src/Propel/Generator/Builder/Util/SchemaReader.php
oojacoboo: This may not be required anymore, but there is a typing conflict between 7.4 and 8.0.  These functions require a `resource` and not an instance of `\XMLParser`.  I believe, now that `$parser` is annotated as mixed it's probably okay to remove.

src/Propel/Generator/Model/Column.php
oojacoboo: I'm all for throwing Exceptions.  But, this lib has a ton of business logic around null values.  I don't know if it's the tests, or these values are encountered during normal model generation, or what.  In many cases, null is being passed to a core string function that converts it to an empty string.  However, with PHP 8, that's not allowed, as there are more strict type checks.  In this case, we're just passing an empty string into `strtolower` instead of `null`.  But there are quite a few cases like this through the codebase.

The best solution here is to type our internal methods/properties.  This will prevent most of these issues and start to narrow down the actual source of the issue.

.github/workflows/ci.yml
oojacoboo: Because we're not excluding PHP 8.1 anymore?

tests/Propel/Tests/Generator/Behavior/Validate/UniqueConstraintTest.php
oojacoboo: It was failing before saving, and other examples all saved, so I assumed so.

.github/workflows/ci.yml
dereuromark: Thats a matrix
So you are now excluding all 4-min runs, not just the one with 8.1

src/Propel/Generator/Model/Column.php
dereuromark: True
But then I would rather have string|null here, what do you think?
As checking against null vs non empty string sounds a bit cleaner from the API level.

src/Propel/Common/Config/Loader/IniFileLoader.php
dereuromark: I think the ones that do not default are the problem, as they need to be written first, before they can be accessed.
So I guess this is fine here.

tests/Propel/Tests/Generator/Behavior/Validate/UniqueConstraintTest.php
dereuromark: After saving I wouldnt expect validation to do anything, so we might actually "cloak" the issue here this way.
It was working before, we should find out how it will work again now without the save() part.

src/Propel/Generator/Model/Column.php
oojacoboo: I'm not really sure I'm following you.  `$visibility` ends up `null` in some cases.  It's then passed to `strtolower` which results in a type error on PHP 8.  Before PHP8, that'd just get converted to an empty string.

tests/Propel/Tests/Generator/Behavior/Validate/UniqueConstraintTest.php
oojacoboo:
```
Propel\Tests\Generator\Behavior\Validate\UniqueConstraintTest::testUniqueValidatorPass
Failed asserting that false is true.
```

Without the save, we get the above. I don't know what's failing validation or how validation works.  Additionally, it fails on `master` branch as well.  So, I'm pretty sure the problem is unrelated to this PR.

dereuromark: Interesting..

src/Propel/Generator/Model/Column.php
dereuromark: Fair enough

dereuromark:
@mringler What do you think?
I think the changes look good so far and show a general improvement towards the type safe future we would like to see.
A good first step - of many to come.

composer.json
```diff
@@ -16,25 +16,25 @@
...
         "symfony/yaml": "^3.4.0||^4.0.0||^5.0.0",
         "symfony/config": "^3.4.0||^4.0.0||^5.0.0",
         "symfony/console": "^3.4.0||^4.0.0||^5.0.0",
         "symfony/filesystem": "^3.4.0||^4.0.0||^5.0.0",
         "symfony/finder": "^3.4.0||^4.0.0||^5.0.0",
         "symfony/translation": "^3.4.0||^4.0.0||^5.0.0",
         "symfony/validator": "^3.4.0||^4.0.0||^5.0.0",
-        "symfony/polyfill-php73": "^1.8",
-        "psr/log": "^1.0"
+        "symfony/polyfill-php73": "^1.8"
```
gechetspr, line 28: I assume we don't need it for PHP7.4

src/Propel/Common/Config/XmlToArrayConverter.php
```diff
@@ -87,7 +87,7 @@ protected static function simpleXmlToArray(SimpleXMLElement $xml): array
 
             // if it's not an array, then it was empty, thus a value/string
             if ($child === []) {
-                $child = static::getConvertedXmlValue($v);
+                $child = self::getConvertedXmlValue($v);
```
gechetspr, line 90: why do we need this change?

src/Propel/Generator/Config/GeneratorConfig.php
```diff
@@ -280,9 +280,9 @@ public function getBehaviorLocator()
      *
      * @return object
      */
-    private function getInstance($className, $arguments = null, $interfaceName = null)
+    private function getInstance(?string $className, $arguments = null, $interfaceName = null)
     {
-        if (!class_exists($className)) {
+        if (!$className || !class_exists($className)) {
```
gechetspr, line 285: mb we can change exception message accordingly?

src/Propel/Common/Config/XmlToArrayConverter.php
dereuromark: Private prop

oojacoboo: One of the static analyzers complains about using static on a private or protected method.  @dereuromark and I already discussed this.

src/Propel/Generator/Config/GeneratorConfig.php
oojacoboo: I've fixed it properly now - 26b0cbe30dedbf5472d0761c1e40dc0a946447e5

dereuromark:
Any second review please? So we can move forward?

gechetspr (approved):

PR #1806, opened 2021-12-06
Fix up typos.

dereuromark:
Time to fix some typos for the next beta release.

We wil have to note the small BC topic in the changelogs:
- public function setSlowThreshold($slowThreshold)

### DIFF
--- a/src/Propel/Generator/Command/AbstractCommand.php
+++ b/src/Propel/Generator/Command/AbstractCommand.php
@@ -136,7 +136,7 @@ abstract class AbstractCommand extends Command
     }
 
     /**
-     * Parses a connection string and returns an array with name, DNS and extra information.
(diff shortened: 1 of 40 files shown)


mringler (approved):

src/Propel/Generator/Command/AbstractCommand.php
```diff
@@ -136,7 +136,7 @@ protected function createDirectory($directory)
     }
 
     /**
-     * Parse a connection string and return an array with name, dsn and extra informations
+     * Parses a connection string and returns an array with name, DNS and extra information.
```
a-sabaa, line 139: I don't think this is DNS, it's actually DSN (data source name).

a-sabaa (changes requested):

a-sabaa (approved):

PR #1814, opened 2021-12-09
Added `isSelfSelected` property to the `clear` method

herasimenko1987:
Added the `isSelfSelected` property to the `Propel\Runtime\ActiveQuery\ModelCriteria::clear()` method to avoid an empty select SQL statement after the `clear()` method was executed.

### DIFF
--- a/src/Propel/Runtime/ActiveQuery/ModelCriteria.php
+++ b/src/Propel/Runtime/ActiveQuery/ModelCriteria.php
@@ -994,6 +994,7 @@ class ModelCriteria extends BaseModelCriteria
         $this->primaryCriteria = null;
         $this->formatter = null;
         $this->select = null;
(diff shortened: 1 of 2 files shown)


dereuromark (approved):

dereuromark:
@mringler what do you think?
We might have added that regression in the beta release or so ( https://github.com/propelorm/Propel2/commit/3caa4bf1d67495fc3ca56fb272b019ba3c51fac4 )

mringler:
@dereuromark That's possible, yes. I think I just reset it to the previous state, but who knows.

@herasimenko1987 It would be great if you could add a test for this. But either way, thank you for finding and fixing it!

mringler (approved):

herasimenko1987:
@mringler, will add a test next week.

PR #1815, opened 2021-12-23
Allow Symfony 6

fbourigault:
This allows Symfony 6 components.

This is a draft ATM because I only ran some tests locally.

CI failure seems unrelated. See https://github.com/composer/composer/issues/10387.


fbourigault:
Now psalm job is fixed, I get 11 errors related to added return types in Symfony 6.

I checked those errors and are all enforcing the 5.x behavior.

Should I add those errors to the baseline to keep the current behavior?

dereuromark:
I guess so.
They are not functionally relevant, are they?

In a next step we need to get rid of at least v3 here.
That might also simplify a few things.

composer.json
```diff
@@ -18,13 +18,13 @@
...
-        "symfony/finder": "^3.4.0||^4.0.0||^5.0.0",
-        "symfony/translation": "^3.4.0||^4.0.0||^5.0.0",
-        "symfony/validator": "^3.4.0||^4.0.0||^5.0.0"
+        "symfony/yaml": "^3.4.0||^4.0.0||^5.0.0||^6.0.0",
+        "symfony/config": "^3.4.0||^4.0.0||^5.0.0||^6.0.0",
+        "symfony/console": "^3.4.0||^4.0.0||^5.0.0||^6.0.0",
+        "symfony/filesystem": "^3.4.0||^4.0.0||^5.0.0||^6.0.0",
+        "symfony/finder": "^3.4.0||^4.0.0||^5.0.0||^6.0.0",
+        "symfony/translation": "^3.4.0||^4.0.0||^5.0.0||^6.0.0",
+        "symfony/validator": "^3.4.0||^4.0.0||^5.0.0||^6.0.0"
```
dereuromark, line 27: We are already not testing v3, so you can just as well remove that now.
Current CI says 4-6 if I see that correctly.

tests/composer/composer-symfony6-min.json
```diff
@@ -0,0 +1,61 @@
...
+    "require": {
+        "php": ">=8.0",
+        "symfony/yaml": "~6.0.0",
+        "symfony/config": "~6.0.0",
+        "symfony/console": "~6.0.0",
+        "symfony/filesystem": "~6.0.0",
+        "symfony/finder": "~6.0.0",
+        "symfony/translation": "~6.0.0",
+        "symfony/validator": "~6.0.0",
+        "symfony/polyfill-php73": "^1.8",
```
floriankraemer, line 27: Do we still need that polyfill? We use >=8.0.

composer.json
fbourigault: I opened another PR to remove Symfony <4.4 support (see #1816).

tests/composer/composer-symfony6-max.json
```diff
@@ -0,0 +1,61 @@
...
+    "require": {
+        "php": ">=8.0",
+        "symfony/yaml": "^6.0.0",
+        "symfony/config": "^6.0.0",
+        "symfony/console": "^6.0.0",
+        "symfony/filesystem": "^6.0.0",
+        "symfony/finder": "^6.0.0",
+        "symfony/translation": "^6.0.0",
+        "symfony/validator": "^6.0.0",
+        "symfony/polyfill-php73": "^1.8",
```
floriankraemer, line 27: Do we still need that polyfill? We use >=8.0.

tests/composer/composer-symfony6-min.json
```diff
@@ -0,0 +1,61 @@
...
+        "psr/log": "^1.0"
+    },
+    "require-dev": {
+        "ext-pdo": "*",
+        "ext-json": "*",
+        "monolog/monolog": "^1.3",
+        "phpstan/phpstan": "^0.12.4",
+        "phpunit/phpunit": "^8.0.0||^9.0.0",
```
dereuromark, line 35: It is also unlikely that v8 would even install here, as we are using v9 only now right?

tests/composer/composer-symfony6-max.json
dereuromark: I guess it is easier though to keep the files similar across all versions.
We should remove things everywhere then if possible/useful.

fbourigault: I copy/pasted from other min/max composer.json files in this directory.

I will remove it as the polyfill is now absent from main composer.json.

tests/composer/composer-symfony6-min.json
fbourigault: Will check against main composer.json file.

dereuromark:
Lets focus on getting this merged first, so we can easier then finalize also the v3 removal PR.

dereuromark (approved):

fbourigault:
I did some cleanup in `tests/composer`:

- I removed `composer-php8.json` as it seems unused.
- I fixed php version requirement.
- I removed php 7.3 polyfill.
- I bumped PHPUnit requirement to 9.5+.

PR #1816, opened 2022-01-03
Remove Symfony <4.4 support

fbourigault:
Fix partially #1629

PR #1818, opened 2022-01-04
Remove Symfony <4.4 support

### LINKED ISSUE #1629: Drop EOL Symfony support
Symfony ^4.3.0, ^5.0.0 are EOL July 2020.

fbourigault:
Replace #1816 (I accidentally removed the branch...)

Fix #1629.


src/Propel/Generator/Command/InitCommand.php
```diff
@@ -411,12 +410,8 @@ private function reverseEngineerSchema(OutputInterface $output, array $options)
      */
     protected function createConsoleHelper(InputInterface $input, OutputInterface $output)
     {
-        /* Check if it runs in Symfony3 env — than use QuestionHelper, because DialogHelper is absent */
-        if (class_exists('\Symfony\Component\Console\Helper\QuestionHelper')) {
-            $helper = new ConsoleHelper3($input, $output);
-        } else {
-            $helper = new ConsoleHelper($input, $output);
-        }
+        $helper = new ConsoleHelper3($input, $output);
```
dereuromark, line 413: shouldn't we have the ConsoleHelper used, why "3"?

fbourigault: The ConsoleHelper was here to support Symfony 2.x (it uses DialogHelper which was removed in 3.0).

IMHO we should remove all the ConsoleHelper stuff and use Symfony helpers.

dereuromark (approved):

fbourigault:
I kept the `ConsoleHelper` as it contains some useful logic. However the `ConsoleHelperInterface` is gone!

dereuromark:
Any 2nd reviewer for this maybe?

src/Propel/Generator/Command/Helper/ConsoleHelper.php
```diff
@@ -158,4 +160,19 @@ public function writeln($messages, $options = 0)
...
+     * @param string $question
+     * @param string|null $default
+     *
+     * @return string
+     */
+    private function formatQuestion($question, $default = null)
+    {
+        if ($default) {
+            return sprintf('<info>%s</info> [<comment>%s</comment>]: ', $question, $default);
+        } else {
```
gechetspr, line 174: Not a blocker, but we don't need this else statement. We can just return a string without default after or even we can prepare it beforehand and add default part if default is not null.

gechetspr (approved):
All good except one minor thing that is just a style issue

src/Propel/Generator/Command/Helper/ConsoleHelper.php
dereuromark: The else can indeed be removed, just return afterwards without else.

fbourigault: TBH, I didn't do anything to this class but renaming it from `ConsoleHelper3` to `ConsoleHelper`.

Do you expect I improve it?

dereuromark: Improving is always nice, but no requirement here.

PR #1819, opened 2022-01-06
Remove support for psr/log v2 and v3

oojacoboo:
Since we're still supporting PHP 7.4, versions 2 and 3 of the `psr/log` interface cannot be used.


dereuromark:
Why cannot they be used?
If another dependency needs them for e.g. PHP8+ and the consumer is using a high enough PHP version, why should they be blocked from using it?
Right now it seems always psr/log (1.1.4) is pulled, that is true.
But the question is: Would this library be compatible with the other majors? And if so, then it should keep it.
If not, then it is true that we should remove until we can establish compatibility.

oojacoboo:
@dereuromark we're still trying to support PHP 7.4.  Version 2 of `psr/log` requires >= 8.0

https://packagist.org/packages/psr/log#2.0.0

dereuromark:
I dont see how removing those versions makes the things you need and say incompatible.
V1 is for you, V2/3 is for PHP8 folks that want to use this library in their code.

oojacoboo:
@dereuromark if version 2 or 3 gets installed, this lib will have interface conflicts.  Therefore, it cannot support 2 or 3 because it's not compatible with version 2 or 3.  And, because Composer doesn't support installing multiple versions of a lib, you have to go with whatever the lowest requirement is.

This has nothing to do with us.  We'd like to update our `psr/log` interfaces, but until this lib is compatible with the updated `psr/log` interface, we cannot.

oojacoboo:
IMO, as I stated in the 8.1 compatibility PR, I'd like to drop 7.4 support.  Tag a 7.4 compatibility release and let that be that for 7.4 support.

dereuromark:
Yes, we are 7.4+ until summer of next month or so.
We just dropped 7.0-7.3 recently, and continue to move forward

Again, none of this contradicts using v1-v3 as per PHP major in use. So there is no technical reason yet to go directly to PHP 8.

oojacoboo:
@dereuromark how will this lib work with v2 of `psr/log`, please explain.  What am I missing here?  I don't understand why you closed this PR.  It's fine that we disagree on supporting 7.4, but that shouldn't detract from the validity of this ticket.  This has nothing to do with that discussion.

oojacoboo:
@dereuromark maybe re-review this PR, please.  I think you're confusing things now.  This PR is for supporting PHP 7.4, not removing it - at all.

dereuromark:
I looked into it, using composer why-not you can easily tell why something is not working out.
But instead of removing those, actually enabling those versions to be tested and therefor everified for production usage seems like the better step forward.
See https://github.com/propelorm/Propel2/pull/1822

Issue #1656, opened 2020-09-12
Strict Typing


oojacoboo:
I looked around and couldn't find an issue opened for this.  I'm sure it's been discussed though.  What's the plan for typing the base classes, or offering the option to have strictly typed base classes?  Not having the base classes strictly typed presents interfaces issue for us and I'm sure many others.  PHP has moved much more in this direction and Propel is really holding us back.

dereuromark:
You can make a PR to opt in this behavior.
Sounds useful for many I am sure.

dereuromark:
cc @nederdirk

nederdirk:
I would _love_ this, but I'm not sure if I have time to go ahead and implement this.

dereuromark:
Maybe we can have a task force of several people? Helping out to get this solved for the next milestone release?

oojacoboo:
I have a PHP8 compatibility branch that includes some typing updates to be in compatibility with PHP 8.1's typed interfaces.  What branch should I target for this?

Additionally, I think we should focus on generated code with a priority over core lib code.  The PR I have will cover the core lib for interface compatibility with typing, at least.

nederdirk:
Some thoughts on this:
- lib should definitely target the lowest common denominator, i.e. php7.4 currently
- for generated code it might be nice if the type signatures could be configured, i.e. a user of the Propel2 lib who targets php8.1 might benefit from generated type signatures _with_ proper union types, where appropriate.

Other thoughts:
- in the past I have tried to add phpstan/psalm annotations (primarily related to generics-annotations to Query and Collection classes) which failed, mostly due to the `BaseModelCriteria::$formatter` property. I believe the array-formatter was the biggest problem, in combination with the `addSelect` method silently changing the formatter to `SimpleArrayFormatter`

oojacoboo:
I've submitted PR #1805 which provides support for PHP 8 now.  This is really only a step in the right direction.  But, all tests are passing and everything is running well on ^8.0.

dereuromark:
@oojacoboo I think we should for now keep at least last minor of 7.x as well. There is no good reason to jump too quickly.
We can still add already all major types everywhere and clean up the API
Wherever we use `...|false` the API could be fixed to a cleaner nullable one of `?...`.

dereuromark:
Next PR is https://github.com/propelorm/Propel2/pull/1833
I also need a bit of help for that one.

oojacoboo:
[...]

PR #1821, opened 2022-01-07
Add return types where possible.

dereuromark:
First part of https://github.com/propelorm/Propel2/issues/1656

- [x] Return types where possible
- [x] Necessary template changes for tests to run
- [ ] Generated code also adjusted

We still will have to also check param types as well all generated code for both then.

I am quite confused about the hooks and what kind of return types are expected

ActiveRecordInterface
- preInsert() - why bool
- postInsert() - why ?int
- preUpdate() / preSave()
- postUpdate() / postSave()
- base*() vs *()

Could all be bool? or do we need this `?int` part?

> Tests: 1221, Assertions: 1287, Errors: 703, Failures: 23, Skipped: 13, Incomplete: 4.

Is anyone able to help out by making additional PRs against this one?

### DIFF
--- a/src/Propel/Generator/Model/Schema.php
+++ b/src/Propel/Generator/Model/Schema.php
@@ -103,9 +103,9 @@ class Schema
     /**
      * Returns the generator configuration
      *
-     * @return \Propel\Generator\Config\GeneratorConfigInterface
+     * @return \Propel\Generator\Config\GeneratorConfigInterface|null
      */
-    public function getGeneratorConfig(): GeneratorConfigInterface
+    public function getGeneratorConfig(): ?GeneratorConfigInterface
     {
         return $this->generatorConfig;
     }
@@ -179,9 +179,9 @@ class Schema
      * @param string|null $name
      * @param bool $doFinalInitialization
      *
-     * @return \Propel\Generator\Model\Database
+     * @return \Propel\Generator\Model\Database|null
      */
-    public function getDatabase($name = null, $doFinalInitialization = true): Database
+    public function getDatabase(?string $name = null, bool $doFinalInitialization = true): ?Database
     {
         // this is temporary until we'll have a clean solution
         // for packaging datamodels/requiring schemas
--- a/src/Propel/Generator/Model/Schema.php
+++ b/src/Propel/Generator/Model/Schema.php
@@ -72,7 +72,7 @@ class Schema
      *
      * @return void
      */
-    public function setPlatform(PlatformInterface $platform)
+    public function setPlatform(PlatformInterface $platform): void
     {
         $this->platform = $platform;
     }
@@ -83,7 +83,7 @@ class Schema
      *
      * @return \Propel\Generator\Platform\PlatformInterface
      */
-    public function getPlatform()
+    public function getPlatform(): PlatformInterface
     {
         return $this->platform;
     }
@@ -95,7 +95,7 @@ class Schema
      *
      * @return void
      */
-    public function setGeneratorConfig(GeneratorConfigInterface $generatorConfig)
+    public function setGeneratorConfig(GeneratorConfigInterface $generatorConfig): void
     {
         $this->generatorConfig = $generatorConfig;
     }
@@ -105,7 +105,7 @@ class Schema
      *
      * @return \Propel\Generator\Config\GeneratorConfigInterface
      */
-    public function getGeneratorConfig()
+    public function getGeneratorConfig(): GeneratorConfigInterface
     {
         return $this->generatorConfig;
     }
@@ -117,7 +117,7 @@ class Schema
      *
      * @return void
      */
-    public function setName($name)
+    public function setName($name): void
     {
         $this->name = $name;
     }
@@ -127,7 +127,7 @@ class Schema
      *
      * @return string
      */
-    public function getName()
+    public function getName(): string
     {
         return $this->name;
     }
@@ -137,7 +137,7 @@ class Schema
      *
      * @return string
      */
-    public function getShortName()
+    public function getShortName(): string
     {
         return str_replace('-schema', '', $this->name);
     }
@@ -152,7 +152,7 @@ class Schema
      *
      * @return array<\Propel\Generator\Model\Database>
      */
-    public function getDatabases($doFinalInitialization = true)
+    public function getDatabases(bool $doFinalInitialization = true): array
     {
         // this is temporary until we'll have a clean solution
         // for packaging datamodels/requiring schemas
@@ -168,7 +168,7 @@ class Schema
      *
      * @return bool
      */
-    public function hasMultipleDatabases()
+    public function hasMultipleDatabases(): bool
     {
         return count($this->databases) > 1;
     }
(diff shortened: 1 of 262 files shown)


dereuromark:
Is there anyone available who could lend a hand?
Check the open questions, maybe PR more fixes or missing changes to complete this topic?

We are blocked by this and the PHP 8.1 compatibility for a new beta release right now.

cristianoc72:
I can help here. 
Just two questions:

1.  could I add `declare(strict_types=1);` statement?
2.  could I try to make the code "more strict", in the sense of avoid `null` when possible?

An example of point n. 2. The method https://github.com/propelorm/Propel2/blob/5ebc1943eeedb3d2ae4e4355b0a2d3e512f4cace/src/Propel/Common/Config/ConfigurationManager.php#L56

could be refactored as the following:
```php
public function __construct(string $path = '', array $extraConf = [])
    {
        if ($path === '') {
            $path = getcwd();
        }
        
        $this->config = $this->loadConfig($path, $extraConf);
        $this->process();
    }
```

cristianoc72:
> ActiveRecordInterface
> 
> Could all be bool? or do we need this `?int` part?

`pre*` hooks could return bool. This value is used in the generated `save()` method to check if there are some rows affected by the hooks.

Imho, no need to return something in `post*` methods.

dereuromark:
> Just some questions

1. I would not mix too many things now, we should isolate those and first get it covered/working under current non strict "cast" rules, and only then also look into other strictness
2. avoiding null itself is usually not a big gain, but probably good in some cases. one has to be careful though, as empty string and null in some cases have a different (semantic) meaning. I would also rather see this cleanup a follow up after we have already full type safety under current "rules". We should right now only fix whats absolutely necessary to get our builds, tests and PHPStan happy again. Follow up PRs (small ones) can tackle than any further improvement in "less types as possible"
3. We also dont want to do that at this point. Important is the public API, as we are setting those properties through that signature. So as long as it goes in and out typed, there is big value in typing the properties. Any such change should also rather be a follow up. I think that we already have way too big PRs here to review properly, and we should then limit the side effects that we cannot review further.

cristianoc72:
👍  fine, only minimum changes.

fbourigault:
I'm doing a first PR to fix all agnostics tests.

src/Propel/Runtime/ActiveQuery/ModelCriteria.php
```diff
@@ -540,7 +547,7 @@ public function getSelect()
      *
      * @return \Propel\Runtime\ActiveQuery\Join the previousJoin for this ModelCriteria
      */
-    public function getPreviousJoin()
+    public function getPreviousJoin(): Join
```
fbourigault: Do we agree this should be `?Join` as `useQuery` may not be called yet?

dereuromark: Many of those might have to be nullable as doc blocks where most likely lying.

src/Propel/Runtime/DataFetcher/PDODataFetcher.php
```diff
@@ -74,7 +74,7 @@ public function getStyle()
      *
      * @return array|null
      */
-    public function fetch($style = null)
+    public function fetch($style = null): ?array
```
fbourigault: Many tests fails here:

```
Propel\Runtime\DataFetcher\PDODataFetcher::fetch(): Return value must be of type ?array, bool returned
```

I don't know what to do here.

src/Propel/Generator/Model/Schema.php
```diff
@@ -105,7 +105,7 @@ public function setGeneratorConfig(GeneratorConfigInterface $generatorConfig)
      *
      * @return \Propel\Generator\Config\GeneratorConfigInterface
      */
-    public function getGeneratorConfig()
+    public function getGeneratorConfig(): GeneratorConfigInterface
```
fbourigault, line 108: Many errors comes from here.
```
TypeError: Propel\Generator\Model\Schema::getGeneratorConfig(): Return value must be of type Propel\Generator\Config\GeneratorConfigInterface, null returned
```

What should we do with this one?

src/Propel/Runtime/DataFetcher/PDODataFetcher.php
dereuromark: We should probably remove type for now and declare also bool value return

src/Propel/Generator/Model/Schema.php
dereuromark: Make those nullable, see comments above

fbourigault: So the global strategy for this PR is to find liars Docblocks?

dereuromark: well, they are found as byproduct of getting the fatal errors and failing tests resolved I guess.

dereuromark: in some cases, where they are directly used and expected to be not nullable, it can also make sense to add a has() method to check for existance. then you can has() before get() instead of nullable get() and having to check for return value.
So this is not quite as simple if we want to create a good API moving forward, but the quick way sure would be the nullable fixes.

dereuromark:
> Declaration of Propel\Tests\Helpers\Bookstore\Behavior\TestAuthor::postSave(?Propel\Runtime\Connection\ConnectionInterface $con = NULL): ?int must be compatible with Propel\Tests\Bookstore\Base\Author::postSave(?Propel\Runtime\Connection\ConnectionInterface $con = NULL): void in /home/runner/work/Propel2/Propel2/tests/Propel/Tests/Helpers/Bookstore/Behavior/TestAuthor.php on line 61

Looks like we missed a few return type changes here on void.

dereuromark:
I fixed some of it

Only 198 errors left :P
Anyone up for the task to help out on a few of those?

dereuromark:
I managed to get almost all tests green, but I cannot figure this one out:
```
There were 8 errors:

1) Propel\Tests\Generator\Model\DatabaseTest::testCombineNamespace with data set #0 (null, null, null, 'No namespaces should leave ta... empty')
Propel\Generator\Exception\InvalidArgumentException: No schema name available

...
```

dereuromark:
@mringler @fbourigault What do you think?

fbourigault:
I tried on a real project and was caught by

`Compile Error: Declaration of AppBundle\Model\Customer\Map\UserTableMap::initialize() must be compatible with Propel\Runtime\Map\TableMap::initialize(): void`

The project is a Symfony application with the PropelBundle.

AFAIK, this error is caused because my User class is loaded by the SecurityBundle.

IMHO, we must have a Propel version which add returns types in generated code but not in inherited Propel classes such as `TableMap` and `ModelCriteria` to allow generation of code with return types added.

dereuromark:
We definitely need to adjust the code generators to include full signature, I agree.
Maybe someone can make a PR on top of this?

As for userland code: They are expected to already have return types present as of right now, or at least for the upcoming beta version here.

PR #1822, opened 2022-01-10
Allow Monolog v2 to enable psr/log v2/v3

dereuromark:
When looking into dependencies, it becomes clear that psr/log v1 is required due to a require-dev (optional) dependendy locking it
We need to prevent this in order to fully test both our min and max promise.

As such, this PR aims to fix it properly - and replaces https://github.com/propelorm/Propel2/pull/1819

Completes https://github.com/propelorm/Propel2/pull/1805

With this PHP7 folks are using psr/log v1, PHP8 folks can use psr/log v2/v3 already.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
     "require-dev": {
         "ext-pdo": "*",
         "ext-json": "*",
-        "monolog/monolog": "^1.3",
+        "monolog/monolog": "^1.3 || ^2.3",
         "phpstan/phpstan": "^1.2",
         "phpunit/phpunit": "^9.5.0",
         "spryker/code-sniffer": "^0.17.2",
--- a/src/Propel/Runtime/Connection/ConnectionWrapper.php
+++ b/src/Propel/Runtime/Connection/ConnectionWrapper.php
@@ -657,7 +657,7 @@ class ConnectionWrapper implements ConnectionInterface, LoggerAwareInterface
      *
      * @return void
      */
-    public function setLogger(LoggerInterface $logger)
+    public function setLogger(LoggerInterface $logger): void
     {
         $this->logger = $logger;
     }


oojacoboo:
Is it necessary for this lib to define the `psr/log` interface if the only dependency to this interface exists in a 3rd party lib/dep?

dereuromark:
Probably not, in that case we could move this also into require-dev and unblock people then fully if no such library is used.
I wonder if "any logger" would still be a good requirement by default though.
Is anyone using this lib without a logger in place at all?

oojacoboo:
If the code doesn't require the interface, we shouldn't be having it as a dependency.  The dependencies should be for the direct use in this lib.  I assumed Propel needed that interface for it's own internal logging logic.

If that's not the case, it should be removed entirely, allowing for `monolog/monolog` to manage it's own dependencies, however necessary.

dereuromark:
```php
class ConnectionWrapper implements ConnectionInterface, LoggerAwareInterface
```
so this is quite deeply integrated

Given that I don't think we can move it, though.

With this PR all constraints are still unblocked as everyone can use the correct psr interfaces as per their needs.
So we should be fine.

oojacoboo:
The issue is this:

```php
    /**
     * @param \Psr\Log\LoggerInterface $logger
     *
     * @return void
     */
    public function setLogger(LoggerInterface $logger)
    {
        $this->logger = $logger;
    }
```

https://github.com/php-fig/log/blob/2.0.0/src/LoggerInterface.php

Version 2 there of the `LoggerInterface` uses union types.  That's a PHP 8 feature.  Therefore, version 2 and version 3 of this interface cannot be supported without also supporting a min of PHP 8.

oojacoboo:
This PR is fine though, I agree.  #1819 still needs merging as well.

a-sabaa (approved):

PR #1823, opened 2022-01-12
Use explode() instead of preg_split().

dereuromark:
as per IDE recommendation.


dmytro-dymarchuk (approved):
The title of PR could be changed to `Use explode() instead of preg_split()` but I'm not sure if it is important.

dimitriyTsemma (approved):

PR #1827, opened 2022-02-15
ModelCriteria::getPreviousJoin() may return null

fbourigault:
What should I do with the lying Docblock?


dereuromark:
We should fix the docblocks

PR #1828, opened 2022-02-15
Ensure ConnectionManagerInterface::getName() always return a string

fbourigault:
This is forced using a constructor.

Also, StandardServiceContainer::setManager() no longer have the manager name
as first argument but retrieve it from the manager itself.


src/Propel/Runtime/Connection/ConnectionManagerPrimaryReplica.php
```diff
@@ -45,6 +45,11 @@ class ConnectionManagerPrimaryReplica implements ConnectionManagerInterface
      */
     protected $isForcePrimaryConnection = false;
 
+    public function __construct(string $name)
```
dereuromark, line 48: Make sure to run Code sniffer here So we are green here

PR #1830, opened 2022-02-15
Monolog return type compatibility

fbourigault:
This happen when running tests on PHP 8.1 with monolog 2.

### DIFF
--- a/tests/composer/composer-symfony6-max.json
+++ b/tests/composer/composer-symfony6-max.json
@@ -29,7 +29,7 @@
     "require-dev": {
         "ext-pdo": "*",
         "ext-json": "*",
-        "monolog/monolog": "^1.3",
+        "monolog/monolog": "^2.3",
         "phpstan/phpstan": "^0.12.4",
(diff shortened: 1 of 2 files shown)


tests/composer/composer-symfony6-max.json
```diff
@@ -29,7 +29,7 @@
     "require-dev": {
         "ext-pdo": "*",
         "ext-json": "*",
-        "monolog/monolog": "^1.3",
+        "monolog/monolog": "^2.3",
```
dereuromark, line 32: Is that an outdated branch?
https://github.com/propelorm/Propel2/blob/a2a4a39155bb1e27548a3e38f4f687f59a14257e/composer.json#L32
We should be able to keep both majors if we add the type

fbourigault: It's a tests related composer.json. Those are quite outdated for dependencies which doesn't come from Symfony.

Probably those composer.json files should be replaced by something which use `symfony/flex` (for the `extra.symfony.require` flag and some `composer up --prefer-lower`).

PR #1833, opened 2022-03-09
Add param strictness where possible. 

dereuromark:
rebased https://github.com/propelorm/Propel2/pull/1831

WIP


oojacoboo:
@dereuromark without digging in and taking a stab out if it, it sounds like the hydrator is mis-calculating the number of columns for a given model.  This is one of the things I hate about AR/Propel, the strict ties to the number of columns defined and even the ordering I believe causes issues in places.

My guess is that some typing is causing a column to not be included as part of hydration.

oojacoboo:
@dereuromark haven't looked into the code, but this one is related to the number of columns in the select not matching up with the table.  This is a common problem with Propel due to its design.  Not sure why the typing caused this, but I guess the query builder isn't properly interpreting types.

dereuromark:
For some reason it affects only PHP81 builds

oojacoboo:
For `StatementWrapper::execute` we should drop support for `$inputParameters` being nullable and work our way backwards from that.  The issue appears to be in the execution of the prepared statement there.  I'm assuming that `args` and `$inputParameters` are an issue.  I don't know why without debugging it at runtime.  But tightening up the typing will help pinpoint the issue.

dereuromark:
Well, that didnt really work, broke all builds with really unclear messages.
It seems to be quite cloaked, what the actual issue is.

oojacoboo:
Well, that would, likely break a lot - anything explicitly passing in null to `execute`.  The default value would become an empty array though for all the cases where there aren't any arguments.  But that's the point, to ensure that they're properly passing in an array.  A null value being passed in is prime for errors in unforeseen ways.

What's in your SQL logs from Propel?  `callUserFunctionWithLogging` logs the error.  You need to find out why there is a mismatch in the `args` being passed

gechetspr (changes requested):

src/Propel/Generator/Builder/Om/AbstractOMBuilder.php
```diff
@@ -9,6 +9,7 @@
 namespace Propel\Generator\Builder\Om;
 
 use Propel\Generator\Builder\DataModelBuilder;
+use Propel\Generator\Builder\Om\AbstractOMBuilder as OmAbstractOMBuilder;
```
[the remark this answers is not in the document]
dereuromark: Might have been done by code sniffer for some reason

src/Propel/Generator/Builder/Om/AbstractOMBuilder.php
```diff
@@ -280,7 +281,7 @@ public function getClassNameFromBuilder($builder, bool $fqcn = false): string
      *
      * @return string
      */
-    public function getClassNameFromTable(Table $table, $fqcn = false): string
+    public function getClassNameFromTable(Table $table, bool $fqcn = false): string
```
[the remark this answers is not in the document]
dereuromark: Should we remove it then?

src/Propel/Generator/Builder/Util/SchemaReader.php
```diff
@@ -429,7 +429,7 @@ public function startElement($parser, $tagName, $attributes): void
      *
      * @return void
      */
-    protected function throwInvalidTagException($tag_name): void
+    protected function throwInvalidTagException(string $tag_name): void
```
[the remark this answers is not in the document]
dereuromark: Sure, but I would like to keep such additional (cs) refactorings out of this already big one and in follow up changes.
For now the prio must be to get CI green for merge.

src/Propel/Generator/Model/Behavior.php
```diff
@@ -255,7 +255,7 @@ public function getParameter($name)
      *
      * @return void
      */
-    public function setTableModificationOrder($tableModificationOrder): void
+    public function setTableModificationOrder(int $tableModificationOrder): void
     {
         $this->tableModificationOrder = (int)$tableModificationOrder;
```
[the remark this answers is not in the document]
dereuromark: THis can be removed in cleanup PRs afterwards (phpstan/psalm can find and remove those)

src/Propel/Generator/Model/Column.php
```diff
@@ -1697,7 +1701,7 @@ public static function generatePhpName($name, $phpNamingMethod = PhpNameGenerato
      *
      * @return string
      */
-    public static function generatePhpSingularName($phpname): string
+    public static function generatePhpSingularName(string $phpname): string
```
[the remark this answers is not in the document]
dereuromark: Same, afterwards

src/Propel/Runtime/Collection/ObjectCollection.php
```diff
@@ -347,8 +349,11 @@ public function getColumnValues($columnName = 'PrimaryKey'): array
      *
      * @return self The list of related objects
      */
-    public function populateRelation($relation, $criteria = null, $con = null): self
-    {
+    public function populateRelation(
+        string $relation,
+        ?Criteria $criteria = null,
+        ?ConnectionInterface $con = null
+    ): self {
```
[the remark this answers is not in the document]
dereuromark: I checked, more correct might be static, but thats PHP8+
So until then, we can either keep `self` or none.

dereuromark:
@oojacoboo After https://github.com/propelorm/Propel2/pull/1841 merged in, we could then merge to master.
Does all look good now?

src/Propel/Runtime/Collection/ObjectCollection.php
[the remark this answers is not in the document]
dereuromark: I added static as docblock and ommited any type for now.

src/Propel/Generator/Builder/Om/AbstractOMBuilder.php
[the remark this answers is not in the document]
dereuromark: Done

src/Propel/Generator/Model/Column.php
```diff
@@ -595,7 +595,7 @@ public function getPhpSingularName(): ?string
      *
      * @return void
      */
-    public function setPhpName($phpName = null): void
+    public function setPhpName(?string $phpName = null): void
```
oojacoboo, line 598: We should remove the default `null` assignment on setters.

src/Propel/Generator/Model/Column.php
```diff
@@ -615,7 +615,7 @@ public function setPhpName($phpName = null): void
      *
      * @return void
      */
-    public function setPhpSingularName($phpSingularName = null): void
+    public function setPhpSingularName(?string $phpSingularName = null): void
```
oojacoboo, line 618: We should remove the default `null` assignment on setters.

src/Propel/Generator/Model/Column.php
```diff
@@ -910,7 +910,7 @@ public function isNotNull(): bool
      *
      * @return void
      */
-    public function setNotNull($flag = true): void
+    public function setNotNull(bool $flag = true): void
```
oojacoboo, line 913: Setters don't need defaults.

src/Propel/Generator/Model/Column.php
```diff
@@ -958,7 +958,7 @@ public function isPrimaryString(): bool
      *
      * @return void
      */
-    public function setPrimaryKey($flag = true): void
+    public function setPrimaryKey(bool $flag = true): void
```
oojacoboo, line 961: No defaults on setters.

src/Propel/Generator/Model/Column.php
```diff
@@ -1584,7 +1584,7 @@ public function getAutoIncrementString(): string
      *
      * @return void
      */
-    public function setAutoIncrement($flag = true): void
+    public function setAutoIncrement(bool $flag = true): void
```
oojacoboo, line 1587: Remove default

src/Propel/Generator/Model/Table.php
```diff
@@ -2004,7 +2004,7 @@ public function isForReferenceOnly(): ?bool
      *
      * @return void
      */
-    public function setForReferenceOnly($flag = true): void
+    public function setForReferenceOnly(bool $flag = true): void
```
oojacoboo, line 2007: default not needed

src/Propel/Generator/Model/Table.php
```diff
@@ -2074,9 +2074,9 @@ public function __clone()
         foreach ($this->columns as $oldCol) {
             $col = clone $oldCol;
             $columns[] = $col;
-            $this->columnsByName[$col->getName()] = $col;
-            $this->columnsByLowercaseName[strtolower($col->getName())] = $col;
-            $this->columnsByPhpName[$col->getPhpName()] = $col;
+            $this->columnsByName[(string)$col->getName()] = $col;
```
oojacoboo, line 2077: Are these casts really needed or were they from WIP commits?

oojacoboo:
@dereuromark Overall, this looks great.  There is obviously a ton of changes here.  It'd be nice to have typed properties as well.  I find it's easier to type methods and properties at the same time.  However, seeing as this is already done, that'd be better in another PR.  I added some comments.

src/Propel/Generator/Model/Table.php
dereuromark: As long as we dont have a get*OrFail() method ( ?string returned ).

dereuromark:
Changes done as per review. Shall we continue?

gechetspr (approved):

src/Propel/Generator/Model/Column.php
dereuromark: This caused quite a few additional tests to fail
Given the BC impact I am a little less sure about this one now.

But I can see why having it not defaulting seems OK moving forward since the naming does not clearly indicate a positive default value (useAsPrimaryKey etc would be a different story).

dereuromark:
@oojacoboo Do you want to make follow up PRs regarding more types?
Especially also around generated code?

oojacoboo:
@dereuromark I think we need to get property typing done.

PR #1841, opened 2022-03-18
Adjust as per review.

dereuromark:
=> https://github.com/propelorm/Propel2/pull/1833

### DIFF
--- a/src/Propel/Generator/Builder/DataModelBuilder.php
+++ b/src/Propel/Generator/Builder/DataModelBuilder.php
@@ -262,11 +262,11 @@ abstract class DataModelBuilder
      * @param \Propel\Generator\Model\Table $table
      * @param string $classname The class of builder
      *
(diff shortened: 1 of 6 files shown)


src/Propel/Generator/Builder/DataModelBuilder.php
```diff
@@ -262,11 +262,11 @@ public function getMultiExtendObjectBuilder(): MultiExtendObjectBuilder
      * @param \Propel\Generator\Model\Table $table
      * @param string $classname The class of builder
      *
-     * @return self
+     * @return static
```
gechetspr, line 265: mb $this?

dereuromark: $builder = new $classname($table);

It cannot possibly be the same object.

gechetspr (approved):

PR #1843, opened 2022-03-21
Adjust generated code for type strictness.

dereuromark:
Follow https://github.com/propelorm/Propel2/pull/1833

How can we more easily debug fatal errors of generated code?

> PHP Fatal error:  A function with return type must return a value in vfs://root/propelQuickBuild/Stuff.php on line 1362

Currently, it goes to memory, so it is not easily introspectable.

Follow PRs to this PR are welcome, I started this but I do need some help to finish it up.

### DIFF
--- a/src/Propel/Generator/Behavior/NestedSet/NestedSetBehaviorObjectBuilderModifier.php
+++ b/src/Propel/Generator/Behavior/NestedSet/NestedSetBehaviorObjectBuilderModifier.php
@@ -281,9 +281,10 @@ if (\$this->isInTree()) {
 /**
  * Execute queries that were saved to be run inside the save transaction
  *
- * @param  ConnectionInterface \$con Connection to use.
+ * @param ConnectionInterface \$con Connection to use.
+ * @return void
  */
-protected function processNestedSetQueries(ConnectionInterface \$con)
+protected function processNestedSetQueries(ConnectionInterface \$con): void
 {
     foreach (\$this->nestedSetQueries as ['callable' => \$callable, 'arguments' => \$arguments]) {
         \$arguments[] = \$con;
@@ -306,9 +307,9 @@ protected function processNestedSetQueries(ConnectionInterface \$con)
  * Proxy getter method for the left value of the nested set model.
  * It provides a generic way to get the value, whatever the actual column name is.
  *
- * @return     int The nested set left value
+ * @return int The nested set left value
  */
-public function getLeftValue()
+public function getLeftValue(): int
 {
     return \$this->{$this->getColumnAttribute('left_column')};
 }
@@ -327,9 +328,9 @@ public function getLeftValue()
  * Proxy getter method for the right value of the nested set model.
  * It provides a generic way to get the value, whatever the actual column name is.
  *
- * @return     int The nested set right value
+ * @return int The nested set right value
  */
-public function getRightValue()
+public function getRightValue(): int
 {
     return \$this->{$this->getColumnAttribute('right_column')};
 }
@@ -348,9 +349,9 @@ public function getRightValue()
  * Proxy getter method for the level value of the nested set model.
  * It provides a generic way to get the value, whatever the actual column name is.
  *
- * @return     int The nested set level value
+ * @return int The nested set level value
  */
-public function getLevel()
+public function getLevel(): int
 {
     return \$this->{$this->getColumnAttribute('level_column')};
 }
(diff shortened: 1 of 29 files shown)


oojacoboo:
> Currently, it goes to memory, so it is not easily introspectable

@dereuromark what does this mean exactly?  I haven't tried debugging the model generators so I'm not exactly clear on what the issue is here?

In general though, for any issue related to debugging a generator, throwing more exceptions is going to likely be the answer, which would require conditional checks before passing values to a typed method/constructor, etc.

dereuromark:
I already found and fixed the code in quesiton.
So should be all good

We currently have a few open questions regarding fluent API and $this - instead of void - on the ObjectBuilder methods.

yaroslav-spryker (approved):

src/Propel/Generator/Behavior/NestedSet/NestedSetBehaviorObjectBuilderModifier.php
```diff
@@ -281,9 +281,10 @@ protected function addProcessNestedSetQueries(string &$script): void
 /**
  * Execute queries that were saved to be run inside the save transaction
  *
- * @param  ConnectionInterface \$con Connection to use.
+ * @param ConnectionInterface \$con Connection to use.
+ * @return void
```
gechetspr, line 285: other classes have an empty line before return param

src/Propel/Generator/Behavior/NestedSet/NestedSetBehaviorObjectBuilderModifier.php
```diff
@@ -671,7 +678,7 @@ protected function addGetParent(string &$script): void
  * @param  ConnectionInterface \$con Connection to use.
```
gechetspr, line 678: docblock should be also updated IMO

src/Propel/Generator/Behavior/NestedSet/NestedSetBehaviorObjectBuilderModifier.php
```diff
@@ -701,7 +708,7 @@ protected function addHasPrevSibling(string &$script): void
  * @param      ConnectionInterface \$con Connection to use.
```
gechetspr, line 708: same

src/Propel/Generator/Behavior/NestedSet/NestedSetBehaviorObjectBuilderModifier.php
```diff
@@ -736,7 +743,7 @@ protected function addGetPrevSibling(string &$script): void
...
```
gechetspr, line 743: same

src/Propel/Generator/Behavior/NestedSet/NestedSetBehaviorObjectBuilderModifier.php
```diff
@@ -766,7 +773,7 @@ protected function addHasNextSibling(string &$script): void
...
```
gechetspr, line 773: same

src/Propel/Generator/Behavior/NestedSet/NestedSetBehaviorObjectBuilderModifier.php
```diff
@@ -801,7 +808,7 @@ protected function addGetNextSibling(string &$script): void
...
```
gechetspr, line 808: same

src/Propel/Generator/Behavior/NestedSet/NestedSetBehaviorObjectBuilderModifier.php
```diff
@@ -932,7 +939,7 @@ protected function addGetChildren(string &$script): void
...
```
gechetspr, line 939: please check this class

src/Propel/Generator/Behavior/NestedSet/NestedSetBehaviorObjectBuilderModifier.php
dereuromark: We dont enforce CS just yet on those files.

src/Propel/Generator/Behavior/NestedSet/NestedSetBehaviorObjectBuilderModifier.php
dereuromark: True, but I think this should be a separate CS cleanup PR as outlined in beta ticket
Mixing all those into is out of scope.

PR #1844, opened 2022-03-21
Fixed the generated code.

limeeugenia:


src/Propel/Generator/Builder/Om/ObjectBuilder.php
```diff
@@ -7135,7 +7135,7 @@ protected function addClear(string &$script): void
      *
      * @return \$this
      */
-    public function clear()
+    public function clear(): void
```
dereuromark, line 7138: I changed all to chainable $this

PR #1847, opened 2022-03-23
Make 8.2 test run

dereuromark:


dereuromark:
Good news, looks like 8.2+ is compatible out of the box.

Issue #1634, opened 2020-07-23
Move templates out of src/


dereuromark:
as per https://github.com/php-pds/skeleton
and PSR4 (which has to manually skip the non-autoloadable files in src/ right now)

PHPStan already has to manually ignore those, CS tooling the same:

        - '%rootDir%/../../../src/Propel/Generator/Command/templates/*'
        - '%rootDir%/../../../src/Propel/Generator/Behavior/AggregateColumn/templates/*'
        - '%rootDir%/../../../src/Propel/Generator/Behavior/I18n/templates/*'
        - '%rootDir%/../../../src/Propel/Generator/Builder/Om/templates/*'
        - '%rootDir%/../../../src/Propel/Generator/Behavior/Validate/templates/*'
        - '%rootDir%/../../../src/Propel/Generator/Behavior/Archivable/templates/*'
        - '%rootDir%/../../../src/Propel/Generator/Behavior/Delegate/templates/*'
        - '%rootDir%/../../../src/Propel/Generator/Behavior/Sortable/templates/*'
        - '%rootDir%/../../../src/Propel/Generator/Behavior/NestedSet/templates/*'

This would make it a lot easier for all tools and the autoloading.

We can either
- create a new top-level `templates/`
- or nest it inside existing `resources/` as `resources/templates/`

I would go with the first as the templates then need more nesting as per namespace
- templates/Behavior/AggregateColumn/...
- templates/Behavior/Archivable/...

So we just switch the /templates/ piece to the front

Ideas? Feedback?

The idea is to check how this can be fully BC to avoid a BC break.
Custom (extensions) code can probably still do either/or based on how we allow this to be configured.

dereuromark:
Also test fixtures should be moved from tests (~Test.php) to normal fixture folder.
- tests/Propel/Tests/Generator/Behavior/NestedSet/Fixtures
- tests/Propel/Tests/Generator/Builder/Om/Fixtures

etc => tests/Fixtures/

dereuromark:
Can this be done in a BC way?

PR #1848, opened 2022-03-24
Move templates into own root level.

dereuromark:
Solves https://github.com/propelorm/Propel2/issues/1634

The idea:

Template paths are by convention in
- templates/
- besides src/ folder (which is required to detect root path)

Any extensions should follow the same convention or pass their custom path to renderTemplate()

BC Impact:
- BC breaking for all template extensions used in user land code
- Must be outlined in changelogs.

If needed, we could add a shim to lookup also in old "folder" - if not found in new primary location.
This would mitigate the impact.


dereuromark:
@mringler @nederdirk
What do you think about it? Any concerns or input?

yaroslav-spryker (approved):

src/Propel/Common/Util/PathTrait.php
```diff
@@ -0,0 +1,78 @@
...
+     *
+     * @param string $path
+     *
+     * @throws \RuntimeException
+     *
+     * @return string
+     */
+    protected function getTemplatePath(string $path): string
+    {
+        if ($this->path) {
```
gechetspr, line 42: I have my concern regarding this code. We have a param that can change a return value, but we will always return the result of the first call of this method. it mb ok for the current usecases, but IMO this is unexpected and not transparent

dereuromark: Should we avoid this lookup you mean?
It could fail if we have nested objects doing getTemplatePath() from different points of extension, sure.

gechetspr: I mean this method should not store result in a property

gechetspr (approved):

dereuromark:
I will merge now.
Let me know if we need to enable BC path lookup shim for the beta release - if needed.

PR #1855, opened 2022-03-29
Add BC shim for template lookup in former directories.

dereuromark:
Follow https://github.com/propelorm/Propel2/pull/1848

Just to make sure we ease the upgrade issues here for folks that use behaviors or other custom repositories with extension and templates.

### DIFF
--- a/src/Propel/Common/Util/PathTrait.php
+++ b/src/Propel/Common/Util/PathTrait.php
@@ -15,10 +15,10 @@ trait PathTrait
     /**
      * Template paths are by convention in
      * - templates/
-     * - besides src/ folder (which is required to detect root path)
+     * - same level as src/ folder (which is required to auto-detect root path)
      *
      * Note:
-     * - Propel/Generator/ prefix is removed from the path
+     * - Propel/Generator/ namespace prefix is removed from the path
      *
      * Examples:
      * - Behavior/BehaviorName/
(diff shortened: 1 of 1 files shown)


dereuromark:
Could someone dev-qa and approve this please?
@mringler @oojacoboo etc

oojacoboo:
Logic seems okay, but I don't fully know what input and output to expect here, so it's hard to give it much scrutiny.

src/Propel/Common/Util/PathTrait.php
```diff
@@ -36,7 +36,15 @@ protected function getTemplatePath(string $path): string
     {
         $srcPos = strpos($path, DIRECTORY_SEPARATOR . 'src' . DIRECTORY_SEPARATOR);
         if ($srcPos === false) {
```
mringler, line 38: I am probably misunderstanding this, but above, it says templates can be on the same level as src folder, i.e. `./Propel2/templates`, which does not contain `/src/`. Wouldn't that throw an error here, even though it is a valid path?

mringler (approved):

dereuromark:
You can see the linked pr for previous template paths

mringler (approved):
Great, really good to give users feedback why something isn't working.

In general, the extraction of the path elements (`$elements`) could be done a bit more readable (I think) as
```php
        $lastSrcPos = strrpos($path, DIRECTORY_SEPARATOR . 'src' . DIRECTORY_SEPARATOR);
        $pathAfterSrc = substr($path, $lastSrcPos + 5);
        $elements = explode(DIRECTORY_SEPARATOR, $pathAfterSrc);
```
but that's fine

PR #1863, opened 2022-04-07
Cleanup AbstractOMBuilder class for generating models

yaroslav-spryker:


src/Propel/Generator/Builder/DataModelBuilder.php
```diff
@@ -518,6 +518,18 @@ public function getDatabase(): ?Database
         return $this->getTable()->getDatabase();
     }
 
+    /**
+     * Convenience method to returns the database for current table.
```
dereuromark, line 522:
```suggestion
     * Convenience method to return the database for current table.
```

src/Propel/Generator/Model/Table.php
```diff
@@ -1861,9 +1861,9 @@ public function getDatabase(): ?Database
     /**
      * Get the database that contains this table.
      *
-     * @throws \Propel\Generator\Exception\LogicException
-     *
      * @return \Propel\Generator\Model\Database
+     *
+     * @throws \Propel\Generator\Exception\LogicException
```
dereuromark, line 1866: Is that change on purpose? Shouldnt CI be red here?

PR #1872, opened 2022-04-19
add type-safe access to StandardServiceContainer

mringler:
When using `Propel::getServiceContainer()`, an instance of ServiceContainerInterface is returned, and users can implement their own service container against this interface.

However, almost all users do not do that, and instead use the StandardServiceContainer, which offers crucial functionality, particularly enabling and configuring the logger.

But since the additional functionality of  StandardServiceContainer is not part of the interface, they have to find those methods without IDE support, and if they use type checkers, those will complain because unknown methods are accessed. See for example [this discussion](https://github.com/propelorm/Propel2/discussions/1871).

Since there are no generics in PHP, a feasible workaround is to add a new method to the Propel runtime object, which allows type-safe access to the StandardServiceContainer.

If there is a better solution, I'd be happy to find out!

### DIFF
--- a/src/Propel/Runtime/Propel.php
+++ b/src/Propel/Runtime/Propel.php
@@ -11,6 +11,7 @@ namespace Propel\Runtime;
 use Propel\Runtime\Adapter\AdapterInterface;
 use Propel\Runtime\Connection\ConnectionInterface;
 use Propel\Runtime\Connection\ConnectionManagerInterface;
+use Propel\Runtime\Exception\PropelException;
 use Propel\Runtime\Map\DatabaseMap;
 use Propel\Runtime\ServiceContainer\ServiceContainerInterface;
 use Propel\Runtime\ServiceContainer\StandardServiceContainer;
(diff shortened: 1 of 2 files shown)


src/Propel/Runtime/Propel.php
```diff
@@ -145,6 +146,29 @@ public static function getServiceContainer(): ServiceContainerInterface
...
+     *
+     * This method allows type-safe access to methods of
+     * StandardServiceContainer, it provides no other advantage over
+     * {@link Propel::getServiceContainer()}
+     *
+     * @throws \Propel\Runtime\Exception\PropelException
+     *
+     * @return \Propel\Runtime\ServiceContainer\StandardServiceContainer
+     */
+    public static function getStandardServiceContainer(): StandardServiceContainer
```
yaroslav-spryker, line 162: It seems that due to the method logic it would be better to call it `getStandardServiceContainerOrFail()` just because this method throws an exception.

yaroslav-spryker (changes requested):

src/Propel/Runtime/Propel.php
dereuromark: We only add orFail() though if we need to keep a nullable alternative. This seems not like that case.

PR #1873, opened 2022-04-19
Set connection debug mode on StandardServiceContainer

mringler:
In order to get log messages about executed queries, the connection has to be set to debug mode. Currently, this is done by accessing the connection through the service container:
```php
serviceContainer = Propel::getServiceContainer();
$connection = $serviceContainer->getWriteConnection('default');
$connection->useDebug(true);
```
This is a bit cumbersome for two reasons (see [this discussion](https://github.com/propelorm/Propel2/discussions/1871)):

1. Users have to interact with Propel internals
2. The method `useDebug()` is only available on ConnectionWrapper, but not on ConnectionInterface, which is what `ServiceContainerInterface::getWriteConnection()` returns, causing type errors when using a checker (or more boilerplate code).

A simple solution is to allow managing of the connection debug mode through the service container.

Currently, the new method to set connection debug mode allows for three values: true, false and null, where null means to leave the connection unchanged. If you prefer it, I can add consts for those values.

### DIFF
--- a/src/Propel/Runtime/ServiceContainer/StandardServiceContainer.php
+++ b/src/Propel/Runtime/ServiceContainer/StandardServiceContainer.php
@@ -111,7 +111,7 @@ class StandardServiceContainer implements ServiceContainerInterface
     /**
      * @var bool|null
      */
-    protected $setWrappedConnetionToDebug;
+    protected $useDebugModeFlag;
 
     /**
      * @return string
@@ -482,10 +482,10 @@ class StandardServiceContainer implements ServiceContainerInterface
     {
         if (
             $connection instanceof ConnectionWrapper
-            && $this->setWrappedConnetionToDebug !== null
-            && $connection->useDebug !== $this->setWrappedConnetionToDebug
+            && $this->useDebugModeFlag !== null
+            && $connection->useDebug !== $this->useDebugModeFlag
         ) {
-            $connection->useDebug($this->setWrappedConnetionToDebug);
+            $connection->useDebug($this->useDebugModeFlag);
         }
     }
 
@@ -673,9 +673,10 @@ class StandardServiceContainer implements ServiceContainerInterface
     }
 
     /**
-     * Create connections in debug mode.
+     * Enable or disable debug output.
      *
-     * This only works with the default ConnectionWrapper.
+     * Sets created connections in debug mode. This only works when the
+     * default ConnectionWrapper is used..
      *
      * @see \Propel\Runtime\Connection\ConnectionWrapper::useDebug()
      *
@@ -683,8 +684,8 @@ class StandardServiceContainer implements ServiceContainerInterface
      *
      * @return void
      */
-    public function setWrappedConnectionDebugMode(?bool $useDebug = true): void
+    public function useDebugMode(?bool $useDebug = true): void
     {
-        $this->setWrappedConnetionToDebug = $useDebug;
+        $this->useDebugModeFlag = $useDebug;
     }
 }
--- a/src/Propel/Runtime/ServiceContainer/StandardServiceContainer.php
+++ b/src/Propel/Runtime/ServiceContainer/StandardServiceContainer.php
@@ -109,7 +109,7 @@ class StandardServiceContainer implements ServiceContainerInterface
     protected $loggerConfigurations = [];
 
     /**
-     * @var bool | null
+     * @var bool|null
      */
     protected $setWrappedConnetionToDebug;
 
@@ -679,7 +679,7 @@ class StandardServiceContainer implements ServiceContainerInterface
      *
      * @see \Propel\Runtime\Connection\ConnectionWrapper::useDebug()
      *
-     * @param bool | null $useDebug
+     * @param bool|null $useDebug
      *
      * @return void
      */
--- a/src/Propel/Runtime/ServiceContainer/StandardServiceContainer.php
+++ b/src/Propel/Runtime/ServiceContainer/StandardServiceContainer.php
@@ -18,6 +18,7 @@ use Propel\Runtime\Adapter\Exception\AdapterException;
 use Propel\Runtime\Connection\ConnectionInterface;
 use Propel\Runtime\Connection\ConnectionManagerInterface;
 use Propel\Runtime\Connection\ConnectionManagerSingle;
+use Propel\Runtime\Connection\ConnectionWrapper;
 use Propel\Runtime\Exception\PropelException;
 use Propel\Runtime\Exception\RuntimeException;
 use Propel\Runtime\Exception\UnexpectedValueException;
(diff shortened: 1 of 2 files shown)


tests/Propel/Tests/Runtime/ServiceContainer/StandardServiceContainerTest.php
```diff
@@ -672,6 +674,40 @@ public function testGetLoggerLazyLoadsLoggerFromConfiguration()
...
+        $this->assertWrappedConnectionInDebugMode($isDebugMode, $readConnection);
+        
+        $writeConnection = $this->sc->getWriteConnection($connectionName);
+        $this->assertWrappedConnectionInDebugMode($isDebugMode, $writeConnection);
+    }
+    
+    protected function assertWrappedConnectionInDebugMode(bool $isDebugMode, ConnectionInterface $connection): void
+    {
+        $this->assertInstanceOf(ConnectionWrapper::class, $connection, 'Connection should be a ConnectionWrapper');
+        $this->assertEquals($isDebugMode, $connection->useDebug);
```
dereuromark, line 709: Best to always use assertSame here

src/Propel/Runtime/ServiceContainer/StandardServiceContainer.php
```diff
@@ -107,6 +108,11 @@ class StandardServiceContainer implements ServiceContainerInterface
      */
     protected $loggerConfigurations = [];
 
+    /**
+     * @var bool | null
```
dereuromark, line 112: Surprised the cs doesnt complain here

mringler: Surprised me too, thought it had to be `mixed`

dereuromark: I refer to the whitespace, should be bool|null

yaroslav-spryker:
Hi @mringler from my perspective the service container does not really need a knowledge about connection it wraps and even more does not need to have a state (`$this->setWrappedConnetionToDebug` property) for debugging connection.
Propel has it`s default way to inject a logger via service container and I saw you already mentioned this in related dicsussion
https://github.com/propelorm/Propel2/discussions/1871

`$loggerConfiguration = [
    'type' => 'stream',
    'path' => 'path/to/file'
    'level' => Logger::INFO, // Defaults to Logger::CRITICAL,
    'bubble' => true, // Defaults to true
];
Propel::getServiceContainer()->setLoggerConfiguration('defaultLogger', $loggerConfiguration);`

mringler:
Hey @yaroslav-spryker,
not sure I understand your comment correctly. The code you posted is about setting a logger, but this is about setting the connection into debug mode - which uses the logger, but is a completely different matter.

In general, the idea is to allow fast and type-safe access to important Propel functionality, i.e. allow for this:
```php
Propel::getServiceContainer()->setWrappedConnetionToDebug(true);
```
instead of this:
```php
$connection = Propel::getServiceContainer()->getWriteConnection('default');
if ($connection instanceof ConnectionWrapper) {
    $connection->useDebug(true);
}
```
Note that this snipped is not available in the [documentation](http://propelorm.org/documentation/07-logging.html#enabling-the-debug-mode), the functionality is not apparent when browsing through the ServiceContainer or ConnectionInterface, and it requires users to know about the `ConnectionWrapper` class. So basically, you cannot use this functionality unless someone tells you about it, even though it is crucial to get connection output during development (without it, you are left with guessing why a query is not working).

I think that leads to the impression of bad usability. And hence my proposal to ease things up, similar to how `setLoggerConfiguration()` is a convenience over setting the logger manually.

You don't see this as an improvement?

src/Propel/Runtime/ServiceContainer/StandardServiceContainer.php
```diff
@@ -107,6 +108,11 @@ class StandardServiceContainer implements ServiceContainerInterface
      */
     protected $loggerConfigurations = [];
 
+    /**
+     * @var bool|null
+     */
+    protected $useDebugModeFlag;
```
dereuromark, line 114:
```suggestion
    protected $useDebugMode;
```
It is already a flag by design since it is bool
No need to mention that IMO.

mringler: Yes, but `useDebugMode` is already taken by the method, and while I think it is possible to use the same name in a method and a variable, it probably leads to unnecessary confusion.

Naming it as the flag which is set by `useDebugMode()` seems like the most readable approach, but I am happy to change it. Let me know if you want to use the same name in both method and variable or prefer another name.

yaroslav-spryker:
@mringler Sorry seems I formulated the idea is not clear enough :(
What I wanted to point out that to my opinion setting debug mode for connection is a bit out of scope of ServiceContainer. If we look at what ServiceContainer has inside - we see only setting and getting previously provided functionality.
Moreover it should not have a logic inside of it like checking an instance of `ConnectionWrapper`.
I would better implement it other ways, maybe via configuration setting?
Propel has already something similar with setting connection wrapper class so maybe it's better to add one more setting from there?

Please note that I don't discuss implementation details but the approach itself :)

src/Propel/Runtime/ServiceContainer/StandardServiceContainer.php
dereuromark: Should we maybe use `$isInDebugMode` or alike?

mringler:
@yaroslav-spryker So your proposal is to bury the debug flag deeper into Propel, i.e. into the ConnectionManagers, the ConnectionFactory, or the ConnectionWrapper? Where exactly would you do that?

mringler:
@yaroslav-spryker Only solution I can think of is to change ConnectionWrapper by making $useDebug (which is public) to a getter, and add a static variable:
```php
class ConnectionWrapper implements ConnectionInterface, LoggerAwareInterface
{
    public static useDebug = false;
    protected useDebugOnInstance = null;

    public isInDebugMode(): bool
    {
        return $this->useDebugOnInstance ?? static::useDebug;
    }
```
Then debug mode can be enabled by changing the static variable:
```php
ConnectionWrapper::useDebug = true;
```
But considering how static variables are not popular in the project, I am not sure if this will find much love. What do you think?

dereuromark:
A static one in this particular case could be fair enough.

mringler:
> A static one in this particular case could be fair enough. Maybe make that separate PR and we can then as a group discuss what solution is better to merge.

Sure, what else to do on vacation: https://github.com/propelorm/Propel2/pull/1874

yaroslav-spryker:
> > A static one in this particular case could be fair enough. Maybe make that separate PR and we can then as a group discuss what solution is better to merge.
> 
> Sure, what else to do on vacation: #1874

@mringler Sorry for the late response but yeah, using ConnectionWrapper seems to be a way more cleaner solution in our case. Thank you very much for additional PR, I've already approved it.

PR #1874, opened 2022-04-20
Set connection debug mode on StandardServiceContainer through class-w…

mringler:
Alternative implementation of https://github.com/propelorm/Propel2/pull/1873, see discussion there

### DIFF
--- a/src/Propel/Common/Config/FileLocator.php
+++ b/src/Propel/Common/Config/FileLocator.php
@@ -27,7 +27,7 @@ class FileLocator extends BaseFileLocator
     {
         if ($configDirectories === null) {
             $configDirectories = [
-                getcwd(),
+                (string)getcwd(),
                 'config',
                 'conf',
             ];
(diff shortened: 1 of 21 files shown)


dereuromark:
I am fine with this approach
As soon as we decided we should also adjust docs IMO.

mringler:
Setting debug mode through the StandardServiceContainer now also allows to set the connection in profiler mode. Similarly to setting debug mode on the ConnectionWrapper, this is done through a static flag in ConnectionFactory.

This allows users to easily use profiled connections (which outputs query execution time) without having to  fumble around with Propel internals. Current method is:

```php
        /** @var ConnectionManagerSingle */
        $manager = Propel::getServiceContainer()->getConnectionManager('default');
        $config  = $manager->getConfiguration();
        $config['classname'] = ProfilerConnectionWrapper::class;
        $manager->setConfiguration($config);
```

yaroslav-spryker (approved):

PR #1881, opened 2022-05-09
MSSQL/SQLSrv Datetime format fixes

DarkAxi0m:
I know there is an older PR https://github.com/propelorm/Propel2/pull/1328  that covers this change, 
But it's based on 2017 code, and does not look like its going to cleaned up anytime soon.

I have just taken the DateTime fix to it's own PR.


dereuromark:
Ideally this would come with a test case, but I am fine with it either way if it improves the situation.

DarkAxi0m:
> Ideally this would come with a test case, but I am fine with it either way if it improves the situation.

I just had a quick look now, am I right in assuming there are no tests on MSSQL DBs currently?

PR #1886, opened 2022-06-13
Fixing Psalm CI Issues

oojacoboo:
For some reason Psalm isn't running the same for CI as local.  It's also running again the `vendor` dir on CI, which is wrong.  Additionally, I get a lot of analysis errors locally when targeting the default config.  It's possible the CI isn't properly using it either.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -53,7 +53,7 @@
         "cs-fix": "phpcbf",
         "stan": "vendor/bin/phpstan analyze",
         "stan-baseline": "vendor/bin/phpstan analyze --generate-baseline",
-        "psalm": "vendor/bin/psalm.phar --show-info=false",
+        "psalm": "vendor/bin/psalm.phar --config=psalm.xml",
         "psalm-update-baseline": "vendor/bin/psalm.phar --update-baseline",
         "psalm-set-baseline": "vendor/bin/psalm.phar --set-baseline=psalm-baseline.xml",
         "psalm-update-report": "vendor/bin/psalm.phar --report=psalm-report.xml",
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
         "phpstan/phpstan": "^1.2",
         "phpunit/phpunit": "^9.5.0",
         "spryker/code-sniffer": "^0.17.2",
-        "psalm/phar": "^4.14",
+        "psalm/phar": "^4.23",
         "mikey179/vfsstream": "^1.6"
     },
     "suggest": {
--- a/composer.json
+++ b/composer.json
@@ -16,15 +16,15 @@
         }
     ],
     "require": {
-        "php": ">=7.4",
-        "psr/log": "^1.0||^2.0||^3.0",
-        "symfony/yaml": "^4.4.0||^5.0.0||^6.0.0",
-        "symfony/config": "^4.4.0||^5.0.0||^6.0.0",
-        "symfony/console": "^4.4.0||^5.0.0||^6.0.0",
-        "symfony/filesystem": "^4.4.0||^5.0.0||^6.0.0",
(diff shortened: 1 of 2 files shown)


mringler:
**tl;dr**: The problem seems to be that Psalm is running in PHP 7.4 mode, but the Symfony package uses PHP 8 features, a simple change to the Github Workflow file should fix it.

Looking at the [output from the failed run](https://github.com/propelorm/Propel2/actions/runs/2488492497/workflow), we can see that Propel uses a Psalm version which is capable of checking PHP 8 features (v4.23), but it is run in php 7 mode:
```
==> Setup PHP
✓ PHP Found PHP 8.1.6
...
 - Locking psalm/phar (4.23.0)
...
 - Locking symfony/yaml (v6.1.0)
...
Run composer psalm
> vendor/bin/psalm.phar --config=psalm.xml
Target PHP version: 7.4 (inferred from composer.json)
```
In the pipeline, PHP 8.1.6 is installed, but psalm still uses 7.4, I guess this is due to the minimal required version in composer.json:
```
    "require": {
        "php": ">=7.4",
        ...
```

The error I looked at happens in the Symfony Yaml-Parser (v6.1):
```
ERROR: ParseError - vendor/symfony/yaml/Parser.php:1047:38 - Syntax error, unexpected T_DOUBLE_ARROW on line 1047 (see https://psalm.dev/173)
                \PREG_INTERNAL_ERROR => 'Internal PCRE error.',
``` 
In the [Parser code](https://github.com/symfony/yaml/blob/6.1/Parser.php), that line is part of a PHP 8 `match` statement: 
```php
    public static function preg_match(string $pattern, string $subject, array &$matches = null, int $flags = 0, int $offset = 0): int
    {
        if (false === $ret = preg_match($pattern, $subject, $matches, $flags, $offset)) {
            $error = match (preg_last_error()) {
                \PREG_INTERNAL_ERROR => 'Internal PCRE error.', // <--------------- here
                \PREG_BACKTRACK_LIMIT_ERROR => 'pcre.backtrack_limit reached.',
                \PREG_RECURSION_LIMIT_ERROR => 'pcre.recursion_limit reached.',
                \PREG_BAD_UTF8_ERROR => 'Malformed UTF-8 data.',
                \PREG_BAD_UTF8_OFFSET_ERROR => 'Offset doesn\'t correspond to the begin of a valid UTF-8 code point.',
                default => 'Error.',
            };

            throw new ParseException($error);
        }

        return $ret;
    }
```
All consecutive errors in the file are probably due to that first parse error.

So now I am pretty sure that this is not a problem with the Propel code, but with CI/CD setup.

I think the fix is to add a command line parameter to the psalm command in the Github Workflow file (ci.yml) in line 161, something like this should work:
```
            - name: Psalm
              run: composer psalm -- --php-version=${{ matrix.php-version }}
```
Being part of the hoi polloi, I cannot change this.

@dereuromark can you try that out, see if it fixes the issue?

oojacoboo:
@mringler thanks for helping dig into this.  Took a little massaging, but looks like I got it working/passing.

composer.json
```diff
@@ -53,7 +53,7 @@
         "cs-fix": "phpcbf",
         "stan": "vendor/bin/phpstan analyze",
         "stan-baseline": "vendor/bin/phpstan analyze --generate-baseline",
-        "psalm": "vendor/bin/psalm.phar --show-info=false",
+        "psalm": "vendor/bin/psalm.phar --config=psalm.xml",
```
dereuromark, line 56: Do we need the config? Isnt it defaulting to this file?

oojacoboo: We probably don't need it, but there are multiple configs and it wasn't clear if there were issues with which config was being targeted while debugging.  I didn't see any harm in leaving it.

PR #1888, opened 2022-06-17
[WIP] Added strict params and return types in generated code

geega:


src/Propel/Generator/Builder/Om/ObjectBuilder.php
```diff
@@ -3953,7 +3953,7 @@ protected function addSetPrimaryKeyMultiPK(string &$script): void
      * @param array \$keys The elements of the composite key (order must match the order in XML file).
```
dereuromark, line 3953: can we specify here the value type? array<..> etc?

dereuromark (approved):

src/Propel/Generator/Builder/Om/ObjectBuilder.php
geega: Looks like for this case we can have an array  with mixed value 
If we have primary key use multiple-column index

```
CREATE TABLE
    item (
        tag VARCHAR(255) NOT NULL,
        id INT NOT NULL,
        PRIMARY KEY(tag, id)
    );
```

PR #1889, opened 2022-06-27
Added a common return type for Collection::getIterator() method

gechetspr:
`OnDemandCollection::getIterator` docblock has not compatible type with the base class. Added a common type for both classes

### DIFF
--- a/src/Propel/Runtime/Collection/Collection.php
+++ b/src/Propel/Runtime/Collection/Collection.php
@@ -201,7 +201,7 @@ class Collection implements ArrayAccess, IteratorAggregate, Countable, Serializa
     }
 
     /**
-     * @return \Propel\Runtime\Collection\CollectionIterator|\Traversable
+     * @return \Propel\Runtime\Collection\CollectionIterator|\Traversable|\Countable
      */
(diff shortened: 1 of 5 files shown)


dereuromark (approved):

src/Propel/Runtime/Collection/Collection.php
```diff
@@ -201,7 +201,7 @@ public function setData(array $data): void
     }
 
     /**
-     * @return \Propel\Runtime\Collection\CollectionIterator
+     * @return \Propel\Runtime\Collection\CollectionIterator|\Propel\Runtime\Collection\IteratorInterface
```
dereuromark, line 204: With
> class CollectionIterator extends ArrayIterator implements IteratorInterfac
is this still needed here then?

gechetspr: OnDemandIterator is not a subtype of Collection iterator. So I can leave only interface here, but those iterator classes are quite different under the hood, because Collection one implements ArrayIterator and OnDemand one - not

dereuromark (approved):

PR #1892, opened 2022-07-09
Remove deprecated 

mringler:
I'm getting warnings because declarations of `setPrimaryKey()` do not match between extending classes:

> PHP Warning:  Declaration of Nivm\Model\Person\Base\Person::setPrimaryKey(?int $key = NULL): void should be compatible with Nivm\Model\Util\Base\ChangedBy::setPrimaryKey($pk): void in /srv/www/html/nivm/src/Nivm/Model/Person/Base/Person.php on line 2700

setPrimaryKey(?int $key = NULL): void vs
setPrimaryKey($pk): void

I am extending tables via concrete_inheritance behavior, and the parent table does not have a primary key. Apparently this leads to the parent class having a stub for setPrimaryKey(), which is defined differently than the function on the child class. Since the parent class has not PK declaration itself, and can be parent to tables with different PK types, it is not possible to insert a type hint there.

Fortunately, adding `setPrimaryKey()` to tables without PK has been marked as deprecated since 2011, so I think the problem can be easily solved by just removing that function.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         }
     ],
     "require": {
-        "php": ">=7.4 || >=8.0",
+        "php": ">=7.4",
         "psr/log": "^1.0 || ^2.0 || ^3.0",
         "symfony/yaml": "^4.4.0 || ^5.0.0 || ^6.0.0",
         "symfony/config": "^4.4.0 || ^5.0.0 || ^6.0.0",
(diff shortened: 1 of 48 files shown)


mringler:
Uff, test fail because of problem with test setup:
>Error: dealerdirect/phpcodesniffer-composer-installer contains a Composer plugin which is blocked by your allow-plugins config. You may add it to the list if you consider it safe.

Not sure where this comes from or how to handle it

dereuromark (approved):

dereuromark:
This is now solved, you can probably rebase

mringler:
Oh, completely forgot about this. Merged in master, now tests run through

PR #1893, opened 2022-07-09
added simple typed properties

a4blue:
The Idea of this PR is to prepare and add typed properties in the classes generated by propel.
Goal was to still maintain BC Compatibility

In the future strict return types would be great, but it would cause compatibility issues with many projects

### DIFF
--- a/src/Propel/Generator/Behavior/NestedSet/NestedSetBehaviorObjectBuilderModifier.php
+++ b/src/Propel/Generator/Behavior/NestedSet/NestedSetBehaviorObjectBuilderModifier.php
@@ -311,7 +311,7 @@ protected function processNestedSetQueries(ConnectionInterface \$con): void
  */
 public function getLeftValue(): int
 {
-    return \$this->{$this->getColumnAttribute('left_column')} ?: 0;
+    return \$this->{$this->getColumnAttribute('left_column')} ?? 0;
 }
 ";
     }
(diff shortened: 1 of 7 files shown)


mringler (changes requested):
That's an interesting contribution!

However, I am not happy with the idea to clear instance property values with `unset`. I believe it is part of the OOP contract that class properties are always available on an instance. If they hold no value, they are set to `null`, which is the value of a variable that has no value.

You should never try to remove a class property from an instance, even if the language allows it, as it will send you to software hell pretty much instantly. You can see that from other changes in this PR, which are now necessary to deal with the introduced uncertainty of possibly missing class variables. An example is `return \$this->$clo ?? null;";`, which should be redundant, but now is absolutely necessary to avoid `Undefined property` errors, which shows how this approach introduces a whole new class of errors.
Errors which will soon become the stuff of nightmares, when instances of a class start to not behave as declared.

To be blunt, I am convinced that in order to clear a class variable, the difference between setting it to null and actually removing it from the object is not a matter of taste or efficiency, but the latter is simply the wrong approach, due to its catastrophic consequences.

Please let me know if there is an upside I am not seeing, or if I misunderstand the code. For now, I strongly suggest Propel not going down this route.

dereuromark:
Any update on this one?

mringler:
@dereuromark I think this can be closed too.
As said above, the proposed changes are dangerous, and there was no response from the author in six month.

PR #1896, opened 2022-08-01
Saving object with modified primary key 

### LINKED ISSUE #978: Saving object with modified primary key
I've tried searching for a similar issue but couldn't find any (here at least, I did find something similar but with no clear solution on the forums).
How can I save an object if the PK has been modified ? The log shows that new PK is used in the WHERE clause (e.g: Original PK was 25)
>[2015-08-09 07:20:43] defaultLogger.INFO: UPDATE `author` SET `ID`=55 WHERE author.ID=55 [] []

Sandbox: http://sandbox.propelorm.org/486a545

Or am I supposed to to delete the old object and save a new one with the new PK ?

Thanks :wink:

basteyy:
This fix put the primary keys into an array while building the object. From this array, the values will be used to create the criteria for updating. After updating an object, the values in the array will be regenerated to ensure multiple changes in one request. 

Fixes https://github.com/propelorm/Propel2/issues/978


src/Propel/Generator/Builder/Om/ObjectBuilder.php
```diff
@@ -6377,6 +6383,13 @@ protected function doSave(ConnectionInterface \$con" . ($reloadOnUpdate || $relo
                 }
 ";
             }
+
+            // Update primary key values cache
+            if($col->isPrimaryKey()) {
```
dereuromark, line 6388: also some CS issues here

basteyy: Thanks, will fix it (https://github.com/propelorm/Propel2/runs/7615260378?check_suite_focus=true#step:10:186)

src/Propel/Generator/Builder/Om/ObjectBuilder.php
```diff
@@ -2719,6 +2719,12 @@ protected function addHydrateBody(string &$script): void
                 } elseif ($col->isPhpPrimitiveType()) {
                     $script .= "
             \$this->$clo = (null !== \$col) ? (" . $col->getPhpType() . ') $col : null;';
+                    if($col->isPrimaryKey()) {
```
dereuromark, line 2722: Same

basteyy: Changed

dereuromark:
Why are those checks failing
Didnt you use latest master as reference?

basteyy:
> Why are those checks failing Didnt you use latest master as reference?

I used latest master. They failed because of:

```
Error: dealerdirect/phpcodesniffer-composer-installer contains a Composer plugin which is blocked by your allow-plugins config. You may add it to the list if you consider it safe.
See https://getcomposer.org/allow-plugins

In PluginManager.php line 762:
                                                                               
  dealerdirect/phpcodesniffer-composer-installer contains a Composer plugin w  
  hich is blocked by your allow-plugins config. You may add it to the list if  
   you consider it safe.                                                       
  You can run "composer config --no-plugins allow-plugins.dealerdirect/phpcod  
  esniffer-composer-installer [true|false]" to enable it (true) or disable it  
   explicitly and suppress this exception (false)                              
  See https://getcomposer.org/allow-plugins                                    
```

See https://github.com/propelorm/Propel2/runs/7263266989?check_suite_focus=true

All other current PR also failing.

dereuromark:
Yeah, but we have the same here whitelisted
So this seems to be composer bug/regression
I opened a ticket for now: https://github.com/composer/composer/issues/10981

dereuromark:
Master is fixed, can you please rebase?

dereuromark:
I wonder if this should come with a test case
cc @mringler

templates/Builder/Om/baseObjectAttributes.php
```diff
@@ -24,3 +24,11 @@
      * @var array
      */
     protected $virtualColumns = [];
+
+    /**
+    * The primary key columns stored with the original values in that array
+    * to support later changes of the primary keys itself.
+    * @see https://github.com/propelorm/Propel2/issues/978
+    * @var array
```
dereuromark, line 32: should this be `array<string, int|string>` instead?

mringler:
> I wonder if this should come with a test case

Yes, @basteyy, could you please add a test verifying the updated behavior?

If you want to run test on your machine, there is a [page with hints how to get it running](http://propelorm.org/documentation/cookbook/working-with-test-suite.html). It is partially outdated though, specifically the part about "custom credentials". Not sure if that ever worked, but on MySQL, AFAIK you need to use user "admin" and password "" (empty string). Might be easier to use SQLite.
Anyway, let me know if you want help with that.

dereuromark:
Any update on this one?

basteyy:
> Any update on this one?

Next week I will try to update it.

dereuromark:
ping @basteyy

basteyy:
> ping @basteyy

Setting up the local test environment needed time, and I'm struggling with a test case. Thanks for pinging, try to modify the patch end of this week.

dereuromark:
Testing should now be easier, see updated docs

PR #1902, opened 2022-08-08
load table maps lazily

### LINKED ISSUE #1901: Call to initDatabaseMaps initializes too many tables.
Hello everyone!

We've noticed that StandardServiceContainer::initDatabaseMaps call takes 10-100 ms and with a regular process execution life cycle, is mostly wasting time.
Regular process uses 1-20 tables, but initializing 300+ ones (for our project) makes very low sense.
Any ideas why it's implemented like this?
Maybe there's a better approach?

Thanks for your thoughts!

Code:
https://github.com/propelorm/Propel2/blob/master/src/Propel/Runtime/ServiceContainer/StandardServiceContainer.php#L261

mringler:
fixes #1901

When adding table classes to a Propel database object, currently those classes are immediately instantiated, event though many of those classes will not be used in one cycle, leading to unnecessary overhead, particularly on databases with lots of tables.

A simple fix is to add tables as class strings, and only instantiate them when they are actually used.

The way it is implemented here, the `$tables` and `$tablesByPhpName` properties of DatabaseMap now can contain both TableMap objects or class strings, which is not ideal, but it seems to be the most seamless integration of this change. Let me know if there is a better way.


src/Propel/Runtime/Map/DatabaseMap.php
```diff
@@ -38,14 +38,14 @@ class DatabaseMap
...
-     * @var array<\Propel\Runtime\Map\TableMap>
+     * @var array<\Propel\Runtime\Map\TableMap|class-string>
      */
     protected $tables = [];
 
     /**
      * Tables in the database, using table phpName as key
      *
-     * @var array<\Propel\Runtime\Map\TableMap>
+     * @var array<\Propel\Runtime\Map\TableMap|class-string>
```
dereuromark, line 48: Should this be `...|class-string<\Propel\Runtime\Map\TableMap>>` ?

src/Propel/Runtime/Map/DatabaseMap.php
```diff
@@ -120,6 +134,41 @@ public function addTableFromMapClass(string $tableMapClass): TableMap
...
+     * Classes added like this will only be instantiated when accessed
+     * through {@link DatabaseMap::getTable()},
+     * {@link DatabaseMap::getTableByPhpName()}, or
+     * {@link DatabaseMap::getTables()}
+     *
+     * @param class-string $tableMapClass The name of the table map to add
+     *
+     * @return void
+     */
+    public function registerTableMapClass($tableMapClass): void
```
dereuromark, line 150: should this have a type?

templates/Builder/Om/tableMapConstants.php
```diff
@@ -13,10 +13,15 @@
      */
     public const TABLE_NAME = '<?php echo $tableName ?>';
 
+    /**
+     * The php name of this class (PascalCase)
```
dereuromark, line 17: nitpick:
```suggestion
     * The PHP name of this class (PascalCase)
```

src/Propel/Runtime/Map/DatabaseMap.php
mringler: Ah, I didn't know this works. Nice!

templates/Builder/Om/tableMapConstants.php
mringler: No, good point

src/Propel/Runtime/Map/DatabaseMap.php
dereuromark: Well, `PHPDoc tag @var has invalid value` 
I guess we also need to adjust the sniffer then to allow this kind of syntax :)

dereuromark: Or is PHPStan not handling this well?
Because the syntax should be working as such

src/Propel/Runtime/Map/DatabaseMap.php
```diff
@@ -143,11 +143,11 @@ public function addTableFromMapClass(string $tableMapClass): TableMap
      * {@link DatabaseMap::getTableByPhpName()}, or
      * {@link DatabaseMap::getTables()}
      *
-     * @param class-string $tableMapClass The name of the table map to add
+     * @param class-string<\Propel\Runtime\Map\TableMap>|string $tableMapClass The name of the table map to add
```
dereuromark, line 146: |string is from sniffer?
We can adjust the sniffer to allow this in a future fix/PR

src/Propel/Runtime/Map/DatabaseMap.php
mringler: I had it as string, but then cs-fix added that into the @param declaration, so I removed it again. Now I have added it again, let me know if you do prefer it the other way,

src/Propel/Runtime/Map/DatabaseMap.php
mringler: Yes, I think that would be nice

src/Propel/Runtime/Map/DatabaseMap.php
mringler: but it seems to work, doesn't it?

src/Propel/Runtime/Map/DatabaseMap.php
dereuromark: See https://github.com/spryker/code-sniffer/commit/93cdaa1257fb7c7d7a425fba943316f77f5d9a4d

dereuromark: Lets keep it for now until we tag a new sniffer release, then we can clean up

mringler: Nice!

profuel:
Local testing with 350 tables:
initDatabaseMaps is executed almost twice faster with this fix. Execution time of the rest of code didn't change

PR #1904, opened 2022-08-22
Fix invalid duplicate indices using ArchivableBehavior for some platforms

### LINKED ISSUE #1894: ArchivableBehavior archive table uses same index name as original table
Hi there,

we're experiencing an issue when using the ArchivableBehavior for tables with indices.
The behavior copies the indices from the original table without modifying the index name. 
This leads to multiple indices with same name. At least in `PostgreSQL` this is not allowed and results in a failing migration:
```
[42P07] ERROR: relation "index-original_table-column" already exists
``` 

One of my colleagues introduced a workaround in our project by setting the copied index name to an empty string.
This way a unique index name will be generated. 

Related code:
https://github.com/propelorm/Propel2/blame/master/src/Propel/Generator/Behavior/Archivable/ArchivableBehavior.php#L129-L132

Workaround:
```
foreach ($table->getIndices() as $index) {
   $copiedIndex = clone $index;
   $copiedIndex->setName('');
    $archiveTable->addIndex($copiedIndex);
}
```

I'd like to get rid of our custom workaround for this issue. 
We should discuss how we can solve this in Propel. 

Some ideas:
- postfix copied indices with `-archive` (e.g. `index-original_table-column` → `index-original_table-column-archive`)
- set empty string as index name so Propel will generate unique index name itself

If we can agree on a solution, I'll try to create a PR for this.

felixgeyer:
Resolves #1894 

Please let me know, if we can write tests for this behavior depending on different platforms.
Without the fix, test setup will fail because of the error described in the linked issue.

To keep it more simple, we could also discuss to never copy the original's index name which might be a breaking change if someone relies on the archive index name.

### DIFF
--- a/src/Propel/Generator/Behavior/Archivable/ArchivableBehavior.php
+++ b/src/Propel/Generator/Behavior/Archivable/ArchivableBehavior.php
@@ -15,6 +15,7 @@ use Propel\Generator\Model\Column;
 use Propel\Generator\Model\Index;
 use Propel\Generator\Model\Table;
 use Propel\Generator\Platform\PgsqlPlatform;
+use Propel\Generator\Platform\PlatformInterface;
 use Propel\Generator\Platform\SqlitePlatform;
 
 /**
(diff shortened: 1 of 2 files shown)


src/Propel/Generator/Behavior/Archivable/ArchivableBehavior.php
```diff
@@ -128,6 +131,11 @@ protected function addArchiveTable(): void
             // copy the indices
             foreach ($table->getIndices() as $index) {
                 $copiedIndex = clone $index;
+                // Some platforms require unique index names across whole database
+                // by unsetting the name, Propel will generate a unique name based on table and columns
+                if ($platform instanceof PgsqlPlatform || $platform instanceof SqlitePlatform) {
```
yaroslav-spryker, line 136: Can we move this check to separate method? At least it will be more clan in terms of code.

yaroslav-spryker (changes requested):

PR #1910, opened 2022-11-02
load databases without accessing classes

mringler:
This MR changes the way tables are registered during initialization to improve performance. Instead of building the internal table lookup structure by accessing static class properties of the table classes, the table lookup is loaded directly from the init file.

#### Background

During initialization, available database tables are registered in Propel. Currently this is done through the qualified class name (i.e. `\Propel\Tests\BookstoreSchemas\Map\BookTableMap`), which is handed to a `DatabaseMap`, which then adds it to its table lookup maps. There are two kinds of lookup, access by name (schema and table name, i.e. `bookstore_schemas.book`) or by PHP name (either set in schema.xml or generated from name, i.e. `\Book`).
Currently, the lookup maps are created by accessing static properties of the supplied class. It was reported that this leads to a measurable slowdown when working with lots of tables (300+), see [here](https://github.com/propelorm/Propel2/pull/1909#issuecomment-1292112844).

#### Proposed Solution
Instead of handing over only the table class name to the DatabaseMap, we can also hand over the name and PHP name, making the static access obsolete. Even simpler, instead of reading the raw table data from the init file (usually `loadDatabase.php`) and then building up the lookup structure every time, we can just write and then retrieve that lookup structure from the init file, and simply load it into the DatabaseMap as it is.
#1909 does something similar, and it reportedly leads to a speed up.

#### Changes
* The `DatabaseMap` class gets a method to dump its lookup maps and a method to load them again from dump.
* The `TableMapLoaderScriptBuilder` class (which generates the `loadDatabase.php` file) does not just write an array of table class names, but registers the tables in a local DatabaseMap, and then writes a dump to the init file.
* During initialization, the dump is loaded back into the DatabaseMaps (through `StandardServiceContainer`).

The changes do not impact BC, even after the update, an old init file can still be used.

#### Result
On my machine, with PHP 8.1 and a project with 40 tables, initialization through static access takes around 0.8 ms (+- 0.3ms), while initialization through dumps takes around 0.25 ms (+- 0.1ms). I am actually surprised how clearly and distinct the difference between the two approaches is. I guess it comes from the time it takes PHP to load the classes, but there is also the repeated calls to a subroutine and writing to the lookup arrays individually. I have tested it through the php dev server, not on an actual webserver, which might give different results.
Of course we are talking about less than a thousand of a second, so this will not make a difference to most users. But it won't hurt either.

I don't have a project with more tables available, but maybe @profuel can supply some measurements for a larger set of tables? I would assume that the static approach grows linear with the number of tables, as each new table leads to another class being loaded, where the dump approach stays mostly the same, as adding another new table just means to read two more lines from a file.

### DIFF
--- a/src/Propel/Runtime/Map/DatabaseMap.php
+++ b/src/Propel/Runtime/Map/DatabaseMap.php
@@ -25,6 +25,11 @@ use Propel\Runtime\Propel;
  * @author Hans Lellelid <hans@xmpl.org> (Propel)
  * @author John D. McNally <jmcnally@collab.net> (Torque)
  * @author Daniel Rall <dlr@collab.net> (Torque)
+ *
+ * @psalm-consistent-constructor (instantiated by class name in StandardServiceContainer without arguments)
+ *
+ * @psalm-type \Propel\Runtime\Map\MapType = 'tablesByName' | 'tablesByPhpName'
+ * @psalm-type \Propel\Runtime\Map\TableMapDump array<\Propel\Runtime\Map\MapType, array<string, class-string<\Propel\Runtime\Map\TableMap>>>
  */
 class DatabaseMap
 {
@@ -38,14 +43,14 @@ class DatabaseMap
     /**
      * Tables in the database, using table name as key
      *
-     * @var array<\Propel\Runtime\Map\TableMap|class-string<\Propel\Runtime\Map\TableMap>>
+     * @var array<string, \Propel\Runtime\Map\TableMap|class-string<\Propel\Runtime\Map\TableMap>>
      */
     protected $tables = [];
 
     /**
      * Tables in the database, using table phpName as key
      *
-     * @var array<\Propel\Runtime\Map\TableMap|class-string<\Propel\Runtime\Map\TableMap>>
+     * @var array<string, \Propel\Runtime\Map\TableMap|class-string<\Propel\Runtime\Map\TableMap>>
      */
     protected $tablesByPhpName = [];
 
@@ -121,7 +126,7 @@ class DatabaseMap
     /**
      * Add a new table to the database, using the tablemap class name.
      *
-     * @param string $tableMapClass The name of the table map to add
+     * @param class-string<\Propel\Runtime\Map\TableMap> $tableMapClass The name of the table map to add
      *
      * @return \Propel\Runtime\Map\TableMap The TableMap object
      */
@@ -137,23 +142,26 @@ class DatabaseMap
     /**
      * Dump table maps. Used during configuration generation.
      *
-     * @return array A dump that can be loaded again with {@link DatabaseMap::loadMapsFromDump()}
+     * @return \Propel\Runtime\Map\TableMapDump A dump that can be loaded again with {@link DatabaseMap::loadMapsFromDump()}
      */
     public function dumpMaps(): array
     {
+        /**
+         * @param $tableMap class-string<\Propel\Runtime\Map\TableMap>|\Propel\Runtime\Map\TableMap
+         * @return class-string<\Propel\Runtime\Map\TableMap>
+         */
         $toClassString = fn ($tableMap) => is_string($tableMap) ? $tableMap : get_class($tableMap);
-        $maps = [
+
+        return [
             'tablesByName' => array_map($toClassString, $this->tables),
             'tablesByPhpName' => array_map($toClassString, $this->tablesByPhpName),
         ];
-
-        return $maps;
     }
 
     /**
      * Load internal table maps from dump. Used during Propel initialization.
      *
-     * @param $mapsDump Table map dump as created by {@link DatabaseMap::dumpMaps()}
+     * @param \Propel\Runtime\Map\TableMapDump|array $mapsDump Table map dump as created by {@link DatabaseMap::dumpMaps()}
      *
      * @return void
      */
@@ -208,7 +216,7 @@ class DatabaseMap
      * Registers a list of table map classes (by qualified name) as table maps
      * belonging to this database.
      *
-     * @param array<class-string> $tableMapClasses
+     * @param array<class-string<\Propel\Runtime\Map\TableMap>> $tableMapClasses
      *
      * @return void
      */
--- a/src/Propel/Runtime/Map/DatabaseMap.php
+++ b/src/Propel/Runtime/Map/DatabaseMap.php
@@ -135,6 +135,35 @@ class DatabaseMap
     }
 
     /**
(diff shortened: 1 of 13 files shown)


mringler:
Psalm fails in an unchanged class, I don't understand the error:
```
ERROR: InvalidClass - src/Propel/Generator/Manager/AbstractManager.php:326:32 - Class, interface or enum XsltProcessor has wrong casing (see https://psalm.dev/007)
                    $xsl = new XsltProcessor();

```

src/Propel/Runtime/Map/DatabaseMap.php
```diff
@@ -134,6 +134,35 @@ public function addTableFromMapClass(string $tableMapClass): TableMap
         return $this->getTable($table->getName());
     }
 
+    /**
+     * Dump table maps. Used during configuration generation.
+     *
+     * @return array A dump that can be loaded again with {@link DatabaseMap::loadMapsFromDump()}
```
dereuromark, line 140:
```suggestion
     * @return array<string, mixed> A dump that can be loaded again with {@link DatabaseMap::loadMapsFromDump()}
```
or even more

src/Propel/Runtime/ServiceContainer/StandardServiceContainer.php
```diff
@@ -270,6 +270,23 @@ public function initDatabaseMaps(array $databaseNameToTableMapClassNames = []):
         }
     }
 
+    /**
+     * @param array $databaseNameToTableMapDumps
```
dereuromark, line 274: same

src/Propel/Runtime/Map/DatabaseMap.php
mringler: Hey Mark, thank you for the review. Played around with psalm a bit, let me know what you think

src/Propel/Runtime/Map/DatabaseMap.php
```diff
@@ -134,6 +139,38 @@ public function addTableFromMapClass(string $tableMapClass): TableMap
         return $this->getTable($table->getName());
     }
 
+    /**
+     * Dump table maps. Used during configuration generation.
+     *
+     * @return \Propel\Runtime\Map\TableMapDump A dump that can be loaded again with {@link DatabaseMap::loadMapsFromDump()}
```
dereuromark, line 145: This looks like array<mixed, ..> instead of a object return though.

mringler: Yes, good that you mention it, I am not sure how much I like that. Apparently psalm allows type aliases, which `\Propel\Runtime\Map\TableMapDump` is, I put the declaration in the docblock of the class:
```php
* @psalm-type \Propel\Runtime\Map\MapType = 'tablesByName' | 'tablesByPhpName'
 * @psalm-type \Propel\Runtime\Map\TableMapDump array<\Propel\Runtime\Map\MapType, array<string, class-string<\Propel\Runtime\Map\TableMap>>>
 ```
On one hand, it is nice, because we can make sure that the load and dump methods always use the exact same type, and through an import with `@psalm-import-type`, it can be used in the `StandardServiceContainer` too. So I think in general this is the "correct" approach.
On the other hand, a lot about it seems confusing. I had to give the alias a name that looks like a namespaced type (otherwise it wouldn't work with phpcs, which changed "TableMapDump" to "\Propel\Runtime\Map\TableMapDump", causing a no such type error). Also IDE support for type aliases is not (yet) what it could be, at least in VS Code and Eclipse. Overall, I have the impression that type aliases as a concept is still rather new in PHP, which may or may not keep us from using it. What do you think?

Comment (dereuromark) [reply src/Propel/Runtime/Map/DatabaseMap.php:145]: I think the main docblock (return) should reflect trivial reality (array)
The on top specific phpstan and psalm tags can contain the "new" concepts.

Comment (mringler) [reply src/Propel/Runtime/Map/DatabaseMap.php:145]: Like this then?
```php
    /**
     * Dump table maps. Used during configuration generation.
     *
     * @psalm-return \Propel\Runtime\Map\TableMapDump
     *
     * @return array A dump that can be loaded again with {@link DatabaseMap::loadMapsFromDump()}
     */
```

profuel:
I like the idea, it looks like a next step to optimize loading of a tables.
I cannot promise fast test, but in a week - I will provide you with results.

profuel:
I've tested this locally.
It has 0.5-2 ms `wasted` time of every request which is simply awesome, comparing to the original 10+ ms.

So I'm fully for this solution.

BTW, \Propel\Runtime\ServiceContainer\StandardServiceContainer and especially setConnectionManager has a major change, which lead to a minor change on our project side :)

tests/Propel/Tests/Runtime/Map/DatabaseMapTest.php
```diff
@@ -101,27 +101,38 @@ public function testAddTableFromMapClass()
...
-        };
+        $databaseMap = new DatabaseMapWithGetters('dummyDatabase');
         
         $tableMapClass = BazTableMap::class;
         $databaseMap->registerTableMapClass($tableMapClass);
         
+        $this->assertTableMapContains($databaseMap, $tableMapClass::TABLE_NAME, '\\' . $tableMapClass::TABLE_PHP_NAME, $tableMapClass);
+    }
+
+    protected function assertTableMapContains(DatabaseMapWithGetters $databaseMap, $tableName, $phpTableName, $tableMapClass){
```
dereuromark, line 131:
```suggestion
    protected function assertTableMapContains(DatabaseMapWithGetters $databaseMap, $tableName, $phpTableName, $tableMapClass)
    {
```

I wonder if CS is not active for tests.

dereuromark:
@profuel I dont see any changes to setConnectionManager() etc, can you clarify the major changes that have implications for you?

profuel:
@dereuromark my fault, we have an older Propel on the project. All good here.

dereuromark (approved):
Given the analyis/QA I will now approve this
With a 2nd approve we have green light for merge.

tests/Propel/Tests/Runtime/Map/DatabaseMapTest.php
mringler: Yes, tests are currently not linted

mringler:
Thank you @profuel for the tests!

profuel:
@mringler my pleasure. We need this change on the project, so - mutual goal :)

dereuromark:
Any other approvers? I am only waiting for this for the PR to get merged.

yaroslav-spryker (approved):

Issue #1914, opened 2022-11-15
UUID type support


dereuromark:
Based on https://github.com/propelorm/Propel2/issues/1498

The main idea would be to add UUID as natively supported with fallback to binary(16) or alike where needed.
So all you need to do is to specify `uuid` as native type and it should use the correct/best way as per DB type.

Does that make sense?
Would this be the way forward instead of defining it more manually?
The more manual way would probably have to be to support char(36), binary(16) and alike as low level definitions.

If we go with `uuid` and native support as per DB type:
- Postgres supports it as native type (129bit) directly since v8 afaik
- Mysql can use binary(16) under the hood, just as other ORMs do - e.g. [here](https://github.com/cakephp/cakephp/blob/4.x/src/Database/Type/BinaryUuidType.php)
- ?

And feedback/ideas welcome.

mringler:
Oh, this is quite interesting. Some thoughts:

Implementing support for DB systems with native uuid type should be easy and can probably be done quickly. I think those are Postgres, Oracle and SQL Server (as "uniqueidentifier").

Adding support for systems without native uuid type (MySQL/MariaDB, SQLite) through `binary(16)` or `blob` might be tricky though. The hard part is figuring out where conversion between strings and binary type should happen. We would need a mechanism where vendor-specific data conversion is applied to input and output data, which does not exist at the moment afaik.

Looking at how query execution is done in Propel, our best option is probably to do it inside the SQL Builder and in the Formatter:
 
```mermaid
flowchart LR
Z(Input) --> B
A[Model] --> |Create\nUpdate\nDelete| B[Query/Criteria]
B --> C[SQL Builder]
C --> D(DB)
D --> E[Fetcher]
E --> F[Formatter]
F --> G[Model]
F --> H[Array]
F --> I[...]
linkStyle 0 stroke-dasharray: 5 5;
linkStyle 3 stroke-dasharray: 5 5;
linkStyle 4 stroke-dasharray: 5 5;
classDef highlightNode stroke:yellow;
class C,F highlightNode;
```
In the SQL Builder, it should be rather easy to figure out which values of `INSERT` and `UPDATE` queries need to be converted. Figuring out which values to convert in filter statements (as in `WHERE uuid_column IN (...)`) can be done in the condition classes (`Criterion`), which already have access to the vendor adapter (I seem to remember that Criterion creation does not always work as expected though).

[...]

PR #1915, opened 2022-11-17
Add support for native UUID column type

mringler:
As discussed in #1914, this allows to use `UUID` as column type in schema.xml. Underlying columns are set to the native uuid type of the database system (`uuid` in Postgres and Oracle, `UNIQUEIDENTIFIER` in mssql). In Mysql/MariaDB and SQLite, an error is thrown. Inside the model and query classes, the uuid columns use string values.

This is pretty straight forward, a new type literal is registered and associated with strings, and register the type in the vendor adapters with the corresponding type. For Postgres, I added cast statements used during migration. Tests check generated `create` and `alter` statements and internal mapping of uuid to string.

I have played around with it on Postgres, and it seems to work there.

I also added a commit with two fixes for the test setup scripts:
- `setup.mysql.sh` always used "test" as database name, even though the fixtures are build using an environment variable (`$DB_NAME`) with "test" only as fallback
- in `setup.pgsql.sh` some statements failed when a variable was empty, leading to empty string parameters

### DIFF
--- a/src/Propel/Generator/Model/PropelTypes.php
+++ b/src/Propel/Generator/Model/PropelTypes.php
@@ -342,6 +342,16 @@ class PropelTypes
     public const JSON_TYPE = 'string';
 
     /**
+     * @var string
+     */
+    public const UUID = 'UUID';
+
+    /**
+     * @var string
+     */
+    public const UUID_NATIVE_TYPE = 'string';
+
+    /**
      * Propel mapping types.
      *
      * @var array
@@ -381,6 +391,7 @@ class PropelTypes
         self::BU_TIMESTAMP,
         self::SET,
         self::JSON,
+        self::UUID,
     ];
 
     /**
@@ -421,6 +432,7 @@ class PropelTypes
         self::SET => self::SET_NATIVE_TYPE,
         self::GEOMETRY => self::GEOMETRY,
         self::JSON => self::JSON_TYPE,
+        self::UUID => self::UUID_NATIVE_TYPE,
     ];
 
     /**
@@ -466,6 +478,7 @@ class PropelTypes
         self::BU_DATE => PDO::PARAM_STR,
         self::BU_TIMESTAMP => PDO::PARAM_STR,
         self::JSON => PDO::PARAM_STR,
+        self::UUID => PDO::PARAM_STR,
     ];
 
     /**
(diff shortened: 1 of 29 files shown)


src/Propel/Generator/Model/PropelTypes.php
```diff
@@ -615,6 +628,20 @@ public static function isLobType(string $mappingType): bool
...
+     * Returns whether the given type is a UUID type.
+     *
+     * @param string $type
+     *
+     * @return bool
+     */
+    public static function isUuidType(string $type): bool
+    {
+        return in_array($type, [
+            self::UUID,
```
dereuromark, line 641: we can set third arg true here for strictness

src/Propel/Generator/Platform/PgsqlPlatform.php
```diff
@@ -730,6 +728,18 @@ public function getModifyColumnDDL(ColumnDiff $columnDiff): string
...
+    /**
+     * @param string $type
+     *
+     * @return bool
+     */
+    public function isUuid(string $type): bool
+    {
+        $strings = ['UUID'];
+
+        return in_array(strtoupper($type), $strings);
```
dereuromark, line 740: same

mringler:
>Do we have a way to update docs

Not sure, the only doc for Propel 2 that I know of is at http://propelorm.org/documentation/, but I am not really familiar with it

dereuromark:
Right, we probably want to add UUID info at https://github.com/propelorm/propelorm.github.com then to be published after we merged.

src/Propel/Runtime/Map/ColumnMap.php
```diff
@@ -294,6 +294,16 @@ public function isText(): bool
         ]);
     }
 
+    /**
+     * Whether this column contains UUIDs.
+     *
+     * @return bool
+     */
+    public function isUUID(): bool
```
limeeugenia, line 302: `isUuid` for consistency.

src/Propel/Generator/Platform/PgsqlPlatform.php
```diff
@@ -782,16 +788,18 @@ public function getUsingCast(Column $fromColumn, Column $toColumn): string
             return " USING decode(CAST($name as text), 'escape')";
         }
 
-        if ($fromSqlType === 'DATE' && $toSqlType === 'TIME') {
-            return ' USING NULL';
-        }
```
limeeugenia, line 787: This DATE and TIME handling is dropped, why?

src/Propel/Runtime/Map/ColumnMap.php
mringler: That's right, thank you!

src/Propel/Generator/Platform/PgsqlPlatform.php
mringler: The function returns ` USING NULL` per default, so this was just an extra step to get to the same result.

dereuromark:
Small conflict then we should be good to go

mringler:
Oh, hadn't realized, thank you for the notice!

limeeugenia (approved):

PR #1917, opened 2022-11-24
Add type to store UUIDs as binary (UUID_BINARY)

mringler:
This allows Propel to work with UUIDs stored in a binary data column like `BINARY(16)` in MySQL.

To use it, the column type in `schema.xml` has to be set to `UUID_BINARY`:
```xml
    <table name="my_table">
        <column name="uuid" required="false" type="UUID_BINARY"/>
    </table>
```

In models, fields of that type will always contain the UUID as string, conversion between binary value and string is done during loading (in `hydrate()`) or saving (in `doInsert()` or `buildCriteria()` from `doUpdate()`). Similar conversion happens in the `findBy` methods in the query class. 
This is a different approach than planned in #1914, but I found it integrates better into Propel.

Per default, UUID conversion will swap some bytes around, in accordance with MySQL's [`uuid_to_bin()`](https://dev.mysql.com/doc/refman/8.0/en/miscellaneous-functions.html#function_uuid-to-bin) function. This allows better indexing for version-1 UUIDs. The default behavior can be changed in `schema.xml` by setting the value of `UuidSwapFlag` in vendor information:
```xml
<?xml version="1.0" encoding="ISO-8859-1" standalone="no"?>
<database name="bookstore" defaultIdMethod="native"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    namespace="Propel\Tests\Bookstore"
>
    <vendor type="mysql">
          <parameter name="UuidSwapFlag" value="false"/>
    </vendor>
```

Caveats:
- Feature is in experimental state. It is tested, but only to an extend.
- No migration of columns with data at the moment (generated "ALTER TABLE" statements will fail).
- Changing `UuidSwapFlag` currently does not trigger a migration at all (it should cause all existing UUIDs to be (un)swapped).
- Going around the model classes, for example by using `select()` on the query class or manually setting a formatter like `SimpleArrayFormatter`, will leave UUIDs in their binary form and users have to convert them manually by calling `UuidConverter::binToUuid($uuidData, $swapFlag)` (depending on PDO behavior for the underlying dbms, they will have to read the resource/stream into a string first by calling `stream_get_contents($uuidData)`).
- Similarly, if for some reason the query class fails to detect UUID input and does not automatically convert it into the binary form, users will have to do it manually by calling `UuidConverter::uuidToBin($uuid, $swapFlag)`.

Realistically, this will take a brave user to test and give feedback.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,33 +24,40 @@ jobs:
                     php-version: '7.4'
                   - symfony-version: '6-max'
                     php-version: '7.4'
+        env:
+            DB_NAME: 'propel_tests'
+            DB_USER: 'propel'
+            DB_PW: 'propel'
         steps:
             - name: Install PostgreSQL latest
-              if: matrix.db-type == 'pgsql' && matrix.php-version != '7.3'
+              if: matrix.db-type == 'pgsql' && matrix.php-version != '7.4'
               uses: CasperWA/postgresql-action@v1.2
               with:
-                  postgresql db: 'propel-tests'
-                  postgresql user: 'postgres'
-                  postgresql password: 'postgres'
+                  postgresql db: $DB_NAME
+                  postgresql user: $DB_USER
+                  postgresql password: $DB_PW
 
             - name: Install PostgreSQL min
-              if: matrix.db-type == 'pgsql' && matrix.php-version == '7.3'
+              if: matrix.db-type == 'pgsql' && matrix.php-version == '7.4'
               uses: CasperWA/postgresql-action@v1.2
               with:
                   postgresql version: 9
-                  postgresql db: 'propel-tests'
-                  postgresql user: 'postgres'
-                  postgresql password: 'postgres'
+                  postgresql db: $DB_NAME
+                  postgresql user: $DB_USER
+                  postgresql password: $DB_PW
 
-            - name: Install MariaDb latest
-              if: matrix.db-type == 'mysql' && matrix.php-version != '7.3'
-              uses: getong/mariadb-action@v1.1
+            - name: Install MySQL latest
+              if: matrix.db-type == 'mysql' && matrix.php-version != '7.4'
+              uses: mirromutth/mysql-action@v1.1
+              with:
+                  mysql root password: $DB_PW
 
             - name: Install MariaDb min
-              if: matrix.db-type == 'mysql' && matrix.php-version == '7.3'
+              if: matrix.db-type == 'mysql' && matrix.php-version == '7.4'
               uses: getong/mariadb-action@v1.1
               with:
                   mariadb version: '10.2'
+                  mysql root password: $DB_PW
 
             - name: Setup PHP, with composer and extensions
               uses: shivammathur/setup-php@v2
@@ -79,34 +86,39 @@ jobs:
             - name: Composer install (Symfony version ${{ matrix.symfony-version }})
               run: composer install --no-progress --prefer-dist --optimize-autoloader
 
-            - name: Setup Postgresql database for test suite
-              if: matrix.db-type == 'pgsql'
-              run: tests/bin/setup.pgsql.sh
+            - name: Wait for MySQL server to load
+              if: matrix.db-type == 'mysql'
+              run: |
+                bash -c "
+                for i in {1..10}; do
+                  mysqladmin -h 127.0.0.1 -u root status >/dev/null 2>&1 && exit 0 || sleep 6
+                  echo 'trying again'
+                done;
+                echo 'could not establish connection after 10 tries'
+                exit 1
+                "
               env:
-                  DB_NAME: 'propel-tests'
-                  DB_USER: 'postgres'
-                  DB_PW: 'postgres'
+                MYSQL_PWD: ${{ env.DB_PW }}
 
-            - name: Setup the database for test suite
-              if: matrix.db-type != 'agnostic' && matrix.db-type != 'pgsql'
+            - name: Create MySQL Propel user
+              if: matrix.db-type == 'mysql'
+              run: |
+                mysql -h 127.0.0.1 -u root -e "
+                  CREATE USER '$DB_USER'@'localhost' IDENTIFIED BY '$DB_PW';
+                  CREATE USER '$DB_USER'@'%' IDENTIFIED BY '$DB_PW';
+                  GRANT ALL PRIVILEGES ON *.* TO '$DB_USER'@'localhost';
+                  GRANT ALL PRIVILEGES ON *.* TO '$DB_USER'@'%';
+                  FLUSH PRIVILEGES;
+                "
+              env:
+                MYSQL_PWD: ${{ env.DB_PW }}
+
+            - name: Setup database for test suite
+              if: matrix.db-type != 'agnostic'
               run: tests/bin/setup.${{ matrix.db-type }}.sh
 
-            - name: Run PostgreSQL tests
-              if: matrix.db-type == 'pgsql'
-              shell: 'script -q -e -c "bash {0}"'
-              run: |
-                  if [[ ${{ matrix.php-version }} == '7.4' && ${{ matrix.symfony-version == '5-max' }} ]]; then
-                    export CODECOVERAGE=1 && vendor/bin/phpunit -c tests/pgsql.phpunit.xml --verbose --coverage-clover=tests/coverage.xml
-                  else
-                    vendor/bin/phpunit -c tests/pgsql.phpunit.xml
-                  fi
-              env:
-                  DB_NAME: 'propel-tests'
-                  DB_USER: 'postgres'
-                  DB_PW: 'postgres'
-
-            - name: Run ${{ matrix.db-type }} tests
-              if: matrix.db-type != 'pgsql'
+            - name: Run database tests
+              if: matrix.db-type != 'agnostic'
               shell: 'script -q -e -c "bash {0}"'
               run: |
                   if [[ ${{ matrix.php-version }} == '7.4' && ${{ matrix.symfony-version == '5-max' }} ]]; then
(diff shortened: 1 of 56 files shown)


dereuromark:
Great work!

> type="UUID_BINARY"

vs https://github.com/propelorm/Propel2/pull/1915 and `type="UUID"`

My main question would be: Do need to have different type names per DB?
From a user perspective the internals (and custom decisions on the specific sub type) are not so interesting usually, so lets say you want to use UUID from a framework perspective and support 3+ DB types
The ORM ideally abstracts the internals away, that you can specify `UUID` across your project definitions then and they dont have to be different per DB type.

Is that something we can achieve?

mringler:
Ok, now `UUID_BINARY` is used as fallback type for `UUID` in MySQL and SQLite (all other vendors have a native UUID type). So now the `UUID` type can always be used (even though the result might be surprising). 

I think we should keep both types, as it allows to still use `UUID_BINARY' on systems where `UUID` maps to the native type. This might be useful if an old version of the dbms is used or if the db already uses binary columns and cannot be changed. Also, if you know your database will use `UUID_BINARY` anyway, you might find it cleaner to just specify that. And should MySQL ever support a native UUID type and we change the behavior of `UUID`, users can switch to `UUID_BINARY` if they want to avoid migrating their db. 

@dereuromark what do you think does that make sense?

mringler:
I have added code for automatic migration between `CHAR`/`VARCHAR` and `BINARY` uuid columns on MySQL. If column type is changed in schema.xml, Propel will add code to the migration file, which creates a new column for the converted values and then replaces the old column with that new one.

When changing from a binary column to a `CHAR` type, propel does not know that the column contains uuids, it just sees binary to char. To let Propel know that it is a conversion of uuids, the column has to be marked as such by adding the `content="UUID"` to the declaration in schema.xml (we can use something else if that is better):
```xml
<column name="id" type="varchar" size="36" content="UUID" />
```

I had to change some overall behavior to get this working, my assumption is that these changes are beneficiary:
- In MySQL schema parser, binary types were mapped to BLOBs, now they map to the binary type. 
- When detecting schema changes, a foreign key is considered to have changed when the type of one of the involved columns change (leads to dropping and recreating the Fk constraint).
- Propel combines `ALTER TABLE` statements in migrations, I have moved that code into its own class (AlterTableStatementMerger) and you can add a statement, which tells the merger not to combine statements above with those below (AlterTableStatementMerger::NO_MERGE_ACROSS_THIS_LINE). Also the merger recognizes DML statements and will not merge across them.
- I refactored some statements I ran into because I found them hard to read. I try to not do that, as it tends to obfuscates commits, but didn't want to remove them either.
- The tests only run on MySQL, not MariaDB, so they would not have run on Github. But I do want those tested not just locally, so I changed ci to use MySQL instead of MariaDB.

With the test, I am becoming more confident that this might actually work. Let me know what you think!

src/Propel/Generator/Platform/MysqlPlatform.php
```diff
@@ -53,19 +54,22 @@ class MysqlPlatform extends DefaultPlatform
...
     {
         parent::initialize();
-        unset($this->schemaDomainMap[PropelTypes::UUID]);
         $this->setSchemaDomainMapping(new Domain(PropelTypes::BOOLEAN, 'TINYINT', 1));
         $this->setSchemaDomainMapping(new Domain(PropelTypes::NUMERIC, 'DECIMAL'));
         $this->setSchemaDomainMapping(new Domain(PropelTypes::LONGVARCHAR, 'TEXT'));
-        $this->setSchemaDomainMapping(new Domain(PropelTypes::BINARY, 'BLOB'));
-        $this->setSchemaDomainMapping(new Domain(PropelTypes::VARBINARY, 'MEDIUMBLOB'));
+        $this->setSchemaDomainMapping(new Domain(PropelTypes::BINARY, 'BINARY'));
+        $this->setSchemaDomainMapping(new Domain(PropelTypes::VARBINARY, 'VARBINARY'));
```
asmarovydlo, line 61: Isn't this a BC break, in the case of `MEDIUMBLOB `=> `VARBINARY` because of size?

asmarovydlo (commented):
Great work!

tests/Propel/Tests/Runtime/Adapter/Pdo/OracleAdapterTest.php
```diff
@@ -31,19 +31,23 @@ protected function getDriver()
     {
         return 'oracle';
     }
+    
+    protected function createOracleSql(Criteria $query, &$params = []): string
```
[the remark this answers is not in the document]
mringler: Thank you for the review!
I removed that parameter.

.github/workflows/ci.yml
```diff
@@ -24,33 +24,40 @@ jobs:
...
+                  postgresql password: $DB_PW
 
-            - name: Install MariaDb latest
-              if: matrix.db-type == 'mysql' && matrix.php-version != '7.3'
-              uses: getong/mariadb-action@v1.1
+            - name: Install MySQL latest
+              if: matrix.db-type == 'mysql' && matrix.php-version != '7.4'
+              uses: mirromutth/mysql-action@v1.1
+              with:
+                  mysql root password: $DB_PW
```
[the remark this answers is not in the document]
mringler: Oh, I fought with that quite a bit. The problem is that the setup script drops and creates schemas, which the regular user cannot do. And the docker container does not allow to create more than one table, and the root user is not accessible if a regular user is provided (random password), so we cannot alter things or change permissions..
At the same time, the setup script cannot be started as a different user than the one who executes the tests.
The workaround I ended up on is to add the propel user later in a later step and grant him permissions to create and drop (all really). Let me know if you see a better solution.

tests/Propel/Tests/Runtime/Util/UuidConverterMysqlCompatibilityTest.php
```diff
@@ -0,0 +1,76 @@
...
+            ? hex2bin('aab5d5fd70c111e5a4fbb026b977eb28')
+            : 'aab5d5fd-70c1-11e5-a4fb-b026b977eb28'
+            ;
+        $mysqlBin = $this->executeStatement($sqlStatement, $value);
+
+        $propelBin = $callback($value);
+        $this->assertSame($mysqlBin, $propelBin, $description . ' should match between Propel and MySQL');
+    }
+
+    public function executeStatement(string $statement, string $value)
```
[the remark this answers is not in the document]
mringler: Yes, thank you!

tests/Propel/Tests/Generator/Migration/MysqlMigrateUuidColumnTest.php
```diff
@@ -0,0 +1,311 @@
...
+    use CheckMysql8Trait;
+
+    /**
+     * @return void
+     */
+    public function setUp(): void
+    {
+        parent::setUp();
+        if (!$this->checkMysqlVersionAtLeast8('migration')) {
+            $this->markTestSkipped('Test can only be run on MySQL version >= 8');
```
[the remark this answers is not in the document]
mringler: Problem is that the functions `uuid_to_bin` and `bin_to_uuid` are not available in MySQL < 8 and MariaDB at all, but those are used in the migration. So this just does not work with them.

src/Propel/Generator/Platform/Util/MysqlUuidMigrationBuilder.php
```diff
@@ -0,0 +1,178 @@
...
+use Propel\Generator\Model\Column;
+use Propel\Generator\Model\Index;
+use Propel\Generator\Platform\MysqlPlatform;
+
+/**
+ * Creates migration statements for UUID columns in MySQL.
+ *
+ * @phpstan-consistent-constructor
+ */
+class MysqlUuidMigrationBuilder
```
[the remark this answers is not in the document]
mringler: Yes, I'd like that too.  But since the builtin function to convert UUIDs is only available in MySQL, other platforms, cannot use a pure SQL migration. I think we would have do something like creating the new column, read all data into PHP and then fill the new column with an INSERT, before we can replace the old column. Or write function declarations for the other platforms. Both is a bit too much for now. Maybe someday...

src/Propel/Generator/Platform/MysqlPlatform.php
mringler: Yes, that might be a problem for some users. But I don't see another way around it, those declarations are not accurate anymore, and we need them to be (this produced very hard to find errors).
Should be easy to fix though, they just have to change the type in their schema.xml. But yes, the whole thing is a bit unfortunate.

asmarovydlo: Thanks for the explanation!
What happens if customers have already up and running DB with `VARBINARY => BLOB` and we'll switch to `VARBINARY => VARBINARY`.
As far as I can see from the code, there will be no migration applied and they will continue using `BLOB` as before in the DB.

dereuromark: I wonder if we could seperate those two things, merging this PR with the UUID changes, and further setting the "cleanup" to a separate PR and discussion?
Maybe that helps to also make review easier.

mringler: Hmm, I probably can undo it for the varbinary type. The binary type (currently mapping to blob) is central to working with UUIDs, I cannot just leave it out. Adding a mumbo jumbo alias is not a path a want to start heading into. Should I remove migration and tests for varbinary?

dereuromark: Sounds fair

mringler: Alright, I removed those changes again. I thought I was doing more with varbinary, but it was really just the mapping.

dereuromark (approved):
Lets put cleanup into follow up PR
Nice work

asmarovydlo (approved):

PR #1920, opened 2022-12-08
Create IN filter expressions for subqueries

mringler:
This allows to create IN queries like `WHERE column IN (SELECT otherColumn FROM ...)`.

It can be done by using the `filterBy()` methods:
```php
$innerQuery = BookQuery::create()
  ->filterByTitle('%Galaxy%', Criteria::LIKE)
  ->select('author_id');
AuthorQuery::create()
  ->filterById($innerQuery)
  ->find();
```
which will generate a statement like:
```sql 
SELECT * 
FROM author 
WHERE id IN (
  SELECT author_id
  FROM book
  WHERE book.title LIKE '%Galaxy%'
)
```

For known relations, there are `useInQuery()` methods which work similar as the `useQuery()` methods:
```php
AuthorQuery::create()
  ->useInQuery('Book')
    ->filterByTitle('%Galaxy%', Criteria::LIKE)
  ->endUse()
  ->find();
```

Propel creates convenience methods for existing relations on query classes like `AuthorQuery::useInBookQuery()`, which is easier to use than ` useInQuery('Book')`.

`NOT IN` queries are created by passing `CRITERIA::NOT_IN` to `filterBy()` or the `useNotInQuery()` method.

I have build a similar thing for `EXISTS`, so this is mostly just abstracting from that.


src/Propel/Runtime/ActiveQuery/Criterion/AbstractInnerQueryCriterion.php
```diff
@@ -0,0 +1,149 @@
...
+ * Abstract Criterion for nested filter expression, that bind an inner query
+ * with an operator like IN, EXISTS
+ *
+ * @phpstan-consistent-constructor
+ */
+abstract class AbstractInnerQueryCriterion extends AbstractCriterion
+{
+ /**
+  * @var string|null Left side of the operator, can be empty.
+  */
```
dereuromark, line 27: we might have some cs issues here

mringler: What do you mean? I don't understand

dereuromark: The indentation seems off

mringler: Right, good catch! My editor does that sometimes. Weird that the sniffer doesn't fix it.

src/Propel/Runtime/ActiveQuery/Criterion/AbstractInnerQueryCriterion.php
```diff
@@ -0,0 +1,149 @@
...
+     *
+     * @param \Propel\Runtime\ActiveQuery\ModelCriteria $outerQuery
+     * @param \Propel\Runtime\Map\RelationMap $relation
+     *
+     * @return void
+     */
+    protected function initForRelation(ModelCriteria $outerQuery, RelationMap $relation): void
+    {
+        // does nothing per default
+    }
```
asmarovydlo, line 108: Maybe we could define it `abstract` instead?

asmarovydlo (commented):
Good job 👍

src/Propel/Runtime/ActiveQuery/Criteria.php
```diff
@@ -723,13 +724,23 @@ public function getNewCriterion(string $column, $value = null, $comparison = sel
             // something like $c->add('foo like ?', '%bar%', PDO::PARAM_STR);
             return new RawCriterion($this, $column, $value, $comparison);
         }
+        if ($value instanceof Criteria && $comparison === self::EQUAL) {
+            $comparison = InQueryCriterion::IN;
+        }
```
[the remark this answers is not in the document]
mringler: Changed it so that `$comparison` defaults to `null`, now we can change it to `IN` only when no operator was given. That way, if `WHERE col = (SELECT...)` is ever implemented (currently causes an exception), it can still be built by passing in an explicit `EQUAL` operator along with the query.

I don't think it would make sense to generally use `=` as the default operator when a query is given, because from what I know, `WHERE col = (SELECT...)` is just a special case of `IN` where the subquery returns exactly one element. So while both give you the same result, `=` does not always work, while `IN` does.
And if no operator is given, I think we should pick the option that always works, not the special case.
What do you think?

src/Propel/Runtime/ActiveQuery/Criterion/ExistsCriterion.php
```diff
@@ -28,79 +28,41 @@ class ExistsCriterion extends AbstractCriterion
...
      *
      * @param \Propel\Runtime\ActiveQuery\ModelCriteria $outerQuery
-     * @param \Propel\Runtime\ActiveQuery\ModelCriteria $existsQuery
-     * @param string|null $typeOfExists Either ExistsCriterion::TYPE_EXISTS or ExistsCriterion::TYPE_NOT_EXISTS
-     * @param \Propel\Runtime\Map\RelationMap|null $relationMap where outer query is on the left side
+     * @param \Propel\Runtime\Map\RelationMap $relation
+     *
+     * @return void
      */
-    public function __construct(
```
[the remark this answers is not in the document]
mringler: Ok, this class is now called `ExistsQueryCriterion` and the old ExistsCriterion extends it with the old constructor for BC.

src/Propel/Runtime/ActiveQuery/Criterion/AbstractInnerQueryCriterion.php
mringler: Right, thank you, didn't realize that I didn't need a default anymore.

src/Propel/Runtime/ActiveQuery/Criteria.php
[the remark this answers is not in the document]
mringler: I have generalized the IN query, so that `filterBy()` accepts a query as value along with any operator. So now it is possible to build statements like `WHERE col = (SELECT...)`

asmarovydlo (approved):

tests/Propel/Tests/Generator/Builder/Om/QueryBuilderTest.php
```diff
@@ -1312,7 +1312,7 @@ public function testUseRelationExistsQuery()
     public function testUseRelationNotExistsQuery()
     {
         $expected = BookQuery::create()
-        ->useExistsQuery('Author', null, null, ExistsCriterion::TYPE_NOT_EXISTS)
+        ->useExistsQuery('Author', null, null, ExistsQueryCriterion::TYPE_NOT_EXISTS)
```
asmarovydlo, line 1315: [Minor] IMO better not to change the test, to make sure that it still works with child class `ExistsCriterion` instead of `ExistsQueryCriterion`
Same story with https://github.com/propelorm/Propel2/pull/1920/files#diff-9dd69f4d7d65724ce303aaa7434eec44b30f936b56f94fb9afe3d1bd6df748e5

mringler: Ok, I changed those back.
Thank you for the review! That was very helpful.

mringler:
PR for documentation is at propelorm/propelorm.github.com#429

tests/Propel/Tests/Generator/Builder/Om/QueryBuilderTest.php
asmarovydlo: Thanks for the PR, it very useful improvement!

PR #1924, opened 2022-12-11
UUID adjustments for mariadb

mringler:
Allows to use native uuid type on MariaDB by enabling it in the propel configuration:
```yaml
propel:
    adapters:
      mysql:
        uuidColumnType: native # <------ here
```

While the actual change is quite small, I had to went through a lot of places to find a suitable entry point. I had to refactor some methods along the way, and I think it is best to keep these changes. I have put them in different commits if this helps reviewing!

Also, this includes #1922 , agnostic tests are running

### DIFF
--- a/src/Propel/Generator/Platform/MysqlPlatform.php
+++ b/src/Propel/Generator/Platform/MysqlPlatform.php
@@ -89,8 +89,6 @@ class MysqlPlatform extends DefaultPlatform
     }
 
     /**
-     * Setter for the tableEngineKeyword property
-     *
      * @param string $tableEngineKeyword
      *
      * @return void
@@ -101,8 +99,6 @@ class MysqlPlatform extends DefaultPlatform
     }
 
     /**
-     * Getter for the tableEngineKeyword property
-     *
      * @return string
      */
     public function getTableEngineKeyword(): string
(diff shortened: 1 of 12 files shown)


src/Propel/Generator/Platform/MysqlPlatform.php
```diff
@@ -80,17 +84,49 @@ protected function initialize(): void
...
+
+        if ($defaultTableEngine = $mysqlConfig['tableType']) {
             $this->defaultTableEngine = $defaultTableEngine;
         }
-        if ($tableEngineKeyword = $generatorConfig->get()['database']['adapters']['mysql']['tableEngineKeyword']) {
+        if ($tableEngineKeyword = $mysqlConfig['tableEngineKeyword']) {
             $this->tableEngineKeyword = $tableEngineKeyword;
         }
+        if ($uuidColumnType = $mysqlConfig['uuidColumnType']) {
+            $enable = strtolower($uuidColumnType) === 'native';
```
asmarovydlo, line 97: [Minor] Wouldn't be better to have a standalone platform for MariaDb that extends a MySQL one?

asmarovydlo (approved):

src/Propel/Generator/Platform/MysqlPlatform.php
mringler: Yes, that would be great, but that is its own PR

PR #1932, opened 2022-12-27
Implemented migration rollback.

romanhavrylko:
- Adjusted `MigrationMigrateCommand::execute()` with providing a new option `migrate-to-version` which allows defining the version to migrate database.
- Adjusted `MigrationStatusCommand::execute()` with providing a new option `last-version` which allows receiving the last executed version of the migration.

### DIFF
--- a/src/Propel/Generator/Command/Executor/RollbackExecutor.php
+++ b/src/Propel/Generator/Command/Executor/RollbackExecutor.php
@@ -159,7 +159,7 @@ class RollbackExecutor
         foreach ($statements as $statement) {
             try {
                 if ($this->input->getOption(static::COMMAND_OPTION_VERBOSE)) {
-                    $this->output->writeln(sprintf('Executing statement "%s"', $statement));
+                    $this->output->writeln(sprintf('Executing statement `%s`', $statement));
                 }
 
                 $conn->exec($statement);
@@ -168,11 +168,11 @@ class RollbackExecutor
                 if ($this->input->getOption(static::COMMAND_OPTION_FORCE)) {
                     //continue, but print error message
                     $this->output->writeln(
-                        sprintf('<error>Failed to execute SQL "%s". Continue migration.</error>', $statement),
+                        sprintf('<error>Failed to execute SQL `%s`. Continue migration.</error>', $statement),
                     );
                 } else {
                     throw new RuntimeException(
-                        sprintf('<error>Failed to execute SQL "%s". Aborting migration.</error>', $statement),
+                        sprintf('<error>Failed to execute SQL `%s`. Aborting migration.</error>', $statement),
                         0,
                         $e,
                     );
--- a/src/Propel/Generator/Command/Executor/RollbackExecutor.php
+++ b/src/Propel/Generator/Command/Executor/RollbackExecutor.php
@@ -80,7 +80,7 @@ class RollbackExecutor
 
         $migration = $this->migrationManager->getMigrationObject($currentVersion);
 
-        $canBeRollback = $this->isFake() || $migration->preDown($this->migrationManager);
+        $canBeRollback = $this->isFake() || $migration->preDown($this->migrationManager) !== false;
         if (!$canBeRollback && !$this->isForce()) {
             $this->output->writeln('<error>preDown() returned false. Aborting migration.</error>');
 
--- a/src/Propel/Generator/Command/Executor/RollbackExecutor.php
+++ b/src/Propel/Generator/Command/Executor/RollbackExecutor.php
@@ -66,33 +66,21 @@ class RollbackExecutor
     }
 
     /**
-     * @param list<int> $previousTimestamps
+     * @param int $currentVersion
+     * @param int|null $previousVersion
      *
      * @return bool
      */
-    public function executeRollbackToPreviousVersion(array &$previousTimestamps): bool
+    public function executeRollbackToPreviousVersion(int $currentVersion, ?int $previousVersion = null): bool
     {
-        $nextMigrationTimestamp = array_pop($previousTimestamps);
-        if (!$nextMigrationTimestamp) {
-            $this->output->writeln('No migration were ever executed on this database - nothing to reverse.');
-
-            return false;
-        }
-
         $this->output->writeln(sprintf(
             'Executing migration %s down',
-            $this->migrationManager->getMigrationClassName($nextMigrationTimestamp),
+            $this->migrationManager->getMigrationClassName($currentVersion),
         ));
 
-        $nbPreviousTimestamps = count($previousTimestamps);
-        $previousTimestamp = 0;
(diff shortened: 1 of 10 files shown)


src/Propel/Generator/Command/Executor/RollbackExecutor.php
```diff
@@ -0,0 +1,190 @@
...
+                }
+
+                $conn->exec($statement);
+                $res++;
+            } catch (Exception $e) {
+                if ($this->input->getOption(static::COMMAND_OPTION_FORCE)) {
+                    //continue, but print error message
+                    $this->output->writeln(
+                        sprintf('<error>Failed to execute SQL "%s". Continue migration.</error>', $statement),
+                    );
```
dereuromark, line 172: Can we please use ` for code fencing in all messages?

src/Propel/Generator/Manager/MigrationManager.php
```diff
@@ -264,13 +265,21 @@ public function updateLatestMigrationTimestamp(string $datasource, int $timestam
...
         $sql = sprintf(
-            'INSERT INTO %s (%s) VALUES (?)',
+            'INSERT INTO %s (%s, %s) VALUES (?, ?)',
             $this->getMigrationTable(),
-            $platform->doQuoting('version'),
+            $platform->doQuoting(static::COL_VERSION),
+            $platform->doQuoting(static::COL_EXECUTION_DATETIME),
         );
+
+        $executionDatetime = date('Y-m-d H:i:s');
```
yaroslav-spryker, line 278: Could you please move the date template to the protected const?

yaroslav-spryker (changes requested):

src/Propel/Generator/Manager/MigrationManager.php
romanhavrylko: @yaroslav-spryker, moved.

mringler (changes requested):
Very nice!

Some smaller issues I think should be changed, mostly typos and some things I found hard to read Let me know what you think!

I'm not through all the way, need to take a break now.

yaroslav-spryker (approved):

src/Propel/Generator/Manager/MigrationManager.php
```diff
@@ -295,14 +304,25 @@ public function getMigrationTimestamps(): array
     }
 
     /**
-     * @return array<int>
+     * @param int|null $version
+     *
+     * @return list<int>
      */
-    public function getValidMigrationTimestamps(): array
+    public function getValidMigrationTimestamps(?int $version = null): array
```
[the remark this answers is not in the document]
romanhavrylko: Thank you for the suggestion! Created a new method.

src/Propel/Generator/Command/Executor/RollbackExecutor.php
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    /**
+     * @param list<int> $previousTimestamps
+     *
+     * @return bool
+     */
+    public function executeRollbackToPreviousVersion(array &$previousTimestamps): bool
+    {
+        $nextMigrationTimestamp = array_pop($previousTimestamps);
```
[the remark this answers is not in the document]
romanhavrylko: Thank you for the suggestion! I think it should have only the necessary arguments should increase readability.

src/Propel/Generator/Manager/MigrationManager.php
```diff
@@ -150,34 +160,27 @@ public function getAllDatabaseVersions(): array
...
+                $migrationTimestamps += $this->getMigrationData($name);
             } catch (PDOException $e) {
                 $this->createMigrationTable($name);
                 $migrationTimestamps = [];
             }
         }
 
-        sort($migrationTimestamps);
+        usort($migrationTimestamps, function (array $a, array $b) {
+            if ($a[static::COL_EXECUTION_DATETIME] === $b[static::COL_EXECUTION_DATETIME]) {
```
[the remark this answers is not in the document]
romanhavrylko: According to the investigation that was made by @valerio8787, Propel migrate down command may roll back migration in the wrong order. E.g. we have two releases of the application, which contain the following migrations:
```
release 1 (migration applied 01.05.2022):
migration_001.php
migration_008.php
```
```
release 2 (migration applied 01.06.2022):
migration_005.php (migration created during development earlier than migration_008.php from release 1 and has an earlier version)
```

After the deployment of all versions, Propel migration table will have the following state:
```
version
--
001
008
005
```

Now if we run `propel migrate:down` command (that rollback only one migration) it will roll back migration with version 008 first, as Propel orders executed migrations by version and then rolls back a migration with max version value.

src/Propel/Generator/Command/MigrationStatusCommand.php
```diff
@@ -106,6 +122,13 @@ protected function execute(InputInterface $input, OutputInterface $output): int
             }
         }
 
+        if ($input->getOption(static::COMMAND_OPTION_LAST_VERSION) && $oldestMigrationTimestamp) {
+            $output->writeln(sprintf(
+                'The last executed version of the migration is %s',
```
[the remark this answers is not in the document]
romanhavrylko: According to the initial idea, if this option is passed, the command should return only the last version.

mringler (approved):
Nice, thank you for the changes, I think it is great now.

src/Propel/Generator/Command/Executor/RollbackExecutor.php
```diff
@@ -0,0 +1,196 @@
...
+    public function executeRollbackToPreviousVersion(int $currentVersion, ?int $previousVersion = null): bool
+    {
+        $this->output->writeln(sprintf(
+            'Executing migration %s down',
+            $this->migrationManager->getMigrationClassName($currentVersion),
+        ));
+
+        $migration = $this->migrationManager->getMigrationObject($currentVersion);
+        if (!$this->isFake() && $migration->preDown($this->migrationManager) === false) {
+            if (!$this->isForce()) {
```
abitskil, line 83: * Could you rewrite the conditions to get rid of this level of nesting?

mringler: @abitskil Can you give an example of what you have in mind?

mringler: At a glance, I am not sure that would work, these are actually nested conditions. You can make it look different, like `if (a && b) {...} else if (a) {...}`, but I am  not sure if that is better. To me, there is no obvious "good" solution here.
Which is why I would like to know what you, @abitskil, had in mind. It seems like three people are now thinking about your comment.

dereuromark:
The PHPStan issue seems unrelated.
:+1:

dereuromark:
@yaroslav-spryker @mringler Can you maybe approve again?

abitskil (approved):

PR #1934, opened 2022-12-27
Update README: add link to test documentation

mringler:
Updated the "Contribute" segment of the README to feature the link to the test documentation more prominently. Knowing how to run tests locally is very important for new contributors, I think it makes sense to link to the page directly.

There is also a link to a roadmap, but the page has not been updated in 2.5 years. Should we remove that link?


dereuromark:
Yeah probably should

PR #1935, opened 2022-12-28
Manage foreign keys on archivable behavior

mringler:
Allows better management of foreign keys in archivable behavior.

Basically implements #1484 with tests, where FKs can be copied from the source table to an automatically generated archive table.
Also, it adds the possibility to declare new FKs as part of the behavior declaration.
Finally, it adds a parameter `sync`, which, if enabled, copies columns, indexes and optionally fks to an archive table that is not automatically generated, allowing for flexible overriding of archive table properties.

I have added documentation for these options along with the different use cases they cover at propelorm/propelorm.github.com#433

### DIFF
--- a/src/Propel/Generator/Behavior/Archivable/ArchivableBehavior.php
+++ b/src/Propel/Generator/Behavior/Archivable/ArchivableBehavior.php
@@ -10,8 +10,10 @@ namespace Propel\Generator\Behavior\Archivable;
 
 use Propel\Generator\Builder\Om\AbstractOMBuilder;
 use Propel\Generator\Exception\InvalidArgumentException;
+use Propel\Generator\Exception\SchemaException;
 use Propel\Generator\Model\Behavior;
 use Propel\Generator\Model\Column;
+use Propel\Generator\Model\ForeignKey;
 use Propel\Generator\Model\Index;
 use Propel\Generator\Model\Table;
 use Propel\Generator\Platform\PgsqlPlatform;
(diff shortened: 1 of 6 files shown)


src/Propel/Generator/Behavior/Archivable/ArchivableBehavior.php
```diff
@@ -90,79 +92,176 @@ public function modifyTable(): void
...
-            // create the version table
-            $archiveTable = $database->addTable([
-                'name' => $archiveTableName,
-                'phpName' => $this->getParameter('archive_phpname'),
-                'package' => $table->getPackage(),
-                'schema' => $table->getSchema(),
-                'namespace' => $table->getNamespace() ? '\\' . $table->getNamespace() : null,
-                'identifierQuoting' => $table->isIdentifierQuotingEnabled(),
+
+        // copy all the columns
```
yaroslav-spryker, line 154: Good job so far!

Can we please move all exact functionality for their own methods to meet SRP?
So in the end we have this method something like
```
$this->syncColumns($sourceTable, $archiveTable);
$this->syncForeignKeys($sourceTable, $archiveTable);
$this->syncIndixes($sourceTable, $archiveTable);
// so on
```

Thank you in advance!

yaroslav-spryker (changes requested):

src/Propel/Generator/Behavior/Archivable/ArchivableBehavior.php
mringler: Thank you for the review!
Method is split up now.

yaroslav-spryker (approved):

mringler:
@dereuromark Thank you for merging. Documentation is at https://github.com/propelorm/propelorm.github.com/pull/433

PR #1943, opened 2023-01-18
Fixed cs migration template

michbeck:


dereuromark (dismissed):

templates/Manager/migration_template.php
```diff
@@ -9,24 +9,47 @@
...
+     */
     public $comment = '<?= $commentString ?>';
 
-    public function preUp(MigrationManager $manager)
+    /**
+     * @param \Propel\Generator\Manager\MigrationManager $manager
+     *
+     * @return void
+     */
+    public function preUp(MigrationManager $manager): void
```
mringler, line 22: `preUp()` does not have to return anything, but if it returns `false`, migration will be aborted. So the PHP return type is probably `mixed`? Maybe `psalm` allows for something like `false|void`.

templates/Manager/migration_template.php
```diff
@@ -9,24 +9,47 @@
...
         // add the post-migration code here
     }
 
-    public function preDown(MigrationManager $manager)
+    /**
+     * @param \Propel\Generator\Manager\MigrationManager $manager
+     *
+     * @return void
+     */
+    public function preDown(MigrationManager $manager): void
```
mringler, line 42: Same as `preUp()`

mringler (changes requested):
Hey, nice, just one problem with the return types of `preUp()` and `preDown()`

templates/Manager/migration_template.php
dereuromark: so it is `null|false|void` in docblock it seems
But kind of a code smell IMO legacy wise
Could we maybe use a different way to abort moving forward? Like using a specific exception that we catch?

dereuromark: @michbeck Can we omit the actual type for now and just document as I suggested?
We can then in a follow up PR fix it up to use an exception based halting instead of false smell.

mringler: Does psalm or stan complain if a method declares to throw an exception, but doesn't?
It would be a problem if this gives a warning:
```php
    /**
     * @param \Propel\Generator\Manager\MigrationManager $manager
     *
     * @throws \...\AbortMigrationException
     * 
     * @return void
     */
    public function preUp(MigrationManager $manager)
    {
        // add the pre-migration code here
    }
```

dereuromark: No, thats fine IMO

mringler (approved):

dereuromark:
Follow up: Fix it using exception and void return.

PR #1945, opened 2023-01-18
Fix CS by removing inline assignment.

dereuromark:
Follow https://github.com/propelorm/Propel2/pull/1944

The remaining issues need some code refactoring.

Issue #1921, opened 2022-12-09
New Beta release


dereuromark:
We should plan a new beta release for the upcoming weeks.
The UUID feature is a great topic for it.

Maybe we can further include also some cleanup?

I think we should also:
- [x] Check open issues for what should be included, especially regression fixes and bugfixes
- [x] Finish open PRs as good as we can to include
- [x] Remove some of those silencing https://github.com/propelorm/Propel2/blob/b542279830944d1634ed1816f1f01287403155e0/phpcs.xml#L31-L45 and general improvement of CS where possible
- [ ] https://github.com/propelorm/Propel2/pull/1938
- [x] PHP 8.2+ support / CI check
- [x] PHP Stan level 7+ (Found 165 errors until level 7) or at least fix fixable easy wins here

dereuromark:
Anyone able to assist? So we can get this done this month?

PR #1946, opened 2023-01-19
Fix CS by removing inline assignments.

michbeck:
Follow of #1945 for #1921

### DIFF
--- a/src/Propel/Generator/Model/ForeignKey.php
+++ b/src/Propel/Generator/Model/ForeignKey.php
@@ -703,7 +703,9 @@ class ForeignKey extends MappingModel
     {
         $mapping = [];
         for ($i = 0, $size = count($this->localColumns); $i < $size; $i++) {
-            if ($right = $this->foreignColumns[$i]) {
+            $right = $this->foreignColumns[$i];
+
+            if ($right) {
                 $right = $this->getForeignTable()->getColumn($right);
             } else {
                 $right = $this->localValues[$i];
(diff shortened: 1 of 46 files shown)


mringler (approved):
Nice!

dereuromark:
The order in some cases cannot change 
Otherwise this creates Bugs. Some of this can be seen in current test fails, others might Fall through the Cracks of CI and code review

src/Propel/Generator/Model/ForeignKey.php
```diff
@@ -466,7 +466,8 @@ public function getForeignTableName(): ?string
         }
 
         $database = $this->getDatabase();
-        if ($database && ($schema = $this->parentTable->guessSchemaName()) && $platform && $platform->supportsSchemas()) {
+        $schema = $this->parentTable->guessSchemaName();
```
mringler, line 469: Umph, test fail because apparently `$database && ($schema = $this->parentTable->guessSchemaName())` tests if parent table has a database, so that `Table::guessSchemaName()` does not fail with an exception:
```php
    public function guessSchemaName(): ?string
    {
        return $this->schema ?: $this->database->getSchema();
    }
```
Pretty horrible. I think it should be
```php
    public function guessSchemaName(): ?string
    {
        if($this->schema){
            return $this->schema;
        }
        return $this->database ? $this->database->getSchema() : null;
    }
```

mringler (changes requested):

mringler (approved):

PR #1947, opened 2023-01-20
Fix up datetime PHPStan error.

dereuromark:
and fix some PHPStan issues of level 7

### DIFF
--- a/src/Propel/Generator/Behavior/Sortable/SortableBehaviorObjectBuilderModifier.php
+++ b/src/Propel/Generator/Behavior/Sortable/SortableBehaviorObjectBuilderModifier.php
@@ -289,6 +289,7 @@ protected \$oldScope;
                     $script = str_replace($search, $replace, $script);
                 }
             } else {
+                /** @var string $scope */
                 $scope = current($this->behavior->getScopes());
                 $name = strtolower($this->behavior->getTable()->getColumn($scope)->getName());
 
--- a/src/Propel/Runtime/DataFetcher/AbstractDataFetcher.php
+++ b/src/Propel/Runtime/DataFetcher/AbstractDataFetcher.php
@@ -49,6 +49,7 @@ abstract class AbstractDataFetcher implements DataFetcherInterface
      */
     public function fetchColumn($index = null)
     {
+        /** @var array|null $next */
         $next = $this->fetch();
 
         if (!$next) {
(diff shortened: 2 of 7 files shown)


src/Propel/Generator/Behavior/Sortable/SortableBehaviorObjectBuilderModifier.php
```diff
@@ -289,6 +289,7 @@ public function objectFilter(string &$script, AbstractOMBuilder $builder): void
                     $script = str_replace($search, $replace, $script);
                 }
             } else {
+                /** @var string $scope */
                 $scope = current($this->behavior->getScopes());
```
mringler, line 293: Just out of curiosity, why `/** @var string $scope */` and not `$scope = (string) current(...);`?

src/Propel/Runtime/DataFetcher/AbstractDataFetcher.php
```diff
@@ -49,6 +49,7 @@ public function getDataObject()
      */
     public function fetchColumn($index = null)
     {
+        /** @var array|null $next */
```
mringler, line 52: I think `fetch()` returns false if there are no more columns to fetch

mringler (approved):
Nice!

src/Propel/Generator/Behavior/Sortable/SortableBehaviorObjectBuilderModifier.php
dereuromark: In this context it looked like an impossible way to get another value here, so I just annotated it for humans and tools to know what to expect.
Sometimes casts are also not "valid" from Stan perspective since they might not work on certain return types.

src/Propel/Runtime/DataFetcher/AbstractDataFetcher.php
dereuromark: Yeah, we could check on this. I just thought in this context it seems impossible (theoretical problem).

mringler: I think the check happens in the next line:
```php
        if (!$next) {
            return null;
        }
```

dereuromark: array|bool|null

The bool true is the problem here
And instead of just checking on this, I simplified to the expected outcome as docblock.

michbeck (approved):

PR #1948, opened 2023-01-20
Fix up PHPStan errors.

dereuromark:
Fix some more PHPStan issues of level 7

~120 issues left

### DIFF
--- a/src/Propel/Generator/Reverse/PgsqlSchemaParser.php
+++ b/src/Propel/Generator/Reverse/PgsqlSchemaParser.php
@@ -243,7 +243,7 @@ class PgsqlSchemaParser extends AbstractSchemaParser
         } elseif (!$table->getDatabase()->getSchema()) {
             $stmt = $this->dbh->query('SHOW search_path');
             if ($stmt === false) {
-                throw new RuntimeException('Query returned no statement.');
+                throw new RuntimeException('Could not retrieve search_path from database.');
             }
             $searchPathString = $stmt->fetchColumn();
 
--- a/src/Propel/Generator/Reverse/PgsqlSchemaParser.php
+++ b/src/Propel/Generator/Reverse/PgsqlSchemaParser.php
@@ -17,6 +17,7 @@ use Propel\Generator\Model\Index;
 use Propel\Generator\Model\PropelTypes;
 use Propel\Generator\Model\Table;
 use Propel\Generator\Model\Unique;
+use RuntimeException;
 use stdClass;
 
 /**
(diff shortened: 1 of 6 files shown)


src/Propel/Generator/Reverse/PgsqlSchemaParser.php
```diff
@@ -633,6 +668,9 @@ protected function addSequences(Database $database): void
             AND c.relkind = 'S'
             AND n.nspname IN ($searchPath);
         ");
+        if ($stmt === false) {
+            throw new RuntimeException('prepare() returned no statement.');
```
mringler, line 672: Uff, thank you for adding all those! Might make sense to change `prepare()` to throw an exception instead of doing the stupid `false` return

src/Propel/Generator/Reverse/SqliteSchemaParser.php
```diff
@@ -150,6 +150,7 @@ protected function parseTables(Database $database, ?Table $filterTable = null):
 
         $sql = str_replace('%filter%', $filter, $sql);
 
+        /** @var array $dataFetcher */
         $dataFetcher = $this->dbh->query($sql);
```
mringler, line 154: Return type is actually `\Propel\Runtime\DataFetcher\DataFetcherInterface|\PDOStatement|false`. I think we can we do:
```php
        /** @var \Traversable $dataFetcher */
        $dataFetcher = $this->dbh->query($sql);
```

mringler (approved):
Nice also

src/Propel/Generator/Reverse/PgsqlSchemaParser.php
```diff
@@ -239,6 +242,9 @@ protected function addColumns(Table $table, int $oid): void
             $params = [$schema];
         } elseif (!$table->getDatabase()->getSchema()) {
             $stmt = $this->dbh->query('SHOW search_path');
+            if ($stmt === false) {
+                throw new RuntimeException('Query returned no statement.');
```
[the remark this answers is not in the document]
dereuromark: If you make comments with suggested code changes, I can directly commit them in :)

mringler, line 246:
```suggestion
                throw new RuntimeException('Could not retrieve search_path from database.');
```

src/Propel/Generator/Reverse/SqliteSchemaParser.php
```diff
@@ -150,6 +150,7 @@ protected function parseTables(Database $database, ?Table $filterTable = null):
 
         $sql = str_replace('%filter%', $filter, $sql);
 
+        /** @var array $dataFetcher */
```
mringler, line 153:
```suggestion
        /** @var \Traversable $dataFetcher */
```

mringler (approved):
Ah, so that is how those suggestions work.
I never accept them, because they lead to merge commits in my feature branch, but if you like them, I'll try to think of them in the future.

dereuromark:
If we squash merge, they will not :)
And we should do that with such dirty commits anyway then by default.

michbeck (approved):

PR #1950, opened 2023-01-24
Fix up IDE found errors/problems.

dereuromark:
PHPStorm found those

### DIFF
--- a/src/Propel/Generator/Model/Domain.php
+++ b/src/Propel/Generator/Model/Domain.php
@@ -168,11 +168,11 @@ class Domain extends MappingModel
     /**
      * Sets the domain description.
      *
-     * @param string $description
+     * @param string|null $description
      *
      * @return void
      */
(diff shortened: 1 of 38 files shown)


src/Propel/Generator/Model/Domain.php
```diff
@@ -190,11 +190,11 @@ public function getName(): ?string
     /**
      * Sets the domain name.
      *
-     * @param string $name
+     * @param string|null $name
      *
      * @return void
      */
-    public function setName(string $name): void
+    public function setName(?string $name): void
```
dereuromark, line 197: I dont know if we need this, but the getter is nullable as well

> getName(): ?string

mringler: I think it makes sense. If you can set it, you probably should be able to unset it

mringler (approved):

michbeck (approved):

PR #1951, opened 2023-01-24
Fixes phpstan errors, set PHPStan to level 7

michbeck:
PHP Stan fixes for #1921

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,12 +24,12 @@
         "symfony/filesystem": "^4.4.0 || ^5.0.0 || ^6.0.0",
         "symfony/finder": "^4.4.0 || ^5.0.0 || ^6.0.0",
         "symfony/translation": "^4.4.0 || ^5.0.0 || ^6.0.0",
-        "symfony/validator": "^4.4.0 || ^5.0.0 || ^6.0.0",
-        "ext-xml": "*"
+        "symfony/validator": "^4.4.0 || ^5.0.0 || ^6.0.0"
     },
     "require-dev": {
         "ext-pdo": "*",
         "ext-json": "*",
+        "ext-xml": "*",
         "monolog/monolog": "^1.3 || ^2.3 || ^3.0",
         "phpstan/phpstan": "^1.2",
         "phpunit/phpunit": "^9.5.0",
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,8 @@
         "symfony/filesystem": "^4.4.0 || ^5.0.0 || ^6.0.0",
         "symfony/finder": "^4.4.0 || ^5.0.0 || ^6.0.0",
         "symfony/translation": "^4.4.0 || ^5.0.0 || ^6.0.0",
-        "symfony/validator": "^4.4.0 || ^5.0.0 || ^6.0.0"
+        "symfony/validator": "^4.4.0 || ^5.0.0 || ^6.0.0",
+        "ext-xml": "*"
     },
     "require-dev": {
         "ext-pdo": "*",
--- a/composer.json
+++ b/composer.json
@@ -24,12 +24,12 @@
         "symfony/filesystem": "^4.4.0 || ^5.0.0 || ^6.0.0",
         "symfony/finder": "^4.4.0 || ^5.0.0 || ^6.0.0",
         "symfony/translation": "^4.4.0 || ^5.0.0 || ^6.0.0",
-        "symfony/validator": "^4.4.0 || ^5.0.0 || ^6.0.0"
+        "symfony/validator": "^4.4.0 || ^5.0.0 || ^6.0.0",
+        "ext-xml": "*"
     },
     "require-dev": {
         "ext-pdo": "*",
         "ext-json": "*",
-        "ext-xml": "*",
         "monolog/monolog": "^1.3 || ^2.3 || ^3.0",
         "phpstan/phpstan": "^1.2",
         "phpunit/phpunit": "^9.5.0",
--- a/src/Propel/Generator/Behavior/Archivable/ArchivableBehavior.php
+++ b/src/Propel/Generator/Behavior/Archivable/ArchivableBehavior.php
@@ -309,7 +309,7 @@ class ArchivableBehavior extends Behavior
     }
 
     /**
-     * @psalm-param array{local_column: string, foreign_table: string, foreign_column: string, relation_only?: string} $fkParameterData
+     * @psalm-param array{name?: string, localColumn: string, foreignTable: string, foreignColumn: string, relationOnly?: string} $fkParameterData
      *
      * @param \Propel\Generator\Model\Table $table
      * @param array $fkParameterData
@@ -320,14 +320,12 @@ class ArchivableBehavior extends Behavior
      */
     protected function createForeignKeyFromParameters(Table $table, array $fkParameterData): void
     {
-        if (
-            empty($fkParameterData['localColumn']) ||
-            empty($fkParameterData['foreignColumn'])
-        ) {
+        if (empty($fkParameterData['localColumn']) || empty($fkParameterData['foreignColumn'])) {
             $tableName = $this->table->getName();
 
             throw new SchemaException("Table `$tableName`: Archivable behavior misses foreign key parameters. Please supply `localColumn`, `foreignTable` and `foreignColumn` for every entry");
         }
+
         $fk = new ForeignKey($fkParameterData['name'] ?? null);
         $fk->addReference($fkParameterData['localColumn'], $fkParameterData['foreignColumn']);
         $table->addForeignKey($fk);
--- a/src/Propel/Generator/Behavior/Archivable/ArchivableBehavior.php
+++ b/src/Propel/Generator/Behavior/Archivable/ArchivableBehavior.php
@@ -320,7 +320,10 @@ class ArchivableBehavior extends Behavior
      */
     protected function createForeignKeyFromParameters(Table $table, array $fkParameterData): void
     {
-        if (empty($fkParameterData['localColumn']) || empty($fkParameterData['foreignColumn'])) {
+        if (
+            empty($fkParameterData['localColumn']) ||
+            empty($fkParameterData['foreignColumn'])
+        ) {
             $tableName = $this->table->getName();
 
             throw new SchemaException("Table `$tableName`: Archivable behavior misses foreign key parameters. Please supply `localColumn`, `foreignTable` and `foreignColumn` for every entry");
--- a/src/Propel/Generator/Builder/Om/AbstractOMBuilder.php
+++ b/src/Propel/Generator/Builder/Om/AbstractOMBuilder.php
@@ -910,10 +910,9 @@ abstract class AbstractOMBuilder extends DataModelBuilder
 
         foreach ($fk->getMapping() as $mapping) {
             [$localColumn, $foreignValueOrColumn] = $mapping;
-            $localColumnName = $localColumn->getPhpName();
             $localTable = $fk->getTable();
-            if (!$localColumnName) {
-                throw new RuntimeException(sprintf('Could not fetch column of table %s.', $localTable->getName()));
+            if (!$localColumn) {
+                throw new RuntimeException(sprintf('Could not resolve column of foreign key `%s` on table `%s`', $fk->getName(), $localTable->getName()));
             }
 
             $tableName = $fk->getTableName();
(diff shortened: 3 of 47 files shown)


composer.json
```diff
@@ -24,7 +24,8 @@
         "symfony/filesystem": "^4.4.0 || ^5.0.0 || ^6.0.0",
         "symfony/finder": "^4.4.0 || ^5.0.0 || ^6.0.0",
         "symfony/translation": "^4.4.0 || ^5.0.0 || ^6.0.0",
-        "symfony/validator": "^4.4.0 || ^5.0.0 || ^6.0.0"
+        "symfony/validator": "^4.4.0 || ^5.0.0 || ^6.0.0",
+        "ext-xml": "*"
```
dereuromark, line 28: We have them in require-dev for now it seems.

michbeck: Does it make sense to add it to `require-de`v instead of `require`?

dereuromark: If it is optional, so if there are ways to use the library without the need of it.

dereuromark: Optional means that even in require mode there are some cases where it would not be needed.

dereuromark: Why did you move it to require again?

src/Propel/Generator/Behavior/Archivable/ArchivableBehavior.php
```diff
@@ -320,14 +320,12 @@ protected function reapplyBehaviors(array $behaviors)
...
-        if (
-            empty($fkParameterData['localColumn']) ||
-            empty($fkParameterData['foreignColumn'])
-        ) {
+        if (empty($fkParameterData['localColumn']) || empty($fkParameterData['foreignColumn'])) {
             $tableName = $this->table->getName();
 
             throw new SchemaException("Table `$tableName`: Archivable behavior misses foreign key parameters. Please supply `localColumn`, `foreignTable` and `foreignColumn` for every entry");
         }
+
```
dereuromark, line 328: Can we remove those noisy changes?

src/Propel/Generator/Platform/SqlitePlatform.php
```diff
@@ -640,7 +643,12 @@ protected function getVersion(): string
 
         //if php_sqlite3 extension is not installed, we need to query the database
         $connection = new PdoConnection('sqlite::memory:');
+        $pdoStatement = $connection->query('SELECT sqlite_version()');
+
+        if ($pdoStatement === false) {
+            throw new RuntimeException('query() did not return a result set as a statement object.');
```
gechetspr, line 649: `PdoConnection::query()` looks more understandable IMO

src/Propel/Generator/Reverse/MssqlSchemaParser.php
```diff
@@ -86,12 +87,18 @@ protected function getTypeMapping(): array
...
+     * @throws \RuntimeException
+     *
      * @return int
      */
     public function parse(Database $database, array $additionalTables = []): int
     {
         $dataFetcher = $this->dbh->query("SELECT TABLE_NAME FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_TYPE = 'BASE TABLE' AND TABLE_NAME <> 'dtproperties'");
 
+        if ($dataFetcher === false) {
+            throw new RuntimeException('query() did not return a result set as a statement object.');
```
gechetspr, line 99: same

src/Propel/Generator/Reverse/OracleSchemaParser.php
```diff
@@ -204,16 +205,11 @@ protected function addColumns(Table $table): void
...
     protected function addIndexes(Table $table): void
     {
-        /** @var \PDOStatement $stmt */
-        $stmt = $this->dbh->query("SELECT INDEX_NAME, COLUMN_NAME FROM USER_IND_COLUMNS WHERE TABLE_NAME = '" . $table->getName() . "' ORDER BY COLUMN_NAME");
-
-        $indices = [];
-        while ($row = $stmt->fetch(PDO::FETCH_ASSOC)) {
-            $indices[$row['INDEX_NAME']][] = $row['COLUMN_NAME'];
-        }
+        $indicesIndexedByIndexName = $this->getIndiciesIndexedByIndexName($table);
```
gechetspr, line 208: But in the end those are column names indexed be index name

src/Propel/Generator/Util/QuickBuilder.php
```diff
@@ -342,6 +344,11 @@ public function updateDB(ConnectionInterface $con): ?Database
         foreach ($statements as $statement) {
             try {
                 $stmt = $con->prepare($statement);
+
+                if ($stmt === false) {
+                    throw new RuntimeException('prepare() failed and did not return statement object for execution.');
```
gechetspr, line 349: I would use class name with methods for visibility

gechetspr (changes requested):

src/Propel/Generator/Behavior/Archivable/ArchivableBehavior.php
dereuromark: Aktuelly, this looks fine.

src/Propel/Generator/Reverse/OracleSchemaParser.php
```diff
@@ -228,6 +224,24 @@ protected function addIndexes(Table $table): void
         }
     }
 
+    /**
+     * @param \Propel\Generator\Model\Table $table
+     *
+     * @return array<scalar, non-empty-list<null|scalar>>
+     */
+    protected function getIndiciesIndexedByIndexName(Table $table): array
```
dereuromark, line 232: Based on comment above, should this be
```suggestion
    protected function getColumnNamesIndexedByIndexName(Table $table): array
```
?

dereuromark:
@mringler This would then be the last PR merged for the release of new beta this week
Is this OK for you?

src/Propel/Generator/Reverse/OracleSchemaParser.php
michbeck: Already done

src/Propel/Generator/Manager/MigrationManager.php
```diff
@@ -439,9 +464,11 @@ public function findMigrationClassNameSuffix(int $timestamp): string
         $path = $this->getWorkingDirectory();
         if ($path && is_dir($path)) {
             $files = scandir($path);
```
dereuromark, line 466:
```suggestion
            $files = scandir($path) ?: [];
```
saves you the if case

src/Propel/Generator/Model/ColumnDefaultValue.php
```diff
@@ -94,7 +94,7 @@ public function getValue()
      */
     public function setValue($value): void
     {
-        $this->value = $value;
+        $this->value = (string)$value;
```
dereuromark, line 97: what about null?
also getValue() cannot be string|int|null anymore then?

src/Propel/Generator/Model/ColumnDefaultValue.php
```diff
@@ -118,10 +118,9 @@ public function equals(ColumnDefaultValue $other): bool
...
         // special case for current timestamp
         $equivalents = ['CURRENT_TIMESTAMP', 'NOW()'];
-        if (in_array(strtoupper($this->getValue()), $equivalents) && in_array(strtoupper($other->getValue()), $equivalents)) {
-            return true;
-        }
+        $value = strtoupper((string)$this->getValue());
+        $otherValue = strtoupper((string)$other->getValue());
 
-        return false; // Can't help, they are different
+        return in_array($value, $equivalents) && in_array($otherValue, $equivalents);
```
dereuromark, line 124: true should be always set for stringish comparison (3rd param)

dereuromark:
Nice work!

src/Propel/Generator/Model/NameFactory.php
```diff
@@ -42,18 +42,19 @@ class NameFactory
     /**
      * Factory method which retrieves an instance of the named generator.
      *
-     * @param string $name The fully qualified class name of the name
-     *                     generation algorithm to retrieve.
+     * @param string $nameGeneratorClassName The fully qualified class name of the name generation algorithm to retrieve.
```
mringler, line 45: isn't this of type `class-string`

src/Propel/Generator/Reverse/OracleSchemaParser.php
```diff
@@ -228,6 +224,24 @@ protected function addIndexes(Table $table): void
...
+     *
+     * @return array<scalar, non-empty-list<null|scalar>>
+     */
+    protected function getColumnNamesIndexedByIndexName(Table $table): array
+    {
+        /** @var \PDOStatement $stmt */
+        $stmt = $this->dbh->query("SELECT INDEX_NAME, COLUMN_NAME FROM USER_IND_COLUMNS WHERE TABLE_NAME = '" . $table->getName() . "' ORDER BY COLUMN_NAME");
+
+        $columnNamesIndexedByIndexName = [];
+        while ($row = $stmt->fetch(PDO::FETCH_ASSOC)) {
```
mringler, line 238: why don't we need the `if ($stmt === false)` check here?

src/Propel/Runtime/ActiveQuery/Criteria.php
```diff
@@ -1063,15 +1063,17 @@ public function addJoin($left, $right, ?string $joinType = null)
         $join->setIdentifierQuoting($this->isIdentifierQuotingEnabled());
 
         // is the left table an alias ?
-        $dotpos = strrpos($left, '.');
-        $leftTableAlias = substr($left, 0, $dotpos);
-        $leftColumnName = substr($left, $dotpos + 1);
+        /** @phpstan-var string $left */
+        $length = strrpos($left, '.') ?: null;
```
mringler, line 1067: Why do you prefer `$length` over `$dotpos`? The latter seems more declarative?

mringler (commented):
Very nice! Seems like you are literally kicking the crap out of Propel.

src/Propel/Generator/Builder/Om/AbstractOMBuilder.php
```diff
@@ -913,7 +913,7 @@ protected static function getRelatedBySuffix(ForeignKey $fk): string
             $localColumnName = $localColumn->getPhpName();
             $localTable = $fk->getTable();
             if (!$localColumnName) {
-                throw new RuntimeException(sprintf('Could not fetch column: %s in table %s.', $localColumnName, $localTable->getName()));
+                throw new RuntimeException(sprintf('Could not fetch column of table %s.', $localTable->getName()));
             }
```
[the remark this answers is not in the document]
dereuromark: Then this was on me, since I saw sth weird/impossible going on before that:
https://github.com/propelorm/Propel2/commit/5531e13fa46f9a0ae31e0594b0a859c97ec747eb#diff-9f9a6879f541dc6ac7340c74a31dd8464e45c8d254a89930718f1080c52e65caL915

src/Propel/Generator/Reverse/OracleSchemaParser.php
dereuromark: I guess the while truish implicist cast takes care of this?

mringler:
> @mringler This would then be the last PR merged for the release of new beta this week Is this OK for you?

Awesome!

dereuromark:
Also, only 981 errors more to go for level 8 then :P

src/Propel/Runtime/ActiveQuery/Criteria.php
dereuromark: $dotPos sounds fair

michbeck: The result is used as `$length` for substr method to retrieve the leftTableAlias, as well to get `$offset` to retrieve the leftColumnName which is which is depending on the `$length` used before.

src/Propel/Generator/Builder/Om/AbstractOMBuilder.php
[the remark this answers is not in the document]
mringler: but you were right

src/Propel/Runtime/ActiveQuery/Criteria.php
mringler: Yes, that I can see. But length of what though? I get that `$dotpos` is a weird abbrev, even coming from a function named `strpos`. But at least I can infer that it is the position of a dot. The name `$length` tells me nothing, except that it is some point in an unknown dimension. Shouldn't a rename rather be more precise than less, i.e. `$firstDotPosition` or `$seperatorIndex` or such?

dereuromark: I would also say, that the length name of the target method is also coming to a point later in code.
So here we don't know that yet, so the name usually (also in other pieces of the code) is named after what is done here, the position finding.

michbeck:
> Also, only 981 errors more to go for level 8 then :P

Is this a note to be a volunteer? :P

PR #1963, opened 2023-04-23
avoid name collision in cross ref relations

### LINKED ISSUE #941: Additional foreign keys on cross-reference table produce duplicate query methods
TestCase: http://sandbox.propelorm.org/1fb3e57

Attaching additional foreign keys to a cross-reference table produces duplicate methods throwing: `Fatal error: Cannot redeclare method filterByXXXXX`

mringler:
Fixes #941

When you have three FKs on a cross-reference table, Propel will generate methods on the referenced tables, where the name uses a combination of the other two FKs. For example, if a  cross-reference table has three FKs to the tables `A`, `B` and `C`, Propel generates a method `clearABs()` on table `C`. The methods are `clear*`, `init*`, `get*`, `set*`, `count*`, `doAdd*` and `remove*`.

This produces collisions when the name of the cross-reference table is a combination of the names of two of the referenced tables. In above example, that would happen when the cross-reference table is named `AB`. (Somewhat unfortunately, this is the name suggested in the [documentation](http://propelorm.org/documentation/04-relationships.html#many-to-many-relationships).
In this case, Propel adds some methods with the same name, once where it handles the cross-reference table (`clearABs` to clear associated elements in the cross ref table), and once again where it handles the tables connected by the cross-reference table (`clearABs` to clear associated As and Bs). This leads to an exception.

The name for the cross-ref relations is generated in `AbstractOmBuilder::getCrossFKsPhpNameAffix()`, this change tests whether the current table has a FK relation to another table with the same name, and if so, prefixes the relation name with `Cross`. The above method would then be called `clearCrossABs()`.

From what I can see, this only affects names that could not occur before because of the error, so I don't think this affects BC.

### DIFF
--- a/tests/Propel/Tests/Issues/Issue941Test.php
+++ b/tests/Propel/Tests/Issues/Issue941Test.php
@@ -13,14 +13,15 @@ use Propel\Generator\Util\QuickBuilder;
 use Propel\Runtime\Collection\ObjectCollection;
 use Propel\Tests\TestCase;
 use Recherche;
+use Category;
 use RechercheNature;
 
 /**
- * This test proves the bug described in https://github.com/propelorm/Propel2/issues/989.
+ * This test proves the bug described in https://github.com/propelorm/Propel2/issues/941.
  *
  * @group database
  */
-class Issue989Test extends TestCase
+class Issue941Test extends TestCase
 {
     /**
      * @return void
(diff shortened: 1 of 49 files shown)


tests/Propel/Tests/Issues/Issue941Test.php
```diff
@@ -0,0 +1,104 @@
...
+            </database>
+            ';
+            QuickBuilder::buildSchema($schema);
+        }
+    }
+
+    /**
+     * @return void
+     */
+    public function testIssue989()
```
floriankraemer, line 74: In your PR description, you write
> Fixes https://github.com/propelorm/Propel2/issues/941

but the method here is named `testIssue989()`?

olhalivitchuk (approved):

gechetspr (approved):

Issue #1203, opened 2016-05-31
Support ON UPDATE CURRENT_TIMESTAMP in MySQL


ExplodingCabbage:
The internet suggests two alternatives, but neither is ideal:
- https://groups.google.com/forum/#!msg/propel-users/_cMRwwDAM6I/oM9iALZkEQAJ suggests the timestampable behaviour, but this only works if all your interactions with MySQL go through the ORM layer.
- http://stackoverflow.com/questions/30538858/propel-on-update-inside-schema-xml suggests `defaultExpr="CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP"`, but my colleague who tested this informs me that using that results in Propel constantly wanting to modify the column again when running `propel diff`, presumably because it reads the current default value from the DB as just `CURRENT_TIMESTAMP`.

It would be nice to have some way to do this cleanly.

dereuromark:
Is someone able to make a PR here with suggested changes?

PR #1967, opened 2023-05-23
Support ON UPDATE CURRENT_TIMESTAMP in MySQL

Mau04:
This PR adds support for `defaultExpr="CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP"` for MySQL as requested in #1203. Previously, a migration would always contain an ALTER Statement to set `ON UPDATE CURRENT_TIMESTAMP` although it is already like that in the database. The root cause was that the `Extra` part was not checked in the `MysqlSchemaParser` class.

### DIFF
--- a/src/Propel/Generator/Reverse/MysqlSchemaParser.php
+++ b/src/Propel/Generator/Reverse/MysqlSchemaParser.php
@@ -264,6 +264,8 @@ class MysqlSchemaParser extends AbstractSchemaParser
         // BLOBs can't have any default values in MySQL
         $default = preg_match('~blob|text~', $nativeType) ? null : $row['Default'];
 
+        $extra = $row['Extra'];
+
         $propelType = $this->getMappedPropelType($nativeType);
         if (!$propelType) {
             $propelType = Column::DEFAULT_TYPE;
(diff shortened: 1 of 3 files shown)


src/Propel/Generator/Reverse/MysqlSchemaParser.php
```diff
@@ -264,6 +264,8 @@ public function getColumnFromRow(array $row, Table $table): Column
         // BLOBs can't have any default values in MySQL
         $default = preg_match('~blob|text~', $nativeType) ? null : $row['Default'];
 
+        $extra = $row['Extra'];
```
mringler, line 267: Can you explain why this is necessary here? When I look at the code, it does not look like `$row` is changed between here and line 300, where you use it.

mringler (approved):
Nice! Small question, the rest looks good!

src/Propel/Generator/Reverse/MysqlSchemaParser.php
mringler: Makes sense. Personally, I'd prefer it in line 300, as there is no benefit to moving it further up at the moment.

Mau04: Sounds good to me. I changed it accordingly.

mringler:
@Mau04  It looks like the old maintainer of Propel2, @dereuromark, has left the project, and it looks like there is no successor at Spryker at the moment. I have contacted the owner, and he told me that the project will keep being maintained by Spryker, but I haven't heard back if/when there will be someone who actively maintains Propel2, i.e. merges this PRs.
For what it's worth, I am very sorry for the situation! This should have been merged month ago, with all the thanks and commendations that are in order. I found getting acquainted with Propel and creating a PR takes effort, and I wish your first contribution to the project had been a more rewarding experience.
I guess when/if someone takes over, contributors like you and I can decide if they want to come back.
Best wishes

a-sabaa:
Hello @mringler & @Mau04,
We are currently at Spryker creating a development strategy for Propel maintenance under the general product roadmap of Spryker..we have been lacking in support in the recent months of this repository and we're working diligently to fix it processually. We'll come back to this PR during the next few days and we'll handle future contribution with a better response time and higher level of contribution.

Thank you for your time and effort on the repository,
We're looking forward to further contribution!

olhalivitchuk (approved):

PR #1979, opened 2023-09-27
Improve PHP 8.2 compatibility : Partially-supported callable are deprecated

romaincousseau:
The propel gen command when a table in schema.xml depends on a baseClass generates code
in the "Base" namespace that is not compatible with PHP 8.2. 

For example, when saving the object, a deprecation warning is displayed: 
`Deprecated: Use of "self" in callables is deprecated in ... on line ...`

This is a known issue whith PHP8.2 :
[https://php.watch/versions/8.2/partially-supported-callable-deprecation](url)


smhg:
@PhilinTv this can be closed as it was addressed in #2015.

PR #1985, opened 2023-11-23
PHP8.2 - Using ${var} in strings is deprecated, use {$var} instead

scp:


PhilinTv:
Good catch, thanks @scp!

PR #1986, opened 2023-11-29
Bugfix/issue1489 mysql types

### LINKED ISSUE #1489: Diff for LONGVARCHAR false positive (MySQL)
Running propel:diff with a LONGVARCHAR column type finds differences, even they don't exist.

Cause:
During reverse lookup of the schema:
SQL::MEDIUMTEXT is parsed as Propel::LONGVARCHAR 
see MySQLSchemaParser::$mysqlTypeMap

Schema based on mapping config:
Propel::LONGVARCHAR is mapped to SQL::TEXT
see MySQLPlatform::initialize()

MysqlPlatform::getColumnDDL() is creating:
from: "`foobar` TEXT"
to: "`foobar` MEDIUMTEXT"
because it uses the raw sqlType stored in the column object.

mringler:
Fixes #1489

See https://github.com/mringler/Propel2/pull/3

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,13 +16,15 @@ jobs:
             matrix:
                 php-version: [ '7.4', '8.2' ]
                 db-type: [ sqlite, mysql, pgsql, agnostic ]
-                symfony-version: [ '4-min', '4-max', '5-min', '5-max', '6-min', '6-max' ]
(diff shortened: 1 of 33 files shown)


mringler:
@PhilinTv I rebased to current master, no more merge conflicts. Are there other improvements you'd like to see here?

PR #1987, opened 2023-12-11
Make compatible with 8.2 ArrayIterator interface updates

oojacoboo:
`ArrayIterator` interface has gone with explicit `true` return values for sort methods.  However, PHP less than 8.2 will crap out on this.  We, instead, add the `#[\ReturnTypeWillChange]` attribute until we can make 8.2 the minimum supported version.

This replaces/closes #1984


oojacoboo:
@mringler yea, was looking at old logs and thinking it was still failing - fixed.

PhilinTv (approved):

PR #1989, opened 2024-01-21
Fix Collection return type in Query class template

### LINKED ISSUE #1975: PHPStorm code inspection errors
Hello,

During its inspection, PHPStorm returns errors like these in generated base query classes, complaining that the method `Collection` is already defined:

image

The code still works fine but obviously it's a bit inconvenient. I dealt with it by removing the following piece of code from the `baseQueryClassHeader.php` file, deleting all query classes and then re-running `propel model:build`:

```
<?php foreach($columns as $column):?>
 * @method     <?= $modelClass ?>[]|Collection findBy<?= $column->getPhpName() ?>(<?= $column->getPhpType() ?>|array<<?= $column->getPhpType() ?>> $<?= $column->getName() ?>) Return <?= $modelClass ?> objects filtered by the <?= $column->getName() ?> column
 * @psalm-method Collection&\Traversable<<?= $modelClass ?>> findBy<?= $column->getPhpName() ?>(<?= $column->getPhpType() ?>|array<<?= $column->getPhpType() ?>> $<?= $column->getName() ?>) Return <?= $modelClass ?> objects filtered by the <?= $column->getName() ?> column
<?php endforeach;?>
```

Another issue is the word `match` in the same file - it's now a reserved word in PHP so some comments were highlighting it as an error. I fixed it by putting `match` between single quotes on line 67.

What are your thoughts?

tsybenko:
Fixes https://github.com/propelorm/Propel2/issues/1975

### DIFF
--- a/templates/Builder/Om/baseQueryClassHeader.php
+++ b/templates/Builder/Om/baseQueryClassHeader.php
@@ -77,12 +77,12 @@
  * @method     <?= $modelClass ?> requireOneBy<?= $column->getPhpName() ?>(<?= $column->getPhpType() ?> $<?= $column->getName() ?>) Return the first <?= $modelClass ?> filtered by the <?= $column->getName() ?> column and throws <?= $entityNotFoundExceptionClass ?> when not found
 <?php endforeach;?>
  *
- * @method     <?= $modelClass ?>[]|Collection find(?ConnectionInterface $con = null) Return <?= $modelClass ?> objects based on current ModelCriteria
- * @psalm-method Collection&\Traversable<<?= $modelClass ?>> find(?ConnectionInterface $con = null) Return <?= $modelClass ?> objects based on current ModelCriteria
+ * @method     <?= $modelClass ?>[]|\Propel\Runtime\Collection\Collection find(?ConnectionInterface $con = null) Return <?= $modelClass ?> objects based on current ModelCriteria
+ * @psalm-method \Propel\Runtime\Collection\Collection&\Traversable<<?= $modelClass ?>> find(?ConnectionInterface $con = null) Return <?= $modelClass ?> objects based on current ModelCriteria
  *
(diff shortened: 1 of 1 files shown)


PhilinTv:
Thanks @tsybenko for the PR!

Unfortunately the issue is not in the FQCN, the a wrong recognition of psalm class intersection (&), which triggers Phpstorm to recognise Collection as a magic function name 🙈 

With FQCN, IDEs don't recognize this row complete, so we can use that to temporary fix it for psalm notation without losing information.

templates/Builder/Om/baseQueryClassHeader.php
```diff
@@ -77,12 +77,12 @@
  * @method     <?= $modelClass ?> requireOneBy<?= $column->getPhpName() ?>(<?= $column->getPhpType() ?> $<?= $column->getName() ?>) Return the first <?= $modelClass ?> filtered by the <?= $column->getName() ?> column and throws <?= $entityNotFoundExceptionClass ?> when not found
 <?php endforeach;?>
  *
- * @method     <?= $modelClass ?>[]|Collection find(?ConnectionInterface $con = null) Return <?= $modelClass ?> objects based on current ModelCriteria
- * @psalm-method Collection&\Traversable<<?= $modelClass ?>> find(?ConnectionInterface $con = null) Return <?= $modelClass ?> objects based on current ModelCriteria
+ * @method     <?= $modelClass ?>[]|\Propel\Runtime\Collection\Collection find(?ConnectionInterface $con = null) Return <?= $modelClass ?> objects based on current ModelCriteria
```
PhilinTv, line 80: Standard Phpdoc is fine with this notation, as there is already a declaraction above in use.
Please update only psalm.

PhilinTv (changes requested):
Standard phpdoc should remain, otherwise it triggers overqualification issue in IDE.

Issue #1662, opened 2020-10-22
BLOB / TEXT is allowed to have default value in MySQL 8


SpikedCola:
Hiya, I've run into a message running `propel diff` if defaultExpr="null" for a TEXT column:
`In MysqlPlatform.php line 427:
  BLOB and TEXT columns cannot have DEFAULT values. in MySQL.`

I'd never heard of this limitation, and since I was translating from sql that already had a default null text column into propel, this was confusing.

MySQL docs for 8.0 show that a default is indeed allowed: https://dev.mysql.com/doc/refman/8.0/en/data-type-defaults.html
> The BLOB, TEXT, GEOMETRY, and JSON data types can be assigned a default value only if the value is written as an expression, even if the expression value is a literal:

However another page in the manual seems to contradict this: https://dev.mysql.com/doc/refman/8.0/en/blob.html
> BLOB and TEXT columns cannot have DEFAULT values.

In `MysqlPlatform->getColumnDDL()` if I comment out the check for blob/text
>         if ($sqlType === 'DATETIME') {
>             $def = $domain->getDefaultValue();
>             if ($def && $def->isExpression()) {
>                 // DATETIME values can only have constant expressions
>                 $sqlType = 'TIMESTAMP';
>             }
>         } elseif ($sqlType === 'DATE') {
>             $def = $domain->getDefaultValue();
>             if ($def && $def->isExpression()) {
>                 throw new EngineException('DATE columns cannot have default *expressions* in MySQL.');
>             }
>         }/* elseif ($sqlType === 'TEXT' || $sqlType === 'BLOB') {
>             if ($domain->getDefaultValue()) {
>                 throw new EngineException('BLOB and TEXT columns cannot have DEFAULT values. in MySQL.');
>             }
>         }*/

Valid SQL is generated and works for me through phpmyadmin even on 5.7.
> CHANGE `colname` `colname` TEXT DEFAULT null;

This works for my use case but obviously this check is here for a reason. Can anyone provide some additional information? Thanks!

dereuromark:
Can you make a PR here with your suggested changes please?

PR #1993, opened 2024-03-05
bugfix for #1662 - allow defaultValue for TEXT and BLOB fields (mariadb)

hevengo:
- removed blob/text default value checks
- implemented unquote of default values on TEXT/BLOB fields
- tested with mariadb 11.2.3


smhg (changes requested):
This definitely needs tests which CI can run on both MariaDB and MySQL:

- a simple test case to prove setting a default on BLOB or TEXT works
- a case where the default value itself is wrapped in single quotes

PR #1994, opened 2024-03-20
Allow to use non-primary columns in crossref tables

vol4onok:
Hi guys,

Not sure what was the reason for the change, but in Propel1 it was not mandatory to make the crossref table foreign keys also primary keys. It was handy when the crossref table relation had to be referenced further elsewhere.

I don't see any negative effects enabling this, but please correct me if I'm wrong.

Here is a pseudo-example:
```
<table name="user_group" isCrossRef="true">
  <column name="id" type="integer" primaryKey="true" required="true" />
  <column name="user_id" type="integer"/>
  <column name="group_id" type="integer"/>
  <foreign-key foreignTable="user">
    <reference local="user_id" foreign="id"/>
  </foreign-key>
  <foreign-key foreignTable="group">
    <reference local="group_id" foreign="id"/>
  </foreign-key>
</table>

<table name="user_group_logs">
    <column name="id"type="INTEGER" primaryKey="true" autoIncrement="true" required="true" />
    <column name="user_group_id" type="INTEGER" />
    <column name="log" type="VARCHAR" />
    <foreign-key foreignTable="user_group">
        <reference local="user_group_id" foreign="id" />
    </foreign-key>
</table>
```
https://github.com/propelorm/Propel2/pull/1559


vol4onok:
@PhilinTv It was taken here https://github.com/propelorm/Propel2/pull/1559. I covered this improvement. 
I needed to understand what this fix does In tests everything is clear. This fix didn't work for example in desription.
There are 2 cases. But it's an optional feature and can be released.

gechetspr (approved):

PR #1995, opened 2024-03-20
WIP Update CI with newer PHP and PSQL versions

PhilinTv:
1. We need to include the PHP 8.3 in the standard CI process.
2. We need to ensure compatibility with the latest PSQL versions (atm the jobs are stopped).


smhg:
@PhilinTv if you rebase this, only the Pgsql change remains. If still useful, I think it's good to merge?

PR #1998, opened 2024-03-29
Make error logs more informative

Nidhognit:
### Problem statement
1. The logger that logs internal messages is deep in the package:
- It's very hard to extend the logger if you want to log it in more places than just to stream (like New Relic etc.).
- It's not so useful when you see an exception in one place, and details for this exception in another, for HiLoad projects it takes a huge amount of time to compare them.
2. Logs are not informative:
- For HiLoad projects with millions of rows in the table it's very hard to understand what's wrong with data processing when you see only failed SQL without any details about parameters.
- All errors are the same - because we log only SQL all logs are the same, even if the reason and place for them are different.

With more informative logs development and bugfixing will be much more easy.

### DIFF
--- a/src/Propel/Runtime/ActiveQuery/QueryExecutor/AbstractQueryExecutor.php
+++ b/src/Propel/Runtime/ActiveQuery/QueryExecutor/AbstractQueryExecutor.php
@@ -111,7 +111,7 @@ abstract class AbstractQueryExecutor
             }
             $stmt->execute();
         } catch (Throwable $e) {
-            $this->handleStatementException($e, $sql, $stmt ?: null);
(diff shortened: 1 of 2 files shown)


mringler:
Hey @Nidhognit ,

have you checked output after enabling debug mode? This will give you fully formed queries in error messages, which includes the values set in the prepared statement.

You can enable debug mode through the service container:
```php
Propel::getServiceContainer()->useDebugMode()
```
or per connection:
```php
Propel::getServiceContainer()->getWriteConnection('default')->useDebug();
```
The idea behind using debug mode is that Propel should not set parameters in error messages from a production database, as it might expose sensitive data to log files  (see discussion in #1727).

Please check if debug output solves your issue.

PR #2007, opened 2024-05-22
Fix tests for MySQL 8.4: add unique index for complex fks

mringler:
see https://github.com/mringler/Propel2/pull/11

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,13 +16,15 @@ jobs:
             matrix:
                 php-version: [ '7.4', '8.2' ]
                 db-type: [ sqlite, mysql, pgsql, agnostic ]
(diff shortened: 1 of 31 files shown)


smhg:
@mringler would you be able to push a rebase?
I know it is a trivial change, but there should then remain one unique index being added.
Do you remember why you added that one? As tests run without it.

mringler:
Hey Sam,
those are basically the same schema. Not sure if the `schemas/schema.xml` one is even used, but as long as it is not removed, I thought it's best to keep it valid.

This is the fix from perpl, but I think by now there are two (or three?) other PRs that change only the first schema - I am happy to close this PR in favor of those. Let me know!
Best,
Moritz

smhg:
Sorry, I just noticed the other schema was already included in #2034. So feel free to close this. Thanks!

PR #2011, opened 2024-06-28
FIX failing queries on datetime fields set to 1970-01-01 00:00:00 via (int) 0

### LINKED ISSUE #2010: not nullable date and time fields cannot be set on exactly 1970-01-01 00:00:00 via (int)0 as that loosely evaluates to false in Propel\Runtime\Util\PropelDateTime
The offending code:

```
    public static function newInstance($value, ?DateTimeZone $timeZone = null, string $dateTimeClass = 'DateTime')
    {
        if ($value instanceof DateTimeInterface) {
            return $value;
        }
        **if (!$value) {**
            // '' is seen as NULL for temporal objects
            // because DateTime('') == DateTime('now') -- which is unexpected
            return null;
        }
```

It should be rather

```
if ($value === false || $value === null || $value === '') {
    ....
}
```

PR incoming when I figure out running the tests locally ;)

hsegnitz:
The Problem was the not-type-safe null-check in the PropelDateTime util class.
Passing an Integer 0 (that e.g. comes from the SplFileInfo::getMtime()) to the setter of the field, results in the 0 being interpreted as null, thus the update/insert query is failing.

FIxes #2010

Changelog: Improved falsy value handling in PropelDateTime; now treats '', false, and null as null, and '0' as the Unix epoch

### DIFF
--- a/src/Propel/Runtime/Util/PropelDateTime.php
+++ b/src/Propel/Runtime/Util/PropelDateTime.php
@@ -116,7 +116,7 @@ class PropelDateTime extends DateTime
         if ($value instanceof DateTimeInterface) {
             return $value;
         }
-        if (!$value) {
+        if ($value === false || $value === null || $value === '') {
             // '' is seen as NULL for temporal objects
             // because DateTime('') == DateTime('now') -- which is unexpected
             return null;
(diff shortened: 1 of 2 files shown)


hsegnitz:
Hmmm... I'm afraid I might need some help with the failing CI tests, to me that looks unrelated to my change and more like some init failure?

smhg:
@PhilinTv this makes sense to me.

This deserves a note in the release notes I think:
> Remove `defaultValue="0"` from DATETIME|TIMESTAMP column elements in your schema. 0 will now set it to the unix epoch instead of NULL.
This is also applicable if you do the same calling a setter of a generated class.

smhg:
@hsegnitz can you please rebase this PR so tests run? Test errors were fixed recently and that's the easiest way to make them run currently :)
This looks good to merge afterwards (if the breaking change is acceptable to others).

PhilinTv (approved):
Improved falsy value handling in PropelDateTime; now treats '', false, and null as null, and '0' as the Unix epoch

PhilinTv:
@hsegnitz thanks for the fix!

PR #2012, opened 2024-07-23
Templates fix (#1)

elacasse:
Corrections to templates so that no references to nonexistent methods and attributes are generated on build.

Signed-off-by: Eric Lacasse lacasse.eric@gmail.com


smhg (changes requested):
@elacasse thank you for your contribution.
For these changes to be reviewable, I'd suggest to split it up into separate PRs (or at least commits) by what you tried to achieve. I currently can't see where you found "references to nonexistent methods and attributes".

Issue #2009, opened 2024-06-21
Non-static method Propel\Runtime\ActiveQuery\ModelCriteria::delete()` cannot be called statically


phpfui:
Getting the following error from the code generated by `propel init`:

Fatal error: Uncaught Error: Non-static method Propel\Runtime\ActiveQuery\ModelCriteria::delete()` cannot be called statically in C:\websites\php-orm-sql-benchmarks\SOB\Propel2\Base\EmployeeQuery.php on line 222

Indeed, ModelCriteria::delete is not static.

Running PHP 8.2.13 (but this is not a PHP issue, PHP is reported the error correctly)

Composer:
```
"propel/propel": "2.0.0-beta4",
```

The generated method in question:
```php
	public function delete(?ConnectionInterface $con = null) : int
	{
		if (null === $con) {
			$con = Propel::getServiceContainer()->getWriteConnection(EmployeeTableMap::DATABASE_NAME);
		}

		$criteria = $this;

		// Set the correct dbName
		$criteria->setDbName(EmployeeTableMap::DATABASE_NAME);

		// use transaction because $criteria could contain info
		// for more than one table or we could emulating ON DELETE CASCADE, etc.
		return $con->transaction(static function() use ($con, $criteria) {
			$affectedRows = 0; // initialize var to track total num of affected rows

			EmployeeTableMap::removeInstanceFromPool($criteria);

			$affectedRows += ModelCriteria::delete($con);  // this is line 222
			EmployeeTableMap::clearRelatedInstancePool();

			return $affectedRows;
		});
	}
```

rizwanjiwan:
Hey, I tired reproducing this in my project (php 8.3.6). Seems to work fine. Since the generated Query base classes inherit from ModelCriteria, ModelCriteria::delete($con) reads to me to be equivalent to parent::delete($con): http://php.adamharvey.name/manual/en/keyword.parent.php

It's not recommended, but still valid PHP.

I'm doing this on my end (on my implementation of an employee model): EmployeeQuery::create()->filterById(2)->delete()

I'm happy to try and help if you let me know if you see it differently or have more information.

phpfui:
So it looks like it should be parent::delete($con) and not ModelCriteria::delete($con).

I think this changed in a recent version of PHP. Because ModelCriteria::delete does not reference $this, it must be static. parent::delete has an implicit $this, so it calls the parent method, which is not static. That should work, but I don't know the code, which is why I asked.

[...]

PR #2016, opened 2024-09-13
use parent:: instead of ModelCritera::

rizwanjiwan:
As recommended in php docs http://php.adamharvey.name/manual/en/keyword.parent.php

#2009

### DIFF
--- a/tests/Fixtures/schemas/schema.xml
+++ b/tests/Fixtures/schemas/schema.xml
@@ -89,7 +89,6 @@
             <reference local="customer_id" foreign="id"/>
         </foreign-key>
         <foreign-key foreignTable="bookstore_contest" onDelete="cascade">
-            <reference local="bookstore_id" foreign="bookstore_id"/>
             <reference local="contest_id" foreign="id"/>
         </foreign-key>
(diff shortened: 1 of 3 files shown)


tests/Fixtures/schemas/schema.xml
```diff
@@ -89,7 +89,6 @@
             <reference local="customer_id" foreign="id"/>
         </foreign-key>
         <foreign-key foreignTable="bookstore_contest" onDelete="cascade">
-            <reference local="bookstore_id" foreign="bookstore_id"/>
```
smhg, line 92: @rizwanjiwan What was your reasoning to remove this? This apparently made the CI error go away, but I'm wondering what you figured out about it.

smhg: @rizwanjiwan it looks like the foreign key you removed isn't used. I've addressed the issue in PR #2021 by adding the unique index this foreign key needs so it doesn't need to be removed.

rizwanjiwan: Sounds good

PR #2015, opened 2024-09-03
Fix php8.2 deprecation - Use of "parent" in callables is deprecated

ldaspt:
@see https://3v4l.org/dbOp6 when using `is_callable('parent::preSave')`

```
Output for 8.2.0 - 8.2.23, 8.3.0 - 8.3.11
    Deprecated: Use of "parent" in callables is deprecated in /in/dbOp6 on line 15
    Foo::preSave
    Bar::preSave
Output for 5.3.0 - 5.3.29, 5.4.0 - 5.4.45, 5.5.0 - 5.5.38, 5.6.0 - 5.6.40, 7.0.0 - 7.0.33, 7.1.0 - 7.1.33, 7.2.0 - 7.2.34, 7.3.0 - 7.3.33, 7.4.0 - 7.4.33, 8.0.0 - 8.0.30, 8.1.0 - 8.1.29
    Foo::preSave
    Bar::preSave
```

@see https://3v4l.org/OrSnV when using `is_callable([parent::class, 'preSave'])`, the syntax is also compatible with PHP 7.4, minimal version of the `Propel` library

```
Output for 5.5.0 - 5.5.38, 5.6.0 - 5.6.40, 7.0.0 - 7.0.33, 7.1.0 - 7.1.33, 7.2.0 - 7.2.34, 7.3.0 - 7.3.33, 7.4.0 - 7.4.33, 8.0.0 - 8.0.30, 8.1.0 - 8.1.29, 8.2.0 - 8.2.23, 8.3.0 - 8.3.11
    Foo::preSave
    Bar::preSave
Output for 5.4.0 - 5.4.45
    Parse error: syntax error, unexpected 'class' (T_CLASS), expecting identifier (T_STRING) or variable (T_VARIABLE) or '{' or '$' in /in/OrSnV on line 15

    Process exited with code 255.
```

PR #2017, opened 2024-09-26
fix timestampable use of datetimeclass

smhg:
The timestampable behavior doesn't use the `generator.dateTime.dateTimeClass` setting.
I took a few liberties adressing this:

- Reusing the getDateTimeClass method of `Propel\Generator\Builder\Om\ObjectBuilder`. For this it had to become `public`.
- `PropelDateTime::createHighPrecision` expects the output of `microtime` to be formatted with a dot. This places the formatting in a separate (static) method.
This allows keeping the updated_at and created_at microtimes in preInsert exactly the same value (instead of generating them independently).

I've avoided simplifying code to not break things.

What do you think?

### DIFF
--- a/src/Propel/Generator/Behavior/Timestampable/TimestampableBehavior.php
+++ b/src/Propel/Generator/Behavior/Timestampable/TimestampableBehavior.php
@@ -9,6 +9,7 @@
 namespace Propel\Generator\Behavior\Timestampable;
 
 use Propel\Generator\Builder\Om\AbstractOMBuilder;
+use Propel\Generator\Builder\Om\ObjectBuilder;
 use Propel\Generator\Model\Behavior;
 
 /**
@@ -102,7 +103,12 @@ class TimestampableBehavior extends Behavior
     {
         if ($this->withUpdatedAt()) {
             $updateColumn = $this->getTable()->getColumn($this->getParameter('update_column'));
-            $dateTimeClass = $builder->getDateTimeClass($updateColumn);
+
+            $dateTimeClass = 'DateTime';
+
+            if ($builder instanceof ObjectBuilder) {
+                $dateTimeClass = $builder->getDateTimeClass($updateColumn);
+            }
 
             $valueSource = strtoupper($updateColumn->getType()) === 'INTEGER'
                 ? 'time()'
@@ -130,7 +136,12 @@ $mtime = PropelDateTime::formatMicrotime(microtime(true));';
 
         if ($this->withCreatedAt()) {
             $createColumn = $this->getTable()->getColumn($this->getParameter('create_column'));
-            $dateTimeClass = $builder->getDateTimeClass($createColumn);
+
+            $dateTimeClass = 'DateTime';
+
+            if ($builder instanceof ObjectBuilder) {
+                $dateTimeClass = $builder->getDateTimeClass($createColumn);
+            }
 
             $script .= "
 \$highPrecisionCreate = PropelDateTime::createHighPrecision(\$mtime, '$dateTimeClass');";
@@ -146,7 +157,12 @@ if (!\$this->isColumnModified(" . $this->getColumnConstant('create_column', $bui
 
         if ($this->withUpdatedAt()) {
             $updateColumn = $this->getTable()->getColumn($this->getParameter('update_column'));
-            $dateTimeClass = $builder->getDateTimeClass($updateColumn);
+
+            $dateTimeClass = 'DateTime';
+
+            if ($builder instanceof ObjectBuilder) {
+                $dateTimeClass = $builder->getDateTimeClass($updateColumn);
+            }
 
             $script .= "
 \$highPrecisionUpdate = PropelDateTime::createHighPrecision(\$mtime, '$dateTimeClass');";
--- a/src/Propel/Generator/Behavior/Timestampable/TimestampableBehavior.php
+++ b/src/Propel/Generator/Behavior/Timestampable/TimestampableBehavior.php
@@ -101,9 +101,12 @@ class TimestampableBehavior extends Behavior
     public function preUpdate(AbstractOMBuilder $builder): string
     {
         if ($this->withUpdatedAt()) {
-            $valueSource = strtoupper($this->getTable()->getColumn($this->getParameter('update_column'))->getType()) === 'INTEGER'
+            $updateColumn = $this->getTable()->getColumn($this->getParameter('update_column'));
+            $dateTimeClass = $builder->getDateTimeClass($updateColumn);
+
+            $valueSource = strtoupper($updateColumn->getType()) === 'INTEGER'
                 ? 'time()'
-                : '\\Propel\\Runtime\\Util\\PropelDateTime::createHighPrecision()';
+                : "\\Propel\\Runtime\\Util\\PropelDateTime::createHighPrecision(null, '$dateTimeClass')";
 
             return 'if ($this->isModified() && !$this->isColumnModified(' . $this->getColumnConstant('update_column', $builder) . ")) {
     \$this->" . $this->getColumnSetter('update_column') . "({$valueSource});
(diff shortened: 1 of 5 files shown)


smhg:
@dereuromark based on other PR's it seems the CI errors where not introduced here. I think it is better to address at them in a separate PR, no?

smhg:
I get little notifications of GitHub myself, so tagging @PhilinTv and @vol4onok.

smhg:
@gechetspr it seems you are my best bet to ping :) Thank you!

This PR would enable the timestampable behaviour to use `DateTimeImmutable` too (instead of only `DateTime`).

The CI failures are the same as on all other recent PRs (see comment on #2015). I'm not sure what's happening with mysql, but I will try to have a look at the code-style error.

src/Propel/Runtime/Util/PropelDateTime.php
```diff
@@ -73,31 +73,41 @@ protected static function isTimestamp($value): bool
...
      * @throws \InvalidArgumentException
      *
-     * @return \DateTime
+     * @return DateTimeInterface An instance of $dateTimeClass
      */
-    public static function createHighPrecision(?string $time = null): DateTime
+    public static function createHighPrecision(?string $time = null, string $dateTimeClass = 'DateTime'): DateTimeInterface
     {
-        $dateTime = DateTime::createFromFormat('U.u', $time ?: self::getMicrotime());
+        $dateTime = $dateTimeClass::createFromFormat('U.u', $time ?: self::getMicrotime());
```
gechetspr, line 80: can we check that this provided class is an instance of DateTimeInterface in order to make sure that it has createFromFormat method?

gechetspr (changes requested):
Do we have tests for src/Propel/Generator/Behavior/Timestampable/TimestampableBehavior.php ? mb we should update those too to make sure that all works as expected?

src/Propel/Runtime/Util/PropelDateTime.php
smhg: DateTimeInterface::createFromFormat doesn't exist apparently, so I added checks for both classes individually. 
Here the second parameter is also a string (and not DateTime::class), but other methods inside PropelDateTime use that too so I'd rather not mix updating that in this PR.

smhg:
I've resolved your remarks. Good suggestions!
Also phpstan and codesniffer weren't happy yet. I've fixed that.

smhg:
I still need to fix that check it seems. I'll have a look at that.

smhg:
@gechetspr it looks ok now I think.

src/Propel/Runtime/Util/PropelDateTime.php
```diff
@@ -70,34 +71,53 @@ protected static function isTimestamp($value): bool
...
-     * @return \DateTime
+     * @return \DateTimeInterface An instance of $dateTimeClass
      */
-    public static function createHighPrecision(?string $time = null): DateTime
+    public static function createHighPrecision(?string $time = null, string $dateTimeClass = 'DateTime'): DateTimeInterface
     {
-        $dateTime = DateTime::createFromFormat('U.u', $time ?: self::getMicrotime());
+        $allowStringIsA = true;
+
+        if (!is_a($dateTimeClass, DateTime::class, $allowStringIsA) && !is_a($dateTimeClass, DateTimeImmutable::class, $allowStringIsA)) {
```
gechetspr, line 84: Mb we can check against interface and not 2 specific classes?

src/Propel/Runtime/Util/PropelDateTime.php
```diff
@@ -70,34 +71,53 @@ protected static function isTimestamp($value): bool
...
      * @param string|null $time Optional, in seconds. Floating point allowed.
+     * @param string $dateTimeClass Optional, class name of the created object.
      *
      * @throws \InvalidArgumentException
      *
-     * @return \DateTime
+     * @return \DateTimeInterface An instance of $dateTimeClass
      */
-    public static function createHighPrecision(?string $time = null): DateTime
+    public static function createHighPrecision(?string $time = null, string $dateTimeClass = 'DateTime'): DateTimeInterface
```
gechetspr, line 80: just to make sure that everything will work if someone delete DateTime from use statement we can set here '\DateTime' (not sure if DateTime::class will work though)

gechetspr (changes requested):

src/Propel/Runtime/Util/PropelDateTime.php
smhg: The interface doesn't have the `createFromFormat` method:
https://www.php.net/manual/en/class.datetimeinterface.php

I guess using is_a() is the only possibility?

src/Propel/Runtime/Util/PropelDateTime.php
smhg: Me neither, but I'd suggest to not change this in this PR as other methods in this class use this format too (a string).
Better to change this in a future commit separate from this change I think.

mringler (commented):
Nice!
Several two cents on readability though. Hope it makes sense.
Cheers

src/Propel/Generator/Behavior/Timestampable/TimestampableBehavior.php
```diff
@@ -101,9 +103,17 @@ protected function getColumnConstant(string $columnName, AbstractOMBuilder $buil
...
     {
         if ($this->withUpdatedAt()) {
-            $valueSource = strtoupper($this->getTable()->getColumn($this->getParameter('update_column'))->getType()) === 'INTEGER'
+            $updateColumn = $this->getTable()->getColumn($this->getParameter('update_column'));
+
+            $dateTimeClass = DateTime::class;
+
+            if ($builder instanceof ObjectBuilder) {
+                $dateTimeClass = $builder->getDateTimeClass($updateColumn);
+            }
```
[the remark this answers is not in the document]
smhg: Makes sense, thanks!

src/Propel/Runtime/Util/PropelDateTime.php
```diff
@@ -70,34 +71,53 @@ protected static function isTimestamp($value): bool
...
 
-        $dateTime->setTimeZone(new DateTimeZone(date_default_timezone_get()));
+        $dateTime = $dateTime->setTimeZone(new DateTimeZone(date_default_timezone_get()));
 
         return $dateTime;
     }
 
     /**
-     * Get the current microtime with milliseconds. Making sure that the decimal point separator is always ".", ignoring
+     * Format the output of microtime(true) making sure that the decimal point separator is always ".", ignoring
```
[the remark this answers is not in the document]
smhg: I decided to split up fetching and formatting the time up so `updatedAt` and `createdAt` can be _exactly_ the same when you create a new model instance that has the timestampable behaviour. `getMicrotime` should probably be removed, but that's not in the scope of this PR.

src/Propel/Runtime/Util/PropelDateTime.php
```diff
@@ -70,34 +71,53 @@ protected static function isTimestamp($value): bool
...
      * @throws \InvalidArgumentException
      *
-     * @return \DateTime
+     * @return \DateTimeInterface An instance of $dateTimeClass
      */
-    public static function createHighPrecision(?string $time = null): DateTime
+    public static function createHighPrecision(?string $time = null, string $dateTimeClass = 'DateTime'): DateTimeInterface
     {
-        $dateTime = DateTime::createFromFormat('U.u', $time ?: self::getMicrotime());
+        $allowStringIsA = true;
```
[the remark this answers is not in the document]
smhg: Good point: this coding style doesn't seem to be used anywhere in Propel (I think phpmd somehow recommends it?). This whole `is_a` check doesn't read well. `createFromFormat` should be in DateTimeInterface, like everyone seems to expect.

src/Propel/Generator/Behavior/Timestampable/TimestampableBehavior.php
```diff
@@ -123,22 +133,44 @@ public function preUpdate(AbstractOMBuilder $builder): string
...
+            $createColumn = $this->getTable()->getColumn($this->getParameter('create_column'));
+
+            $dateTimeClass = DateTime::class;
+
+            if ($builder instanceof ObjectBuilder) {
+                $dateTimeClass = $builder->getDateTimeClass($createColumn);
+            }
+
+            $script .= "
+\$highPrecisionCreate = PropelDateTime::createHighPrecision(\$mtime, '$dateTimeClass');";
```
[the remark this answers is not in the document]
smhg: This sounds very reasonable, but let's change this in a separate PR. I would like to have this one only be about completing DateTimeImmutable support if that's acceptable.

smhg:
Tagging @PhilinTv to follow up on this.

mringler:
merged into perpl

PhilinTv:
@smhg 
Thanks again for this high-quality contribution! it's well thought-out and cleanly structured. To strengthen the regression safety and ensure config-driven behavior is fully verified, would you be open to adding a few tests?

- A test ensuring created_at and updated_at share the same microtime in preInsert().
- A test verifying that a custom class passed via dateTimeClass is correctly instantiated.
- A negative test asserting that invalid dateTimeClass values throw exceptions as expected.

Let me know if you need guidance - happy to assist!

smhg:
@PhilinTv I've added the first 2 tests you mentioned above. It's definitely nice to have those.

If I understand right, the 3rd test you suggest would test the effect of an invalid `dateTimeClass` configuration property value?
If so, it should probably go into `Propel\Tests\Generator\Builder\Om\GeneratedObjectWithDateImmutableClassTest`.
But if that's the case, I suggest to keep this out of this PR as that is unrelated. Or do you mean something else?

PhilinTv (approved):

PR #2018, opened 2024-11-18
add INT8 to isNumber method

abolfazl-esmaeeli:


smhg:
Someone needs to confirm, but `INT8` seems to be a valid Postgres type.
Thank you @abolfazl-esmaeeli.

Can you please rebase you PR and look at the changes in the review? Thanks!

smhg (changes requested):

PR #2021, opened 2024-12-11
test:mysql fix test setup

smhg:
@gechetspr this should fix the issue with the failing mysql tests.

I don't see how this has ever worked to be honest. If I go back in the git history the offending schema has never been any different in this respect.

When all tests are green, I'd be happy to look at some more deprecation warnings. PHPUnit for instance outputs some.

Edit: FWIW the other failing (phpstan) test is addressed separately in #2020.


smhg:
A fix was also in #2007 I just noticed. You can close both PRs if you merge one of them.

smhg:
@PhilinTv can you please merge this?

I'm not sure yet why #2007 contains a second change that does not seem to be necessary to turn tests green. Me or someone else can follow up on that independently of this one.

PhilinTv (approved):

PR #2027, opened 2025-01-17
Fixed `Profiler::slowThreshold` setter method to accept float value.

matweew:
Using Profiler with custom configuration, like `Propel::getServiceContainer()->setProfilerConfiguration(['slowThreshold' => 0.01]);` it not possible to set float value, when the property itself has `float` type.

### DIFF
--- a/src/Propel/Runtime/Util/Profiler.php
+++ b/src/Propel/Runtime/Util/Profiler.php
@@ -76,13 +76,13 @@ class Profiler
     }
 
     /**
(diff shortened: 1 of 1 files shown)


smhg:
Good catch @matweew. `int` was added in 2ff411a1b133f3377dc9c2053631be8efe2a408e by mistake it seems.
Can you rebase so all tests can run? There was an issue with them previously.
We should be able to get this merged.

gechetspr (approved):

PR #2032, opened 2025-03-04
Log the full error message at once

Nidhognit:
**Problem statement:**
Originally, propel logs have only an exception, without an explanation of what happened.
And it returns the second part of the exception as a public message.

This fix just log them both together.

### DIFF
--- a/src/Propel/Runtime/ActiveQuery/QueryExecutor/AbstractQueryExecutor.php
+++ b/src/Propel/Runtime/ActiveQuery/QueryExecutor/AbstractQueryExecutor.php
@@ -131,13 +131,16 @@ abstract class AbstractQueryExecutor
     protected function handleStatementException(Throwable $e, ?string $sql, $stmt = null): void
     {
         $internalMessage = $e->getMessage();
-        Propel::log($internalMessage, Propel::LOG_ERR);
 
         $isDebugMode = $this->connectionIsInDebugMode();
         if ($isDebugMode && $stmt instanceof StatementWrapper) {
             $sql = $stmt->getExecutedQueryString();
         }
         $publicMessage = "Unable to execute statement [$sql]";
+
+        $fullLogMessage = $publicMessage . PHP_EOL . "Reason: [$internalMessage]";
+        Propel::log($fullLogMessage, Propel::LOG_ERR);
+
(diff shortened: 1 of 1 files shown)


src/Propel/Runtime/ActiveQuery/QueryExecutor/AbstractQueryExecutor.php
```diff
@@ -131,13 +131,16 @@ protected function executeStatement(PreparedStatementDto $preparedStatementDto)
...
         $internalMessage = $e->getMessage();
-        Propel::log($internalMessage, Propel::LOG_ERR);
 
         $isDebugMode = $this->connectionIsInDebugMode();
         if ($isDebugMode && $stmt instanceof StatementWrapper) {
             $sql = $stmt->getExecutedQueryString();
         }
         $publicMessage = "Unable to execute statement [$sql]";
+
+        $fullLogMessage = $publicMessage . PHP_EOL . "Reason: [$internalMessage]";
```
PhilinTv, line 141: Why would you show internal messages outside of debugging mode?
They might contain sensitive data, which logged in production may bring a leak.

Why do you think it's a safe option to show?

Nidhognit: @PhilinTv see line 134, this log was already logged, but for some reason, the propel log half of the message in one log and half in another, for real projects it'a huge pain to understand whats going on in a such way.
I just put full log in one place, and there I added the public log to the internal one

smhg:
@PhilinTv @gechetspr 

Are you sure this is a good idea? This causes double log messages, no?
Generated classes contain `Propel::log()` statements (see 
[Propel\Generator\Builder\Om\QueryBuilder.php](https://github.com/propelorm/Propel2/blob/7548864872608e459e84c1cb46a04b3326ff11b0/src/Propel/Generator/Builder/Om/QueryBuilder.php#L679) for instance) to catch the exception that is thrown here (in `AbstractQueryExecutor`).

Yes, you need to turn debug mode on if you want the full reason. But that's sensible, no?

Also, the mentioned line 134 is empty. @Nidhognit can you please link to the line you mean?

PR #2034, opened 2025-03-12
Use updated calls of GH actions api

PhilinTv:
- GA Action fails with new GH API
- Fix an issue with the test in bookstore unit tests.


smhg:
@PhilinTv if you rebase this is also good to merge. You'll have conflicts as current master already contains newer GitHub action versions, so watch out not to overwrite them.


It closes #2021 too.

PR #2041, opened 2025-04-21
Update supported php versions in CI config

### LINKED ISSUE #2040: Drop PHP 7.4 in CI
- Security support has ended more than two years ago (2 years and 4 months at time of writing)
- Tests in CI for this version are more flaky and have a lot of conditionals involved
- We should encourage users to update to (more) secure runtime versions

eelkevdbos:
- Add currently supported versions to matrix
- Remove versions without security updates

Closes #2040

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,25 +14,16 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php-version: [ '7.4', '8.2' ]
+                php-version: [ '8.3', '8.4' ]
                 db-type: [ sqlite, mysql, pgsql, agnostic ]
                 symfony-version: [ '5-min', '5-max', '6-min', '6-max', '7-min', '7-max']
-                exclude:
-                  - symfony-version: '6-min'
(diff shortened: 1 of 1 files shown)


smhg:
Thank you @eelkevdbos! It absolutely makes sense.

I'm leaning towards not doing this _now_ though. Due to its history Propel likely runs on exceptionally old environments.
I wouldn't be surprised if people are likely still using some versions on PHP 5. (Unofficially) supporting 7.4 might give them a (slightly) easier upgrade path.

Maybe this way we can aim for a 2.0.0 release which is still compatible with PHP 7.4? CI would tell us.

smhg:
FWIW Symfony 5.4 is still supported and that runs on PHP 7.2.5 and above.
So dropping PHP 7.x support in Propel would mean dropping Symfony 5.4 support too.

I've been looking at the whole CI support matrix a bit. It's more tricky than the changes you propose I think.
Currently, old mysql and old postgres tests run only once (on PHP 7.4). You will make them run on all PHP versions. I think it would nice to symplify CI a bit first.

PR #2044, opened 2025-05-03
Limit CI matrix to supported Symfony versions

smhg:
Previously tests would run with Propel using **Symfony components** for each major Symfony versions with the lowest possible version numbers in that major (e.g. `6.0.0` for Symfony 6) and the most recent version within that same major (e.g. `6.4.21` for Symfony 6).

This is a proposal to limit the CI matrix to only the most recent version within each major Symfony version.

I couldn't come up with a reason to run tests on the minimum installable (_unmaintained_) Symfony versions. It would be nice if someone could shed some light here. Maybe in the past Symfony had a different support schedule?

The advantage is that the CI matrix is a bit shorter. This makes things like adding different PHP versions a bit less cumbersome.


PhilinTv:
Yes, that's a lot. Same as the Propel2 compatibility!
The number of tests doesn't bother so much, however let's just strip down compatibility after the next release.

PhilinTv:
@smhg For now the # of version isn't bothering, We can change the full check from every PR to regular, if will be annoying to run to many or flickery tests will appear. 

Meanwhile I'd prefer to keep wide confirmation of compatibility.

PR #2046, opened 2025-05-29
update phpstan to version 2

smhg:
Following the [upgrade guide](https://github.com/phpstan/phpstan/blob/2.1.x/UPGRADING.md) this PR:

- adds the [phpstan/phpstan-deprecation-rules](https://github.com/phpstan/phpstan-deprecation-rules) dependency.
- fixes PHP deprecations:
  - [strftime](https://www.php.net/manual/en/function.strftime.php) (from PHP 8.1): if enabled through the `generator.objectModel.addTimeStamp` setting, a timestamp is added to docblocks of generated classes. `strftime('%c')` is replaced with `gmdate('c')` to add a timestamp that is still human-readable but not locale sensitive.
  - [xml_set_object](https://www.php.net/manual/en/function.xml-set-object.php) (from PHP 8.4): it was not necessary because the `xml_set_element_handler` call specifies the handelers as callables.
- removes a deprecation message which was related to the missing [XSLTProcessor](https://www.php.net/manual/en/class.xsltprocessor.php) class. That can be installed through the PHP XSL extension.
- ignores internal deprecations.
- generates a new phpstan baseline.


PhilinTv (approved):
Fast and clean 👍 
Thanks!

PR #2048, opened 2025-06-04
add supported php versions to ci

smhg:
This adds all PHP versions that Propel should currently work on to CI.
It also disables tests with Symfony components which are not supported on some of these versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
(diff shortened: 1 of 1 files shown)


smhg:
The failling test seems to be an GH issue due to a timeout. It might however be an indicator of too many checks?
In that light #2044 might make sense @PhilinTv?

Additionally, I think there should be a release still supporting these PHP versions. But afterwards we can drop everything below 8.2 I think?

PhilinTv (approved):

PR #2061, opened 2026-01-12
Fix pagination with custom select columns and improve versionable detection

c15k0:
Fix pagination with custom select columns and improve versionable detection

- Fix pagination failure when using custom select columns in Criteria
- Add explicit versionable detection to VersionableBehavior and Table (avoid dynamic properties, required for PHP 8+ compatibility)
- Remove failing PHPStan test caused by dynamic property usage
- Update query class documentation generation for full PhpStorm indexing support


mringler:
Have a look how this was solved in Perpl [VersionableBehavior.php](https://github.com/perplorm/perpl/blob/62597877f3735e6e7f921bb0c14ef45a5625f0d0/src/Propel/Generator/Behavior/Versionable/VersionableBehavior.php#L129) - you can just keep track of the version tables inside the behavior.

PR #2062, opened 2026-01-27
Fix PHP 8.5 deprecation, use (bool) cast in generated code

pableu:
These "non-standard cast names" are deprecated in PHP 8.5 and `(bool)` is supported in all PHP versions that propel supports.


oojacoboo:
Pretty sure this has been resolved in [Perpl](https://github.com/perplorm/perpl).

pableu:
> Pretty sure this has been resolved in [Perpl](https://github.com/perplorm/perpl).

Thanks a ton for the hint, just switched out project to Perpl.

PR #2063, opened 2026-02-10
fix(mysql): allow DEFAULT values for TEXT columns

hevengo:
allow DEFAULT values for TEXT columns on MySQL 8.0.13+ and MariaDB 10.2.1+

Remove the EngineException that blocked DEFAULT values for TEXT columns while keeping it for BLOB types. Update the reverse schema parser to preserve TEXT default values during introspection, with handling for MariaDB's extra single-quote wrapping on TEXT defaults.

### DIFF
--- a/src/Propel/Generator/Platform/MysqlPlatform.php
+++ b/src/Propel/Generator/Platform/MysqlPlatform.php
@@ -464,9 +464,9 @@ DROP TABLE IF EXISTS " . $this->quoteIdentifier($table->getName()) . ";
             if ($def && $def->isExpression()) {
                 throw new EngineException('DATE columns cannot have default *expressions* in MySQL.');
             }
-        } elseif ($sqlType === 'TEXT' || $sqlType === 'BLOB') {
+        } elseif ($sqlType === 'BLOB') {
             if ($domain->getDefaultValue()) {
-                throw new EngineException('BLOB and TEXT columns cannot have DEFAULT values. in MySQL.');
+                throw new EngineException('BLOB columns cannot have DEFAULT values in MySQL.');
             }
         }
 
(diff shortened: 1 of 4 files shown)


mringler:
Very nice! Spotless too (though unconventional boundary characters in regex always trip me up).

The state of Propel2 test suite hides it well, but the added test in MysqlSchemaParserTest currently fails:
```
1) Propel\Tests\Generator\Reverse\MysqlSchemaParserTest::testTextDefaultValues
--- Expected
+++ Actual
@@ @@
-'hello text'
+'_utf8mb3\'hello text\''

/home/runner/work/Propel2/Propel2/tests/Propel/Tests/Generator/Reverse/MysqlSchemaParserTest.php:143
```
(see [run output](https://github.com/propelorm/Propel2/actions/runs/21864289016/job/63101242774?pr=2063))

Once this is figured out, I'd be happy to add this to [Perpl](https://github.com/perplorm/perpl) if you don't mind?

hevengo:
fixed the said test-case - but tests are still failing in obscure ways

mringler:
Oh, that is a new issue with Github. Apparently they broke the tool that runs docker containers in actions (see for example [issue report in old action repo](https://github.com/getong/mariadb-action/issues/18)). @oojacoboo fixed it for Perpl. 

I checked out the PR and pushed a branch [here](https://github.com/perplorm/perpl/compare/develop...perplorm:perpl:bugfix-text-defaults-handling), but tests are still failing (see [run](https://github.com/perplorm/perpl/actions/runs/22026509635)). Tried locally with more slashes (`~^(?:_\w+)?\\\\\'(.*)\\\\\'$~`), but then MariaDB doesn't work anymore (output is `''hello text''`).
Running `composer run cs-fix` fixes the code style complaints automatically.

Let me know if I can be of assistance.

mringler:
I think I figured it out, the issue is MariaDB encodes the default string once, but MySQL does it twice. Have a look at perplorm#118, does that seem ok?

Sorry for butting in, I need to get rid of that branch again.

Checkout Perpl, it's still maintained and improved Propel2. Would have been fun to figure this out together!

PR #2064, opened 2026-04-16
Collection class name performance fix

Nidhognit:
We check whether the collection class name exists many times, and under heavy data load, this becomes a significant performance issue.

image


mringler:
Guessing collection name overall is a weird feature. Looks like a side-effect to me. Error-prone too. Interesting that it also dents performance.

Have a look at [Perpl](https://github.com/perplorm/perpl), it is improved and still maintained Propel. We removed the name-guessing completely.

If you do performance tests, I'd love to know how Perpl does!

gechetspr (approved):

PR #2070, opened 2026-07-28
Fix schema-qualified archivable table lookup

surekap:
## Summary

Fix `ArchivableBehavior` so it reuses an explicitly declared archive table when that table is registered under a schema-qualified name.

The fix covers both:

- a schema declared on the `<database>` element; and
- a schema declared directly on the source and archive `<table>` elements.

## Problem

An archivable source table may reference an existing archive table using the customary unqualified parameter:

```xml
<table name="source_table" schema="archive_schema">
    <behavior name="archivable">
        <parameter name="archive_table" value="archive_table"/>
    </behavior>
</table>

<table name="archive_table" schema="archive_schema">
    <!-- custom archive definition -->
</table>
```

Propel registers the existing table as `archive_schema.archive_table`. However, `ArchivableBehavior::addArchiveTable()` currently checks:

```php
$database->hasTable('archive_table');
```

`Database::hasTable()` performs an exact registry lookup and therefore returns `false`. The behavior then calls `createArchiveTable()`, which inherits the source table's schema and attempts to register the same qualified table again:

```text
Table "archive_schema.archive_table" declared twice
```

This prevents model generation for valid schemas that explicitly define their archive table, for example to retain custom columns, indexes, or constraints.

## Root cause

The behavior assumes archive-table names are stored unqualified. That assumption does not hold when schema support is enabled:

- `Database::getTable()` normalizes an unqualified name using the database schema.
- Tables with their own schema are indexed by their fully qualified name and need a table-schema-aware fallback.
- The original `hasTable()` check does neither.

## Changes

- Resolve the configured archive table with `Database::getTable()` first, allowing database-level schema normalization.
- If it is still unresolved and the source table has its own schema, look up:

  ```text
  <source schema><platform schema delimiter><archive table>
  ```

- Use the platform's configured schema delimiter instead of assuming `.`.
- Do not qualify an `archive_table` parameter that is already qualified.
- Reuse the resolved `Table` instance and preserve the existing `sync` behavior.
- Add data-driven regression coverage for database-level and table-level schemas.

The regression test also verifies that:

- no duplicate archive table is created;
- a custom archive-only column remains present; and
- the existing archive table is not modified when `sync` retains its default value of `false`.

## Backward compatibility

Unqualified schemas keep the existing behavior.

If no matching archive table exists, Propel still creates one through `createArchiveTable()` as before. Explicitly qualified `archive_table` values are looked up directly and are not qualified a second time.

## Validation

Passing:

- `ArchivableBehaviorTest.php`: 25 tests, 56 assertions
- `ArchivableBehaviorSyncTest.php`: 7 tests
- New schema regression cases: 2 tests, 8 assertions
- PHPStan on the modified production class
- PHPCS on the modified production class
- PHPCS reports no violations on the newly added test lines
- `git diff --check`

The agnostic suite completed 2,481 tests locally. It reported two existing platform-dependent `iconv` transliteration failures on macOS; all Archivable behavior tests passed.

Repository-wide PHPStan and Psalm could not be used as clean local signals under PHP 8.5 because existing `xml_parser_free()` / `PDO::ERRMODE_WARNING` deprecations fail PHPStan and the bundled Psalm 4 PHAR crashes on PHP 8.5 deprecations. The affected production class passes PHPStan independently.


surekap:
Full disclosure: I generated the PR with OpenAI Codex. 
I have an external schema with a custom archive table (archivable behavior). That schema and table uses name=spic_sureka which is a different connection to a different database on the same server. It is also part of a different application.

`propel diff` searches for the archive table in the named database connection for the current application (which is a different database) and fails.

mringler:
Interesting problem! Great PR!

Question though: Isn't the problem that the unqualified archive table is generated inside the parent table schema?

To me, this:
```xml
<table name="source_table" schema="archive_schema">
    <behavior name="archivable">
        <parameter name="archive_table" value="archive_for_source_table"/>
    </behavior>
</table>
```
should use a `archive_for_source_table` table in global namespace, while
```xml
<table name="source_table" schema="archive_schema">
    <behavior name="archivable">
        <parameter name="archive_table" value="archive_schema.archive_for_source_table"/>
    </behavior>
</table>
```
should use the namespaced table `archive_schema.archive_for_source_table`.

After all, how would you specify an archive table in global namespace? I think this should be possible, right?

mringler:
I looked how schema is handled in foreign keys and other behaviors. Expected behavior seems to be:
- schema is inherited from parent/source table
- a different schema can be provided via parameter

Currently, the check for an existing target table does not check inside the source table's schema, which this PR fixes.

In Perpl, I have also added the `schema` parameter, the changes also apply to archiveable behavior (see https://github.com/perplorm/perpl/pull/168)

PR #2071, opened 2026-07-29
Fix excluded tables with schema-qualified names

surekap:
## Summary

- match excluded-table entries against both the fully qualified table name and the common table name
- preserve `*` wildcard support while escaping all other regular-expression metacharacters
- add regression coverage for schema-qualified tables and literal pattern characters

## Problem

When a database or table has a schema, `Table::getName()` returns a qualified name such as `grapevine.t_audit_log`. Migration exclusions such as:

```json
{
    "exclude_tables": ["t_*", "oauth_*", "m_*"]
}
```

are therefore ignored, because `t_*` is only compared with `grapevine.t_audit_log`. A migration diff can then include excluded tables and generate destructive statements for them.

The old implementation also interpolated exclusion entries directly into a regular expression. Characters such as `.` were treated as regex operators even though the documented behavior is a table-name pattern with `*` wildcards.

## Fix

`DatabaseComparator` now checks each exclusion against:

- `Table::getName()` for explicitly qualified patterns such as `grapevine.t_*`
- `Table::getCommonName()` for portable patterns such as `t_*`

Patterns are escaped with `preg_quote()` before the escaped `*` is converted to the supported wildcard.

## Verification

- `vendor/bin/phpunit tests/Propel/Tests/Generator/Model/Diff/DatabaseTableComparatorTest.php`
- `vendor/bin/phpcs src/Propel/Generator/Model/Diff/DatabaseComparator.php tests/Propel/Tests/Generator/Model/Diff/DatabaseTableComparatorTest.php`
- `git diff --check`

All pass locally.


mringler:
Nice!
Added to [Perpl](https://github.com/perplorm/perpl)

PR #2075, opened 2026-08-13
for php8.5 version

phili67:


olhalivitchuk:
Hi @phili67,

Thanks for putting this together! PHP 8.5 support landed in the meantime via https://github.com/propelorm/Propel2/pull/2073, merged on 2026-08-21 and released in 2.0.1.

The only real difference is `SchemaReader.php`: this PR keeps `xml_parser_free()` behind
a `version_compare()` guard, while `master` removes the call. Removal looks better to me:
the function has been a no-op since PHP 8.0 ("deprecated since 8.5, as it has no effect
since PHP 8.0") and the package requires `php: >=8.0`, so the guarded branch can never do
anything on a supported version — it just preserves dead code plus a per-call version
check.

So I'd suggest closing this as superseded, unless you see a reason to keep the guarded
variant. Thanks again for caring about 8.5 readiness!

PR #2076, opened 2026-09-14
Add Symfony 8 support

ratenfanger:
Add Symfony 8 support. Done with the help of Claude, but fine tuned and reviewed by myself.

## Changes
 - Constraint constructor signature. Symfony 8's validator constraints drop the legacy single-array-of-options constructor; each option now a positional/named argument.
   - Date.php: add explicit __construct(?string $message, ?string $column, ?array $groups, $payload), forward to parent with null (no options array) plus positional args.
   - Unique.php: same pattern, own parent has no $message param so set manually after.
   - ValidateBehavior.php: generator no longer emits var_export'd options array as sole constructor arg — new buildNamedArguments() renders min: 4, max: 128 style named-argument list instead, since generated code calls new Constraint(...) directly.

- Return type strictness — all Command::configure() overrides (10 files) gain : void return type, matching Console 8's now-enforced signature.

- writeln bitmask narrowing — ConsoleHelper.php: OutputInterface::writeln() now types $options as int<0, 511>; added @phpstan-ignore-next-line since Propel's own doc-block can't express that range and phpcs balks at the syntax.

 - composer.json: widen symfony/* constraints to add || ^8.0.

 - CI: tests/composer/composer-symfony8-{min,max}.json fixtures added, workflow matrix extended.

No behavior change beyond Symfony 8 API compatibility.

## Test
Mostly done by CI (all green in my fork). I also use my fork in a personal project. No problems so far.


ratenfanger:
@gechetspr The failed test job is just an infrastructure failure, retrying the job will fix it.

